PR #6, opened 2016-08-02
[WIP] Initial import of track_oracle.

collinsr:
This builds shared, but I'm not certain the template visibility
hand-waving would pass close scrutiny.

Should bring some tests over.

Removing work-in-progress tag to see if that triggers a buildbot.


collinsr:
Going to close this and resubmit with no WIP tag from the start to see if it triggers a buildbot.

PR #7, opened 2016-08-02
Initial import of track_oracle.

collinsr:
This builds shared, but I'm not certain the template visibility
hand-waving would pass close scrutiny.

Should bring some tests over.

Looking for some buildbot action.


collinsr:
Rebased on current master.

PR #22, opened 2016-10-24
Use a faster method to get the Fully Qualified Domain Name

mleotta:
The previous approach was taking more than 30 seconds to get a FQDN on
some platforms.  The problem was that it iterated through each network
interface and did a domain lookup on each.  Some of these look-ups fail
but the timeout time is very long and non-configurable.

This alternate implementation is based on a suggestion on Stack Overflow:
http://stackoverflow.com/questions/504810/how-do-i-find-the-current-machines-full-hostname-in-c-hostname-and-domain-info

It uses getaddrinfo instead, which provides all the names upfront, so
don't need a look-up for each interface.

If this looks reasonable, we should also push it to upstream KWSYS

### DIFF
--- a/vital/kwiversys/SystemInformation.cxx
+++ b/vital/kwiversys/SystemInformation.cxx
@@ -1798,7 +1798,6 @@ int SystemInformationImplementation::GetFullyQualifiedDomainName(
 
   for(p = info; p != NULL; p = p->ai_next)
     {
-    printf("hostname: %s\n", p->ai_canonname);
     std::string candidate = p->ai_canonname;
     if ((candidate.find(base)!=std::string::npos) && baseSize<candidate.size())
       {
--- a/vital/kwiversys/SystemInformation.cxx
+++ b/vital/kwiversys/SystemInformation.cxx
@@ -1783,53 +1783,32 @@ int SystemInformationImplementation::GetFullyQualifiedDomainName(
   size_t baseSize=strlen(base);
   fqdn=base;
 
-  struct ifaddrs *ifas;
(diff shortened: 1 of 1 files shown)


vital/kwiversys/SystemInformation.cxx
```diff
@@ -1783,53 +1783,32 @@ int SystemInformationImplementation::GetFullyQualifiedDomainName(
...
-    return -2;
+    return gai_result;
     }
 
-  for (ifa=ifas; ifa!=NULL; ifa=ifa->ifa_next)
+  for(p = info; p != NULL; p = p->ai_next)
     {
-    int fam = ifa->ifa_addr? ifa->ifa_addr->sa_family : -1;
-    if ((fam==AF_INET) || (fam==AF_INET6))
+    printf("hostname: %s\n", p->ai_canonname);
```
linus-sherrill: Should this printf be here?

mleotta: no that's a debug statement,  I'll remove it

mleotta:
FYI, I tested this on Mac OS X 10.10 and Ubuntu 14.04 and it seems to work in both cases.  Ubuntu did not have the speed issues to begin with, but Mac did.  So this is likely a Mac-specific problem.

mleotta:
For the record, this change has also been submitted upstream to KWSys:

http://review.source.kitware.com/#/c/21662/

mleotta:
@linus-sherrill we need to revert this change.  The patch was rejected for KWSys.  Apparently this is not an appropriate fix.

PR #25, opened 2016-10-27
Standardize control of adding build tree to loader paths.

linus-sherrill:
CMake files simplified to only use one control (KWIVER_USE_BUILD_TREE)
to control whether the build tree is included in the search path
for processes, clusters, sprokit configuration includes and
algorithm plugins. Previously there were individual controls for each
of these path users, with some options marked as advanced.


mleotta:
Linus, I don't see a reason to remove these advanced options.  Maybe the first should be standardized to use cmake_dependent_option like the rest, but in general they seem to do what you'd expect.  That is, they match that state of KWIVER_USE_BUILD_TREE by default.  If for some reason you want to disable this behavior for a particular component, it's nice to have an advanced option to do that.

The problem with the dashboards is that KWIVER_USE_BUILD_TREE is not set.

PR #31, opened 2016-10-28
Add Travis CI support

mleotta:
This uses the same build approach as MAP-Tk.  We rely on system packages as much as possible to make build time fast.  However, this runs on Ubuntu 12.04, so many of the packages are too old or non-existent.  To work around this we build dependencies from source and then cache them to share between Travis CI runs as long as the Git hash on the project doesn't change.  The code to build dependencies is made for reuse, but in this case only builds Fletch.

Only a subset of KWIVER is enabled at this point.  We can try enabling more in future commits.  The minimal subset of Fletch to support these KWIVER options is used.


.travis.yml
```diff
@@ -0,0 +1,47 @@
...
+    - libgl1-mesa-dev
+    - libxt-dev
+    - libqt4-dev
+    - libatlas-base-dev
+
+script:
+  - export PATH=$HOME/deps/bin:$PATH
+  - mkdir build
+  - cd build
+  - cmake -DKWIVER_ENABLE_ARROWS=ON
```
linus-sherrill: Does this need KWIVER_USE_BUILD_TREE=ON?

mleotta: Yes this will need KWIVER_USE_BUILD_TREE=ON once we enable the sprokit build.  For now it works fine and the Travis CI build is passing.  I suggest we merge it since the test passes and enable that option when we need it.

Better yet, does it make sense to just default KWIVER_USE_BUILD_TREE to ON?  Or maybe default to ON when tests are enabled?  It's kind of odd that the tests don't pass unless you manually enable this option.  Alternatively, Maybe there is a way to explicitly insert the build tree paths when the test code is being run?

PR #32, opened 2016-10-28
Doxygen set project name and version

mleotta:


linus-sherrill:
I did a quick check of the doxygen config files and doxy_project_number is not specified in these files. It looks like there is a spot on Doxyfile.common.in, line 8, but the fill-in is not specified.

mleotta:
I'll add it.  The Doxyfile.common.in in MAP-Tk has had this for a while.  Maybe MAP-Tk should be using the Doxyfile.common.in from KWIVER?  I'll have to look and see if there are any other differences.

linus-sherrill:
The common file is available for any project to use, but it has a lot of sprokit specific aliases. There may be some stylistic assumptions built into that file as well. Also, if this file is used by other projects, changes by the KWIVER project may have unexpected (unwanted) side effects in the others that use this file.

PR #39, opened 2016-11-15
Add polygon

linus-sherrill:
Add initial polygon implementations. There is a minimal implementation in vital. In addition there is a VXL variant.


vital/tests/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ kwiver_discover_tests(core_vector             test_libraries test_vector.cxx)
 kwiver_discover_tests(core_bounding_box       test_libraries test_bounding_box.cxx)
 kwiver_discover_tests(core_vital_uuid         test_libraries test_uuid.cxx)
 kwiver_discover_tests(core_attribute_set      test_libraries test_attribute_set.cxx)
+kwiver_discover_tests(core_polygon            test_libraries test_polygon.cxx)
```
mleotta: These tests should be renamed from "core_*" to "vital_*"

vital/types/polygon.h
```diff
@@ -0,0 +1,214 @@
...
+ * access the implementation specific methods because they usually
+ * provide additional attributes and operations. These derived classes
+ * should supply conversion methods to and from the basic (core)
+ * implementation.
+ *
+ * This class behaviour is considered the specification for all
+ * derived classes.
+ */
+
+class VITAL_EXPORT polygon
```
mleotta: As we discussed, this class is probably simple and small enough that we don't need the abstract base class treatment. We do, however, want to keep conversion functions to and from vgl_polygon in the vxl arrow.  If we later have an algorithm API need for passing around large collections of polygons it may make sense to create a "polygon_set" class that using an abstract base class to avoid copying all the data.

linus-sherrill:
Updated the branch. Simplified the polygon class structure.

mleotta:
@linus-sherrill, can you split the library renaming commit out into as separate pull request?

PR #44, opened 2016-11-17
Remove check for impossible null reference.

mleotta:
Null references do not exist in C++ and we don't need to check for them.
This code was generating the following warning:

warning: reference cannot be bound to dereferenced null pointer in
well-defined C++ code; comparison may be assumed to always evaluate to
false

If there really is a case in exception handling where the reference is
null, it seems that should be handled elsewhere, e.g. before a NULL
pointer is converted to a reference.

### DIFF
--- a/sprokit/src/sprokit/pipeline/process.cxx
+++ b/sprokit/src/sprokit/pipeline/process.cxx
@@ -1803,14 +1803,6 @@ void
 process::priv
 ::connect_input_port(port_t const& port, edge_t const& edge)
 {
-  // This may seem really strange, but it happens during execption
-  // handling.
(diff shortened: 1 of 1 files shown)


hughed2:

hughed2:

hughed2:

mattdawkins:
Any reason this shouldn't be merged? If not I'll do it in a day.

Purg:
LGTM

linus-sherrill:
There was/is an error case where the pipeline throws an error while the pipeline is being created that causes a null reference (which is guaranteed not to happen, but it does). This is caused by an error in a process which is impossible to find without this work-around.

PR #46, opened 2016-11-23
Update defined symbol to match generated export file.

linus-sherrill:
This symbol was not updated when the library name was changed.


mleotta:
looks like the right fix.  I'll merge and let the dashboards verify.

PR #47, opened 2016-11-24
Add missing CMakeDependentOption include in doxygen utils cmake

Purg:
When using kwiver in another project that has not yet explicitly
CMake included ``CMakeDependentOption``, inclusion of ``kwiver-utils``
fails with ``Unknown CMake command "cmake_dependent_option"``. Adding
an include in the doxygen utils CMake file where the use of
``CMakeDependentOption`` is used.


Purg:
@linus-sherrill PTAL

PR #48, opened 2016-11-24
Add export of test utils header dir in build CMake config

Purg:
Added a variable (``KWIVER_TEST_INCLUDE_DIRS``) to be present in ``kwiver-config.cmake`` for a build of
kwiver. This allows projects that use kwiver to not have to replicate
utilities like that defined in ``test_common.h`` that are required for
the ``kwiver-utils-tests`` functions.


Purg:
@linus-sherrill PTAL

linus-sherrill:
This change will only appear in the `kwver-config.cmake` in the build directory (not the install tree). Should this change also be applied to the `kwiver-config-install.cmake.in` file?

Purg:
I don't think so because we don't install the test utility headers that this directory points to. If we do install those utility headers, then it would be relevant to add this directory variable to the installed cmake-config.

PR #51, opened 2016-12-22
Add video_input_process to process plugin.

linus-sherrill:
This process was in the directory, but never registered. This change
adds the process to the list of files to be built,
registers the process and adds the needed data types to the traits.

### DIFF
--- a/sprokit/processes/core/video_input_process.cxx
+++ b/sprokit/processes/core/video_input_process.cxx
@@ -57,9 +57,13 @@ namespace kwiver {
 
 //                 (config-key, value-type, default-value, description )
 create_config_trait( video_filename, std::string, "", "Name of video file." );
-create_config_trait( frame_time, double, "0.3333333", "Inter frame time in seconds" );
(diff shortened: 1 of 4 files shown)


sprokit/processes/core/video_input_process.cxx
```diff
@@ -57,9 +57,13 @@ namespace kwiver {
 
 //                 (config-key, value-type, default-value, description )
 create_config_trait( video_filename, std::string, "", "Name of video file." );
-create_config_trait( frame_time, double, "0.3333333", "Inter frame time in seconds" );
+create_config_trait( frame_time, double, "0.3333333",
```
mleotta: Is this supposed to be 0.03333333?  I'm only asking because 30Hz is a common video rate while 3Hz is much less common.

mleotta:
looks okay to me.

sprokit/processes/core/video_input_process.cxx
linus-sherrill: Good catch. I will fix that.

PR #56, opened 2017-01-23
Support OpenCV 3.X

mleotta:
These changes are to support OpenCV 3.X in KWIVER.  These are fixes for vital and arrows.  I have not build sprokit with OpenCV 3.X yet.  Additional Changes may be needed.


mleotta:
This is needed for KWIVER to build against current Fletch, I'm going to merge it.  We can make further changes after the merge if needed.

PR #63, opened 2017-02-07
added Cereal 1.2.1 as a private dependency

mleotta:
No need to review the code itself, this is a verbatim copy of Cereal 1.2.1 dropped into `vital/3rd_party`.  The question is should we add this to KWIVER and is this the correct way to add it?  Note the top-level CMakeLists.txt change that adds the `vital/3rd_party` directory to the include directories.  This way you can include Cereal headers as `#include <cereal/cereal.hpp>`.  This is required because Cereal includes its own headers in this way.  We are **not** installing Cereal with a KWIVER install, so the Cereal headers can only be included from KWIVER implementation and private header files.

More info on Cereal: http://uscilab.github.io/cereal/


linus-sherrill:
It looks like a capable package. Putting the package under the 3rd-party directory seems a little inspecific like putting things in a closet. There is precedent for putting packages directly under vital (e.g. kwiversys). Alternately we could use "private" or "internal" as directory names.

mleotta:
I don't really care much where we put it, so I'm open to guidance on that.  Originally I just had it directly in vital as `vital/cereal`, but that didn't work because I would need to put vital on the include path to allow including cereal in the way it requires.  Basically we just need a directory somewhere that we are okay with putting on the include path.

mleotta:
I've renamed the encompassing directory to "internal" and added a README for that directory.

PR #66, opened 2017-02-16
Change default edge cap

linus-sherrill:
Added CMake option and supporting code to specify the default edge capacity at compile time. The old default was for unlimited edge capacity which does not work well for large data sets. The edge capacity can also be set in the pipeline file, but this option provides a system wide default.


sprokit/src/sprokit/pipeline/CMakeLists.txt
```diff
@@ -69,24 +69,32 @@ set(utils_build_options)
...
 endif ()
 
 set_source_files_properties(scheduler.cxx
   PROPERTIES
-    COMPILE_DEFINITIONS "${scheduler_build_options}")
+  COMPILE_DEFINITIONS "${scheduler_build_options}")
+
+###
+set(SPROKIT_DEFAULT_EDGE_CAPACITY "40"
+  CACHE STRING "The default capacity for pipeline edges" )
```
mleotta: capacity of 0 == infinity?  If so, it might be good to document that here

PR #67, opened 2017-02-16
Add the KWIVER_USE_BUILD_TREE option to the build.

linus-sherrill:
This option will add the build tree to the plugin loader path so plugins will be found when they are not otherwise installed.


mleotta:
I would prefer to not have to do this to get the tests to pass.  Tests are designed be run from the build tree, possibly before making an install (for which you would not want to build with this option).  As it stands, you either need to set this CMake option or set an environment variable for the tests to pass.

Instead we should do something equivalent to what this CMake options does, but do it all the time for tests (not optionally), and the option should only effect installable tools.

PR #80, opened 2017-03-02
Dev/detector fixes

mattdawkins:


arrows/core/detected_object_set_output_kw18.cxx
```diff
@@ -29,15 +29,21 @@
...
+
 #include <vital/vital_foreach.h>
+#include <vital/util/tokenize.h>
 
+#include <vector>
+#include <fstream>
 #include <time.h>
 
+#include <boost/scoped_ptr.hpp>
+#include <boost/atomic.hpp>
```
linus-sherrill: No boost unless absolutely no other way of implementing, which is rare.

arrows/core/detected_object_set_output_kw18.cxx
```diff
@@ -61,16 +67,22 @@ class detected_object_set_output_kw18::priv
...
+  ~priv() {}
 
   void read_all();
 
   detected_object_set_output_kw18* m_parent;
   kwiver::vital::logger_handle_t m_logger;
   bool m_first;
   int m_frame_number;
+  bool m_write_tot;
+  boost::scoped_ptr< std::ofstream > m_tot_writer;
```
linus-sherrill: Try `std::unique_ptr`

arrows/core/detected_object_set_output_kw18.cxx
```diff
@@ -85,21 +97,64 @@ detected_object_set_output_kw18()
...
 
 
 // ------------------------------------------------------------------
 void
 detected_object_set_output_kw18::
-set_configuration(vital::config_block_sptr config)
-{ }
+set_configuration( vital::config_block_sptr config )
+{
+  d->m_write_tot = config->get_value<bool>( "write_tot" , d->m_write_tot );
```
linus-sherrill: Need to add code to merge in defaults. Otherwise get_value will throw if item is not in config block.

```
set_configuration( vital::config_block_sptr config_in )
{
  // Starting with our generated config_block to ensure that assumed values are present
  // An alternative is to check for key presence before performing a get_value() call.
  vital::config_block_sptr config = this->get_configuration();

  config->merge_config( config_in );
```

arrows/core/detected_object_set_output_kw18.cxx
```diff
@@ -85,21 +97,64 @@ detected_object_set_output_kw18()
...
+  return config;
+}
 
 
 // ------------------------------------------------------------------
 bool
 detected_object_set_output_kw18::
-check_configuration(vital::config_block_sptr config) const
+check_configuration( vital::config_block_sptr config ) const
 {
```
linus-sherrill: The ordering isn't clear between get_configuration(), set_configuration(), and check_configuration().
In some cases check_config() is called first to validate the xxx:type specification for the generic algorithm handling. A nice log message is generated if `type` is not specified of can't be found. All available types are displayed.

This area of configuring an algorithm needs better documentation and examples. The current code does it many different ways so you cant depend on set_configuration() being called first.

PR #84, opened 2017-03-06
Fix assorted WIN32 build issues

mattdawkins:


arrows/core/initialize_cameras_landmarks.cxx
```diff
@@ -143,6 +143,11 @@ class initialize_cameras_landmarks::priv
   {
   }
 
+  /// Destructor
+  virtual ~priv()
```
mleotta: Is this needed or is it left over from experimentation?

arrows/core/initialize_cameras_landmarks.cxx
```diff
@@ -182,8 +187,6 @@ class initialize_cameras_landmarks::priv
   vital::algo::bundle_adjust_sptr bundle_adjuster;
   /// Logger handle
   vital::logger_handle_t m_logger;
-public:
-  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
mleotta: I think if we take one of these out, we really should be taking all of them out. Right @billhoffman ?

arrows/core/initialize_cameras_landmarks.cxx
linus-sherrill: I don't believe the virtual is needed here since there are no other virtual methods.

arrows/core/initialize_cameras_landmarks.cxx
billhoffman: We can git rid of EIGEN_MAKE_ALIGNED_OPERATOR_NEW stuff now

mattdawkins:
I'm abandoning this branch and remaking it with a different branch name because I want to maintain viame/master in it's current state, but still push for this branch to be landed.

PR #85, opened 2017-03-07
Change to README

mattbrown11:
Added steps to the README to resolve implicit dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,60 +6,131 @@ Kitware Image and Video Exploitation and Retrieval
 
 The KWIVER  toolkit is a collection of software tools designed to tackle challenging image and video analysis problems and other related challenges. Recently started by Kitware’s Computer Vision and Scientific Visualization teams, KWIVER is an ongoing effort to transition technology developed over multiple years to the open source domain to further research, collaboration, and product development.
 
-The project is structured with the parent `kwiver` repository working as CMake "super-build" that pulls in a number of KWIVER and other open source projects.
+The project is structured with the parent `kwiver` repository working as a CMake "super-build" that pulls in a number of KWIVER and other open source projects. Following the archery analogy, KWIVER works with arrows (i.e., plugins) to provide flexible integration with other libraries and algorithms.
 
 ## Building KWIVER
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,7 +216,7 @@ After the pipeline completes, you should find a file, `numbers.txt`, in your wor
 
 ### Python Processes
```
mattbrown11: We should probably make mention of the Python bindings to Vital.  Specifically, a good entry point is the self-documenting Jupyter Notebook located in /vital/bindings/python/Kwiver_Python_Demo.ipynb. Perhaps someone with a better feel for the overall architecture can weigh in on how/where this should go.

linus-sherrill:
It is great to have a list of how to install all requirements needed to build fletch, but the description only addresses how to do this for one Ubuntu version. There should be a similar list of requirements for other supported systems (e.g. CentOS, Windows, os-X )

mleotta:
I haven't had a chance to read this in great detail, but my first impression is that we should promote using Fletch as the official way to get the dependencies for KWIVER on all platforms.  There is still value in having instructions for building against system packages on Ubuntu 14.04 or some other platform, but I wouldn't put that first.  It would be better on a secondary page somewhere.  The Ubuntu instructions provide enough packages for a minimal build, but for KWIVER to be useful you really need VXL, OpenCV, Ceres, etc.   We don't want people trying this minimal build on Ubuntu first and then getting frustrated that it can't actually do anything useful.  Some packages, like VXL, don't have a modern enough version packaged for Ubuntu.

mattbrown11:
The updates still assume Fletch is the primary provider of packages.  However, I just added notes on how to install some additional dependencies that Fletch doesn't currently provide.

Perhaps it makes sense to move this into the doc/manuals sub-directory.  I didn't realize until later that there already is an install_guide.rst in that directory.  Perhaps we could point to the system-specific install guides from the main README.

hughed2:

hughed2:

hughed2:

mattbrown11:
What else needs to be done to finish this out?

mattbrown11:
@aaron-bray This is the KWIVER documentation branch I was working on.

mleotta:
This PR is now outdated and I'm going to close it.  The material added here is still useful but should be added to new restructured text file (not the main README) that focuses on building KWIVER against native Debian/Ubuntu packages.  The recommended approach is building against Fletch, but there are cases where building against system packages is needed.

PR #95, opened 2017-03-16
Assorted win32 and cross-platform fixes

mattdawkins:


arrows/core/detected_object_set_output_kw18.cxx
```diff
@@ -33,12 +33,17 @@
...
 #include <vital/util/tokenize.h>
 
-#include <atomic>
 #include <memory>
 #include <vector>
 #include <fstream>
 #include <time.h>
 
+#if (__GNUC__ == 4 && __GNUC_MINOR__ < 5 )
+  #include <cstdatomic>
```
mleotta: This is failing on Ubuntu with Clang.  See the Travis-CI builds.  Somehow it tries to include cstdatomic when building with Clang.  Maybe these GCC version variables are set if the system has GCC, even if you don't use it?

linus-sherrill: I believe this was added to support gcc 4.4 (from older CentOS). I don't believe it is needed for current systems.

mattdawkins: Clang also defines those flags, it can be fixed by adding a #ifndef __clang__ clause, however, in addition to the __GNUC__ checks

(I still like having it because everything else builds on C6 default gcc and I use it a bit)

mattdawkins: "ifndef __clang__"

CMake/kwiver-depends-CUDA.cmake
```diff
@@ -0,0 +1,11 @@
...
+
+option( KWIVER_ENABLE_CUDA
+  "Enable CUDA dependent code and plugins"
+  OFF
+  )
+
+if( KWIVER_ENABLE_CUDA )
+  find_package( CUDA REQUIRED )
+  include_directories(SYSTEM ${CUDA_INCLUDE_DIRS})
+endif( KWIVER_ENABLE_CUDA )
```
mattdawkins: Done

mattdawkins:
I removed most of the blanket python win32 ignores.

Also note that a lot of python stuff was straight up not working on windows previously due to pathing issues and other.

PR #112, opened 2017-05-01
Merge in viame/master (track oracle, detector, python and misc improvements)

mattdawkins:


mattdawkins:
Doesn't build due to issue on current master, I also still need to sort the commits on here to make sure they should all land

mattdawkins:
Note: the branch now passes the automated build procedure but has an error when python is turned on (but this also occurs on master branch which was merged in)

PR #115, opened 2017-05-15
Merge in darknet detector updates

mattdawkins:


vital/types/detected_object_set.cxx
```diff
@@ -126,6 +127,18 @@ add( detected_object_sptr object )
 
 
 // ------------------------------------------------------------------
+void
+detected_object_set::
+add( detected_object_set_sptr detections )
+{
+  VITAL_FOREACH( auto detection, detections->select() )
```
linus-sherrill: It would be more efficient to use:

    m_detected_objects.insert( m_detected_objects.end(), detections.begin(), detections.end() );

vital/types/detected_object_set.cxx
```diff
@@ -209,6 +222,42 @@ select( const std::string& class_name, double threshold )const
...
+scale( double scale_factor )
+{
+  if( scale_factor == 1.0 )
+  {
+    return;
+  }
+
+  VITAL_FOREACH( auto detection, m_detected_objects )
+  {
+    auto bbox = detection->bounding_box();
```
linus-sherrill: It might be useful to add an operator on a detected_object to scale the bbox. - e.g.

    void scale( detected_object& detection, double scale_factor )
    {
        auto bbox = detection.bounding_box();
        bbox = kwiver::vital::scale( bbox, scale_factor );
        detection.set_bounding_box( bbox );
    }

The same could be done for shift().

Similarly, these methods could be converted into operators so they are not part of the object.

hughed2:

hughed2:

hughed2:

PR #121, opened 2017-05-24
Fix how dict is accessed in order to return None default

Purg:
Also fixed some formatting for PEP8 compatibility.


Purg:
PTAL @linus-sherrill

PR #123, opened 2017-05-24
Set the bin install dir so that kwsys executables like dlls are installed.

billhoffman:
This makes sure that .dlls are installed.


vovythevov:
+1

PR #127, opened 2017-05-24
Fix vital image python method calls

as6520:


as6520:
Please take a look at the fix @Purg @linus-sherrill

hughed2:

PR #137, opened 2017-05-26
Hackathon - Add string vector type/port trait in sprokit

Purg:
Also reworded the description of the descriptor_set port trait since it
can and will be used for more than just feature-point descriptors. An
alternaltive is to name a separate type/port trait for other descriptor
types, but the existing one is general in name, so the description
should also be general.


Purg:
PTAL: @mleotta @linus-sherrill

hughed2:

hughed2:

hughed2:

PR #140, opened 2017-05-27
Add python traits and converters

Purg:
Add descriptor_set and string vector plumbing for python sprokit processes in anticipation for a process that converts descriptors into SMQTK format.


Purg:
PTAL @mattdawkins @linus-sherrill @mleotta

PR #141, opened 2017-05-31
Hackathon/master

linus-sherrill:
This is the set of collected changes from the hackathon.


doc/release-notes/1.1.0.txt
```diff
@@ -141,6 +141,9 @@ Arrows - Darknet
    real-time object detection system. Currently requires an external
    build of the Darknet project - https://gitlab.kitware.com/kwiver/darknet.git
 
+Arrows - Burnout
+
+ * Added burnout descriptors.
```
mleotta: @mattdawkins Release notes about this new burnout arrow could use more detail.  A user might want to know what are burnout descriptors?  What are they used for?  What dependences does this new arrow require?

PR #143, opened 2017-05-31
Moving UUID generation code to new arrow.

linus-sherrill:
Move uuid factory to separate arrow


Purg:
Any opinion on possibly naming the arrows module ``libuuid`` instead of ``uuid`` in order to differentiate this from the ``libuuid`` vs. ``uuid`` packages? (we had this confusion during the hackathon)

dstoup:
+1

PR #146, opened 2017-06-01
Publicly Release Changes from AFRL work

mleotta:
This branch contains all changes funded by AFRL and developed internally at Kitware.  These changes have been approved for public release on May 31st 2017.  Case Number: 88ABW-2017-2725 .


Purg:
Are the two failing tests expected / should that be fixed as an addition to this branch? I would imagine we should have a separate branch as this is a book keeping merge due to public release?

mleotta:
The two failing test seem to be only failing on the Clang build and relate to the new thread pool class.  I have a hunch that this is because I didn't add -lpthread to the Clang build flags.  Let me see if that fixes the problem.

mleotta:
Yup, that was it.

Purg:
+1

Obviously didn't look at everything, just passing tests mostly.

PR #154, opened 2017-06-08
Fix typo in test

hughed2:
The inversion shouldn't be there. Test will always fail as is, even if the environment is set properly.


mleotta:
Please direct bug fixes like these to the release branch.  Can you resubmit this with pull request target of `release`?

PR #155, opened 2017-06-08
Refactor tracks and track_set for feature/object tracks

mleotta:
This branch splits out feature track specific properties from the track object.  It does not create custom or templated `track` classes for different applications.  Rather the `track_state` is now a base class with minimal data.  You can create derived classes to put custom data in the track states.  See `feature_track_state`, for example.   The `track_set` class is now a base class that is mostly as it was but loses specific functions for querying feature track specific values.   Those specialized helpers have now moved to `feature_track_set`.  There is also a class for `object_track_set` and `object_track_state`.  All the relevant algorithms have been updated to use `feature_track_set` to indicate that they are designed to operate on feature tracks.


mleotta:
@mattdawkins here is what I have.  It compiles for me, but there are some test failure that need addressing.  Note that the track_state class now also contains a weak_ptr back to the track which contains it (if any).  This will be important later for more efficient manipulation of track data.  This requires that each track state belong to no more than one track. However, you can clone track_states if needed and those clones can share other data like descriptors.

mattdawkins:
I uploaded and tested all of the track and detections python wrappers, I didn't try turning on tests yet though but considering all of the detection, bounding box, etc... wrappers were previously broken and incompletely, the c bindings and python code is not going to hurt. As far as I'm concerned they're good to go.

mleotta:
This PR still needs release notes.   I'm working on that.  Please don't merge this (any other PR) if it doesn't properly update release notes.

vital/bindings/c/algo_def/bundle_adjust.cxx
```diff
@@ -64,7 +64,8 @@ vital_algorithm_bundle_adjust_optimize( vital_algorithm_t *algo,
     auto lmap_sptr = vital_c::LANDMARK_MAP_SPTR_CACHE.get( *lmap );
     auto tset_sptr = vital_c::TRACK_SET_SPTR_CACHE.get( tset );
 
-    a_sptr->optimize( cmap_sptr, lmap_sptr, tset_sptr );
+    a_sptr->optimize( cmap_sptr, lmap_sptr,
+      std::dynamic_pointer_cast< kwiver::vital::feature_track_set >( tset_sptr ) );
```
mleotta: I'm wondering if the C API should have separate types for `feature_trackset`, `object_trackset`, and the base `trackset`.

mattdawkins: I agree, but I was going to do this in a follow-up branch which also adds an object_track derived from track (since object_track needs to contain a detected_object_type which a feature_track doesn't)

vital/bindings/c/types/bounding_box.cxx
```diff
@@ -102,7 +120,7 @@ void vital_bounding_box_destroy( vital_bounding_box_t* bbox)
 TYPE vital_bounding_box_ ## NAME( vital_bounding_box_t *bbox )          \
 {                                                                       \
   STANDARD_CATCH(                                                       \
-    "C::bounding_box::" # NAME, 0,                                      \
```
mleotta: This name change is inconsistent with the rest of Vital.  I'm not arguing that it makes total sense, but other classes use `C::class_name::` to make it clear that the exception is coming out of the C bindings layer.  If we are changing that convention I think it needs to be done globally, and I would not do that on this PR.

vital/bindings/c/types/detected_object.cxx
```diff
@@ -58,17 +58,36 @@ vital_detected_object_t* vital_detected_object_new_with_bbox( vital_bounding_box
                                                               vital_detected_object_type_t* dot)
 {
   STANDARD_CATCH(
-    "C::detected_object:new", 0,
-    auto dotsp = kwiver::vital_c::DOT_SPTR_CACHE.get( dot );
+    "vital_detected_object_new_with_bbox", 0,;
```
mleotta: see other comment about changing the `C::` prefix

vital/bindings/c/types/bounding_box.cxx
mleotta: @Purg do you have any thoughts on this convention?  Apparently both `C::class::function` and `vital_class_function` are in use in different places in the `STANDARD_CATCH` macro.  I think we should have a follow-on branch to standardize these, but I don't know which way we should go.   Thoughts?  Maybe a hybrid like `[C API] vital_class_function`?

mleotta:
This is a rather large branch that both @mattdawkins and myself have contributed to.  It touches many files out of necessity since it refactors a commonly used data type.  I think we are both mostly satisfied with it.  I don't expect anyone else will have time to do a review so I'm planning to merge it unless someone else would like to review it.  The only Travis CI test failures seem to be coming from a time out because Fletch + KWIVER takes too long to build.

PR #161, opened 2017-06-09
Fix a Visual Studio warning about an extra ';'

mleotta:
These macros expand to something like

    if(var == NULL);

Visual Studio doesn't like conditionals followed by a ';' so
this commit uses '{}' instead.  While this works, it does not seem
to be using the macro as intended, a better fix would actually
do something in the '{}', like throw an exception.

### DIFF
--- a/vital/bindings/c/types/landmark.cxx
+++ b/vital/bindings/c/types/landmark.cxx
@@ -235,7 +235,7 @@ vital_landmark_##S##_set_loc( vital_landmark_t *l, \
     REINTERP_TYPE( matrix_t const, loc, loc_ptr ); \
     TRY_DYNAMIC_CAST( vital::landmark_<T>, \
                       vital_c::LANDMARK_SPTR_CACHE.get( l ).get(), \
-                      l_ptr ); \
+                      l_ptr ){} \
     l_ptr->set_loc( *loc_ptr ); \
   ); \
 } \
@@ -249,7 +249,7 @@ vital_landmark_##S##_set_scale( vital_landmark_t *l, T scale, \
     "vital_landmark_" #S "_set_scale", eh, \
     TRY_DYNAMIC_CAST( vital::landmark_<T>, \
                       vital_c::LANDMARK_SPTR_CACHE.get( l ).get(), \
-                      l_ptr ); \
+                      l_ptr ){} \
     l_ptr->set_scale( scale );\
   ); \
 } \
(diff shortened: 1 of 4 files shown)


mleotta:
@Purg I think you wrote this code but @linus-sherrill wrote the `TRY_DYNAMIC_CAST` macro that is being used.  This branch fixes a warning, but I think the warning was indicative of something deeper being wrong.  I don't think the `TRY_DYNAMIC_CAST` macro is being used properly here.  Nothing happens if the dynamic_cast fails.  There is no exception or error message, it just continues to use the null pointer and will crash.

I don't particularly like the `TRY_DYNAMIC_CAST` macro as written.  It stops in the middle of an expression and expects the user to complete the expression after the macro.  Most of the other macro are self contained expressions, so this one is different and likely to be misused.  It would be better if `TRY_DYNAMIC_CAST` throws an exception when the cast fails.

linus-sherrill:
It looks like most of these changes are in a STANDARD_CATCH block, so throwing an exception would be expected. Maybe an another macro could be written that just throws if the dynamic cast fails. This would save repetitive coding of the throw in that file.

linus-sherrill:
I'm not sure I actually wrote that macro. I may have manually merged it from another branch.

Purg:
I think I originally wrote that macro, when vital was a separate repo.

The original intention of the macro was to defer specific handling of a failed cast to the caller, letting them define what happens in the `{}` block:
```
TRY_DYNAMIC_CAST( Foo_sptr, foo, new_var_name )
{
    // cast failed block, respond here in context specific way.
}
```

In other C binding contexts, I have used it for C-binding specific error handling like in following code in `vital/bindings/c/types/descriptor.cxx` and `vital/bindings/c/types/feature.cxx`:
```
...
TRY_DYNAMIC_CAST( vital::descriptor_dynamic<T>, d_sptr.get(), d_dyn )
{
  POPULATE_EH( eh, 1, "Failed to dynamic cast descriptor to " #S " type "
                      "for data access" );
  return 0;
}
...
```

In this specific case, I might argue that some `throw`'ing or error handle population should be occurring here. The current landmark C-binding code looks like it will segfault if you give it a handle pointer that cannot be converted to the requested type.

I concede that this macro many not be the most useful form of this and that the macro's naming may not be particularly reflective of its current function. The macro could easily be changed to throw a specific exception on a failed dynamic cast. If a user then wanted to do something specific on a failed cast (like populating the vital_error_handle with a specific error code than the default of -1), then they would need to wrap the call in a try-catch, which I what I think was trying to avoid with this macro for my specific opinion of convenience.

Purg:
I will add a commit in a minute filling error handling within the `{}` to correctly populate the error handle if the cast failed, setting an error code value of 1. Will update landmark.h doxygen appropriately.

PR #162, opened 2017-06-12
Fix python tests

Purg:
Fix VITAL python tests (`./vital/bindings/python/run_tests.sh`) that are failing since the last time they were checked.

Really, these tests need to be run via CTest tests, but that is out of the scope of this particular branch.

There is still one failing python test with the `Image` binding when converting from/to PIL images when the type is boolean (PIL image mode "1"). PIL uses a 1-bit encoding for memory storage where-as kwiver/vital uses 1 **byte** per boolean value. Thus, just extracting the vital Image memory buffer and sticking it into a mode "1" PIL image views incorrect results. Transitively, going back and forth between from-PIL/to-PIL functions compounds the issue.

### DIFF
--- a/vital/bindings/python/vital/tests/test_rotation.py
+++ b/vital/bindings/python/vital/tests/test_rotation.py
@@ -44,7 +44,7 @@ from vital.types import Rotation
 
 
 def array_normalize(a, dtype=None):
-    a = numpy.asarray(a, None)
+    a = numpy.asarray(a, dtype)
     return a / numpy.linalg.norm(a)
 
 
--- a/vital/bindings/python/vital/tests/test_rotation.py
+++ b/vital/bindings/python/vital/tests/test_rotation.py
@@ -43,6 +43,11 @@ import numpy
 from vital.types import Rotation
 
 
+def array_normalize(a, dtype=None):
+    a = numpy.asarray(a, None)
+    return a / numpy.linalg.norm(a)
+
+
 class TestVitalRotation (unittest.TestCase):
 
     def test_new_default(self):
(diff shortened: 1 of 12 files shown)


Purg:
PTAL @mleotta @mattdawkins @linus-sherrill

vital/bindings/python/vital/tests/test_rotation.py
```diff
@@ -43,6 +43,11 @@
 from vital.types import Rotation
 
 
+def array_normalize(a, dtype=None):
+    a = numpy.asarray(a, None)
```
mleotta: `dtype` is not used, should this `None` actually be `dtype`?

Purg: Oops.

mleotta:
Looks good to me. 👍

mleotta:
I tested this and found some additional issues.  First I didn't have PIL installed so I got errors for all six PIL tests.  Is PIL a hard requirement?  Is it appropriate for tests to fail if you don't have PIL, or should they just not be run?  I'm guessing if from_pil and to_pil are members of the image class then we are assuming that PIL is a requirement.

After I installed PIL (from mac ports, PIL 1.1.7_7) the PIL tests still all fail, now with

    AttributeError: 'module' object has no attribute 'frombytes'

I assume `frombytes` is in a newer version of PIL than I have?

Purg:
@mleotta @linus-sherrill Added an initial `requirements.txt` file for at least documenting what other python modules are required for using the vital python bindings. Version requirements are up for debate. Everything except pillow is the current version that are passing on my system. The Pillow version is a little behind the current version due to a bug I remember seeing a while ago (though I don't remember what that bug was...).

mleotta:
needs the equivalent fix for `frombytes`

mleotta:
Works for me!  And all tests pass.  Apparently that boolean image issue only exists with newer versions of PIL.

Purg:
Well damn.

Purg:
:(

Purg:
That means that supporting old/new PIL/Pillow will be a hassle.

mleotta:
We can figure that out in the future, on another branch.  I'm taking what you have here for now.

PR #166, opened 2017-06-22
Dev/track oracle windows fixes

collinsr:
Strengthen previous gcc5 fixes for template metaprogramming issues.

This builds for me on vs2017.

### DIFF
--- a/track_oracle/core/kwiver_io_base.txx
+++ b/track_oracle/core/kwiver_io_base.txx
@@ -30,10 +30,18 @@ typedef char (&no_io_type)[2];
 typedef char yes_io_type;
 
 struct proxy{ template <typename U> proxy(const U&); };
+# if defined(_MSC_VER)
+#  pragma warning( push )
+#  pragma warning( disable : 4172 ) /* returning address of local variable */
+# endif
 
-no_io_type operator>>( const proxy&, const proxy& ) {}
+no_io_type operator>>( const proxy&, const proxy& ) { char tmp[2]; no_io_type not_used(tmp); return not_used; }
+no_io_type check( no_io_type ) { char tmp[2]; no_io_type not_used(tmp); return not_used; }
 
-no_io_type check( no_io_type ) {}
+# if defined(_MSC_VER)
+#  pragma warning( pop )
+# endif
+
 template <typename U> yes_io_type check( const U& );
 
 
--- a/track_oracle/core/kwiver_io_base.txx
+++ b/track_oracle/core/kwiver_io_base.txx
@@ -30,17 +30,11 @@ typedef char (&no_io_type)[2];
 typedef char yes_io_type;
 
 struct proxy{ template <typename U> proxy(const U&); };
-# if defined(_MSC_VER)
-#  pragma warning( push )
-#  pragma warning( disable : 4172 ) /* returning address of local variable */
-# endif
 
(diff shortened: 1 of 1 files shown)


track_oracle/core/kwiver_io_base.txx
```diff
@@ -30,10 +30,18 @@ typedef char (&no_io_type)[2];
 typedef char yes_io_type;
 
 struct proxy{ template <typename U> proxy(const U&); };
+# if defined(_MSC_VER)
+#  pragma warning( push )
+#  pragma warning( disable : 4172 ) /* returning address of local variable */
```
dstoup: Is returning the address of a local variable a good thing to do?

collinsr: Generally, of course, no, but this code is used as part of the template metaprogramming for I/O routing inside track_oracle and as such should never actually be generated, only used to steer the template instantiations[1]. Originally, the functions had no definitions, just declarations; but gcc5 complained, so an earlier patch added empty implementations. Then visual studio complained that the empty implementations returned nothing, so hence this patch.

[1] Caveat: Not claiming to be a TMP pro. Haven't checked that this code isn't generated. But I am sure it's never used for anything.

mleotta: I'm wondering if other compilers will generate similar warnings.  Regardless of whether the function is actually every called or not, is there a better way to implement it to avoid this warning in the first place?  I suppose one option is to make the dummy variable global.  Does `no_io_type` have to be a reference for this to work?  I haven't fully grasped what the meta-program is doing so I don't have any other suggestions off the top of my head.

collinsr: Off the top of my head, I'm not sure why `no_io_type` is a reference. It's certainly not what I would've tried first. Everywhere else in track_oracle, I think, yes-types are a `char` and no-types are something whose `sizeof()` is guaranteed to be different than yes-type, usually `char[2]`. So if it's a reference here, it was to make something work, but the details escape me at the moment.

That said, I'm certain it would all benefit from a refactoring, which I would love to do when/if I get a chance.

(The background in case anybody cares: track_oracle reads and writes types to both XML and CSV files, see `kwiver_io_base.txx`. Ideally this would also extend to e.g. json in the future. Whenever possible, it wants to use the same method for both: `from_str()` for reading and `to_stream()` for writing. Again whenever possible, it wants to use the type's native `op>>` and `op<<`, so that when you add a type, everything 'just works' and no extra code needs to be written to support the type. **However**, and here's where my memory starts to get fuzzy, the XML reading uses tinyxml, which ends up being string-token based (which is why it reads from a string but outputs to a stream.) But this has problems if the `op>>` representation contains spaces. (For example, vidtk's old timestamp class.) So another layer of indirection is added, defining `kwiver_read` and `kwiver_write` to either use our local space-less methods (if defined in  `kwiver_io_helpers.h`) or the type's `op>>` and `op<<` (if not defined in `kwiver_io_helpers.h`.))

Like I said, I'd love to refactor it, both to simplify this and to make the functionality more generally available in kwiver.

mleotta:
@collinsr I think this is ready to merge, but we need to rebase this on `release` and submit a PR to `release` instead of `master`.  In the interest of time, I may just do this myself and create a new PR

PR #169, opened 2017-06-27
Minor CMake Options Clean Up

mleotta:
These changes are intended to make the KWIVER configuration easier and more clear to new users.

### DIFF
--- a/CMake/kwiver-depends-PROJ.cmake
+++ b/CMake/kwiver-depends-PROJ.cmake
@@ -5,6 +5,11 @@ option( KWIVER_ENABLE_PROJ
   OFF
   )
 
+# Fletch provides the PROJ4 symbols. We need the PROJ symbols.
+if (PROJ4_ROOT AND NOT PROJ_ROOT)
+  set(PROJ_ROOT "${PROJ4_ROOT}")
+endif()
+
 if( KWIVER_ENABLE_PROJ )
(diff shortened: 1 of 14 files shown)


CMake/kwiver-depends-PROJ.cmake
```diff
@@ -1,10 +1,15 @@
...
 
 option( KWIVER_ENABLE_PROJ
-  "Enable PROJ dependent code and plugins"
-  OFF
+  "Enable PROJ dependent code and plugins (Arrows)"
+  ${fletch_ENABLED_PROJ4}
   )
 
+# Fletch provides the PROJ4 symbols. We need the PROJ symbols.
+if (PROJ4_ROOT AND NOT PROJ_ROOT)
```
mleotta: There is a good chance we don't need this at all anymore.  I was mostly relocating this code here from the main CMakeLists.txt, where it certainly doesn't belong (see removed code below).  This is here because of our/their inconsistent naming of PROJ vs PROJ4.  At one point Fletch exported PROJ4_ROOT and not PROJ_ROOT, so a fix was added to copy from one to the other.  Now I believe Fletch exports both PROJ_ROOT and PROJ4_ROOT, so we could consider removing this fix.

mleotta:
Does anyone see any issues with this branch, or should I just merge it?

PR #170, opened 2017-06-28
MSVC gprof check and use the standard sleep 

aaron-bray:
Replacing the unistd.h with something more C++ spec compliantCMAKE found a Perl version of gprof on my machine and made the gprof projects, I don't think these should be attempted if using MSVC, it has it's own profiler you can use
Maybe we can turn this on for msvc users in the future if there is really a need

### DIFF
--- a/sprokit/cmake/support/test.cmake
+++ b/sprokit/cmake/support/test.cmake
@@ -53,7 +53,9 @@ if (VALGRIND_EXECUTABLE)
   _sprokit_declare_tool_group(bbv)
 endif ()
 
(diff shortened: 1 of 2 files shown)


mleotta:
Looks good to me

mleotta:
Oops this should have gone into release.  I'll try to fix that manually

aaron-bray:
how about you just cancel it and I do another PR, to the right place this
time :(

mleotta:
I just bypassed the PR and manually merged it.  It should be all good now.

PR #172, opened 2017-06-28
Include template definitions

aaron-bray:
MSVC will issue warnings if it compiles a file that uses templates but does not include its definitions


mleotta:
Looks okay to me.

@collinsr do you see any issue with this?  I might just merge anyway and revert if it causes trouble.

PR #175, opened 2017-06-29
Log4cplus is working

aaron-bray:
There is already a log4cplusConfig.cmake, so we don't need a FindLog4cplus.cmake
Changed things to all lowercase
Updated the vital logging classes to load and use the log4cplus plugin


mleotta:
Can you also update `log4cxx_factory.cxx` in the corresponding way?  That is, include the appropriate exports.h files and use the correct EXPORT macro.  I know you can't actually build that one now, but we know at least those changes will be needed at some point if someone builds with log4cxx on Windows.

PR #176, opened 2017-06-29
Logger fixes

aaron-bray:


mleotta:
looks good to me

vital/logger/kwiver_logger_factory.h
```diff
@@ -48,7 +48,7 @@ namespace logger_ns {
  * execution (i.e. static initializer time), which is before the
  * initialize method is called.
  */
-class kwiver_logger_factory
+class VITAL_LOGGER_EXPORT kwiver_logger_factory
```
linus-sherrill: I'm not sure this should be exported.

CMake/FindLog4cplus.cmake
```diff
@@ -1,28 +0,0 @@
...
-  find_path( Log4cplus_INCLUDE_DIR log4cplus/logger.h ${Log4cplus_FIND_OPTS})
-  find_library( Log4cplus_LIBRARY log4cplus ${Log4cplus_FIND_OPTS})
-  restore_find_root_context(Log4cplus)
-
-  include( FindPackageHandleStandardArgs )
-  FIND_PACKAGE_HANDLE_STANDARD_ARGS( Log4cplus Log4cplus_INCLUDE_DIR Log4cplus_LIBRARY )
-  if( LOG4CPLUS_FOUND )
-    set( Log4cplus_FOUND TRUE )
-  endif()
-endif()
```
linus-sherrill: Will we still find the files if the package is not supplied by fletch.

aaron-bray:
I had to export it since it is used in the log4cplus_factory.cxx

*class log4cplus_factory*
*  : public kwiver_logger_factory*

Previous to my changes, each logger compiled the same set of files
I changed the cmake projects so vital_logger compiled those files only, and
log4cplus linked with vital_logger
if it contained the same file set, you run into issues where the plugin
projects had to include both the vital_logger_exports and their exports
(vital_log4cplus_logger_exports)

it seemed to me that if multiple projects needed the same files, just put
them in the base project and link to it
since i had to link to it, i had to export this class

aaron-bray:
That is true, this is probably a good back up for not having it in fletch
but being somewhere cmake discoverable
but if it is not supplied by fletch, i think you should be specifying the
log4cplus_DIR cmake variable, which should point to it's
log4cplusConfig.cmake
that is probably what you would want to do on windows, but not linux, I
think it's easier for cmake to find things on linux.
you never know where things are on windows if it's not on the path

at the very least, i think we would need to rename it Findlog4cplus,
log4cplus uses all lowercase in its config and directory names
this does make sense to be in

PR #177, opened 2017-06-30
corrected misplaced space

mleotta:
This was causing build errors


mattdawkins:
This is a change I can get behind.

PR #178, opened 2017-06-30
Fixed a logger exception from assigning a null pointer to a string

mleotta:
Also cleaned up the code a little to avoid the need for ifdefs

### DIFF
--- a/vital/logger/kwiver_logger_manager.cxx
+++ b/vital/logger/kwiver_logger_manager.cxx
@@ -101,34 +101,24 @@ kwiver_logger_manager
   // we provide a method for creating these loggers.
 
   bool try_default(false);
-  std::string factory = std::getenv( PLUGIN_ENV_VAR );
-  if ( factory.empty() )
+  std::string factory_name;
+  char const* factory = std::getenv( PLUGIN_ENV_VAR );
+  if ( factory == 0 )
   {
-    try_default = true;
     // If no special factory is specified, try default name
-#if defined(WIN32)
-    factory = "vital_logger_plugin.dll";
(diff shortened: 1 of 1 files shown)


vital/logger/kwiver_logger_manager.cxx
```diff
@@ -101,34 +101,24 @@ ::kwiver_logger_manager()
   // we provide a method for creating these loggers.
 
   bool try_default(false);
-  std::string factory = std::getenv( PLUGIN_ENV_VAR );
```
mleotta: @aaron-bray This line was causing segfaults when the environment variable is not set.  The results was a large number of failing unit tests.  We need to check that the return value is not null before assigning the `char*` to a `std::string`.

vital/logger/kwiver_logger_manager.cxx
```diff
@@ -101,34 +101,24 @@ ::kwiver_logger_manager()
...
-#if defined(WIN32)
-    factory += ".dll";
-#elif defined(__APPLE__)
-    factory += ".so";
-#else
-    factory += ".so";
-#endif
+    factory_name = factory;
   }
+  factory_name += DL::LibExtension();
```
mleotta: I also factored out all the `#ifdef`s by letting the dynamic loader tell me what the appropriate extension should be.  Let me know if this doesn't work for you.

aaron-bray:
Yeah, That makes sense
My bad

Thanks for finding this

aaron-bray:
That dynamic loader class is pretty cool! It handles a lot of the leg work in a generic way!
I think I may use it for other projects!

PR #179, opened 2017-07-03
Update release notes

mleotta:
In addition to fixing typos, this branch adds a detailed summary of the major changes in this release.  Please read through this and suggest changes.  I'm sure I don't have the full picture.

### DIFF
--- a/doc/release-notes/1.1.0.txt
+++ b/doc/release-notes/1.1.0.txt
@@ -2,7 +2,68 @@ KWIVER v1.1.0 Release Notes
 ===========================
 
 This is a minor release of KWIVER that provides both new functionality and
-fixes over the previous v1.0.0 release.
+fixes over the previous v1.0.0 release.  There are numerous changes in this
+release that have been motivated primarily by three different applications
+of KWIVER.  One is the use of KWIVER in MAP-Tk to provide 3D reconstruction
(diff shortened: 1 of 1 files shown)


mleotta:
@kfieldho, @collinsr, do you have any comment on the accuracy and completeness of what I have written for release notes?

collinsr:
Line 24: 'that allows introspect of these plugins' --> introspection

Otherwise, looks good!

collinsr:
Although there's apparently no contractural requirement (at least on DIVA's part), and arguably redundant to what's in the bulk of the release notes, I think it would be fitting to add an "acknowledgements" section where we explicitly call out that much of this release was funded by DIVA (and the other projects.)

mleotta:
@collinsr Do you think this should be acknowledgements in the release notes, or an acknowledgements section in the readme?  For example MAP-Tk has such an acknowledgements section in the readme: https://github.com/kitware/maptk#acknowledgements.  I probably should copy a similar AFLR acknowledgement to KWIVER as well since much of the MAP-Tk code has moved into KWIVER.  We could add IARPA, NOAA, and others here as well.

collinsr:
Ah-- I think the acknowledgements section is fine.

mleotta:
I'm going to merge this and we can revise further, if needed, on another branch.

PR #186, opened 2017-07-06
Move documentation files into the main doc directory.

linus-sherrill:
Move doc files into main doc directory. The format and content still need to be reviewed.


mleotta:
Looks like a step in the right direction to me.  My only suggestion is that maybe you retain the "sprokit" or "vital" name in the filenames or path so it is more clear which part of kwiver these files are documenting.  For example `doc/manuals/sprokit/pipeline_design.md` or `doc/manuals/sprokit_pipeline_design.md`

Issue #187, opened 2017-07-07
Current set of problems with Python tests


linus-sherrill:
Python tests are uncovering a set of issues.
--------
Test:
/usr/bin/python "/disk2/projects/KWIVER/kwgitlab/build/sprokit/tests/bin/test-python-process_registry" "register_cluster" "." "/disk2/projects/KWIVER/kwgitlab/build/lib/python2.7/dist-packages"

Error: Could not create newly registered process cluster type: No registered converter was able to produce a C++ rvalue of type std::shared_ptr<sprokit::process> from this Python object of type PythonBaseClusterExample

Problem:
It is not clear how to add this converter.

This error appears in several tests.

--------
Test:
/usr/bin/python "/disk2/projects/KWIVER/kwgitlab/build/sprokit/tests/bin/test-python-process_cluster" "api_calls" "." "/disk2/projects/KWIVER/kwgitlab/build/lib/python2.7/dist-packages"

217: Error: Unexpected exception: Python argument types in
217:     sprokit.pipeline.process_cluster.cluster_from_process(BaseProcess)
217: did not match C++ signature:
217:     cluster_from_process(std::shared_ptr<sprokit::process> process):
217:   File "/disk2/projects/KWIVER/kwgitlab/build/lib/python2.7/dist-packages/sprokit/test/test.py", line 92, in run_test
217:     tests[testname](*args, **kwargs)
217:   File "/disk2/projects/KWIVER/kwgitlab/build/sprokit/tests/bin/test-python-process_cluster", line 69, in test_api_calls
217:     if process_cluster.cluster_from_process(p) is not None:

Problem:
converting from python.BaseProcess -> sprokit::process_t (which is std::shared_ptr<propcess>)
does not appear to be working.

This error happens in several tests.

--------
Test:
/usr/bin/python "/disk2/projects/KWIVER/kwgitlab/build/sprokit/tests/bin/test-python-process_registry" "register" "." "/disk2/projects/KWIVER/kwgitlab/build/lib/python2.7/dist-packages"

221: 2017-07-07 09:40:02.350 DEBUG plugin_manager.cxx(184): Adding path(s) "/disk2/projects/KWIVER/kwgitlab/build/lib/modules:/disk2/projects/KWIVER/kwgitlab/build/lib/modules:/disk2/proje
[...]

PR #188, opened 2017-07-07
Fix some python tests.

linus-sherrill:
Fixes for a few python tests but there are still some tests that fail. See #187


mleotta:
@linus-sherrill Please rebase this change on `release` so we can merge into the release branch.

PR #192, opened 2017-07-10
Fixes for windows

aaron-bray:
Fixes an issue with the msvc compiler not able to compile some python binding code without some preprocessing (needs /Ob2 flag set for debug)(it is now set to this for ALL configurations when using MSVC)

Shortening the python scheduler target library name, I chose to shorten by removing 'scheduler' from the end of the target name (it's also in the beginning of the name, twice seemed redundant)

### DIFF
--- a/sprokit/conf/sprokit-macro-python.cmake
+++ b/sprokit/conf/sprokit-macro-python.cmake
@@ -117,6 +117,13 @@ function (sprokit_add_python_library    name    modpath)
 
   sprokit_add_library("python-${safe_modpath}-${name}" MODULE
     ${ARGN})
+  
+ if(MSVC)
+  # Issues arise with the msvc compiler with some projects where it cannot 
+  # compile bindings without the optimizer expanding some inline functions (i.e. debug builds)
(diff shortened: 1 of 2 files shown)


mleotta:
Looks good to me.  Assuming the Travis-CI build passes and there are no further comments I will merge this for testing on the dashboard tonight

aaron-bray:
For the python scheduler target naming change,
I don't know if there is any 'reflection' type stuff going on with naming conventions that something would expect... or if this would affect any end-user python code, such as maybe an import would have to change? I am not up to snuff on python bindings

The other approach I saw was to change configure to config, but that would affect more than just the python naming, and would also require multiple changes/attempts to get everything back to being in sync

I am open to other ideas, but these are the two options I saw

mleotta:
I don't think anyone is actually using the python scheduler yet, but I don't know for sure.  @linus-sherrill @mattdawkins do either of you know?

aaron-bray:
The switch flag was put in by another PR
The python project name change is still in the works and will be part of a different PR

PR #199, opened 2017-07-11
Fixed inconsistent DLL linkage on MSVC

mleotta:
There were multiple source files recompiled into a plugin
and a library.  This change removes the duplicates from the
plugin and make it link to the library instead.

### DIFF
--- a/sprokit/processes/adapters/CMakeLists.txt
+++ b/sprokit/processes/adapters/CMakeLists.txt
@@ -3,35 +3,6 @@
 #
 project( kwiver_adapter_processes )
 
-set( proc_sources
-  register_processes.cxx
-
-  adapter_data_set.cxx   adapter_data_set.h
-  adapter_base.cxx
-  input_adapter_process.cxx
-  output_adapter_process.cxx
-  )
-
-set( private_headers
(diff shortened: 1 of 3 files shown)


mleotta:
@aaron-bray This PR resolves the DLL linkage issues I told you about

aaron-bray:
Cool, that makes sense.
I had not gotten to it yet, but that looks like it would cause the issue and is the right fix!

sprokit/processes/adapters/CMakeLists.txt
```diff
@@ -42,18 +13,23 @@ set( lib_sources
...
-  output_adapter_process.h   output_adapter_process.cxx
+  input_adapter_process.cxx
+  output_adapter_process.cxx
   )
 
 set( public_headers
   adapter_types.h
   adapter_data_set.h
   input_adapter.h
+  input_adapter_process.h
```
linus-sherrill: these in/output_adapter_process headers do not need to be installed.

mleotta: I'm going to merge this for testing on the dashboard tonight, then I can revisit making these headers private.

PR #202, opened 2017-07-11
Add environment specified path when loading plugin_explorer plugins.

linus-sherrill:
The plugin explorer did not look at the configured environment path
when loading plugin explorer extensions.


mleotta:
is this not covered by the code in `vital/plugin_loader/plugin_manager.cxx`?

PR #211, opened 2017-07-13
Remove use of fletch_ROOT variable.

mleotta:
The fletch_ROOT variable has been removed from Fletch since its value is
now the same as fletch_DIR, so we should use fletch_FOUND to detect if
Fletch was found and then use fletch_DIR for the Fletch install path.


CMakeLists.txt
```diff
@@ -224,8 +226,8 @@ if ( KWIVER_ENABLE_MATLAB )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export LD_LIBRARY_PATH=${Matlab_LIBRARY_DIR}:$LD_LIBRARY_PATH\n" )
 endif()
 
-if ( fletch_ROOT )
-  file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export LD_LIBRARY_PATH=${fletch_ROOT}/lib:$LD_LIBRARY_PATH\n" )
+if ( fletch_FOUND )
+  file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export LD_LIBRARY_PATH=${fletch_DIR}/lib:$LD_LIBRARY_PATH\n" )
```
linus-sherrill: The fletchConfig.cmake file exports fletch_ROOT to point to the install directory. The fletch_DIR points to the directory that contains fletchConfig.cmake and is used by find_module(). These two variables may point to the same directory, but they do not have to. I think adding fletch_ROOT to the paths is more robust.

mleotta: That is no longer true.  If you update and rebuild your Fletch you'll see that fletch_ROOT no longer exists.  fletchConfig.cmake has been moved into the install dir so now fletch_DIR must point to the install dir instead of the build dir and it serves both purposes.

I think that change in Fletch makes sense, but it would have been better if it was done more gradually to prevent things from breaking.  Oh well.

linus-sherrill: I see that fletch_ROOT has been removed. There must have been some reason for that, but all other packages have a ROOT specification.

PR #212, opened 2017-07-13
Create a batch script to configure the environment on windows

aaron-bray:
Also replaced fletch_ROOT with the user provided fletch_DIR

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,18 +36,17 @@ set(KWIVER_BINARY_DIR           "${CMAKE_BINARY_DIR}")
 #
 # kwiver module config file.
 set(KWIVER_CONFIG_FILE          "${KWIVER_BINARY_DIR}/kwiver-config.cmake")
-# Create initial setup script
+# Create initial setup shell script
 set(KWIVER_SETUP_SCRIPT_FILE    "${KWIVER_BINARY_DIR}/setup_KWIVER.sh" )
-if(WIN32)
-  set(KWIVER_SETUP_BATCH_FILE    "${KWIVER_BINARY_DIR}/setup_KWIVER.bat" )
-  if(CMAKE_CXX_COMPILER_ID MATCHES MSVC)
-   if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.10)
-      set(_vcVersion vc15)
-    elseif(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19)
-      set(_vcVersion vc14)
-    else()
-      message(FATAL_ERROR "KWIVER requires Visual Studio 2015 or greater")
-    endif()
+# Create initial setup batch script
+set(KWIVER_SETUP_BATCH_FILE    "${KWIVER_BINARY_DIR}/setup_KWIVER.bat" )
+if(CMAKE_CXX_COMPILER_ID MATCHES MSVC)
+ if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.10)
+    set(_vcVersion vc15)
+  elseif(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19)
+    set(_vcVersion vc14)
+  else()
+    message(FATAL_ERROR "KWIVER requires Visual Studio 2015 or greater")
   endif()
 endif()
 
@@ -194,16 +193,16 @@ configure_file(
   ${KWIVER_SETUP_SCRIPT_FILE}
   @ONLY
   )
-if(WIN32)
-  configure_file(
-    ${KWIVER_CMAKE_DIR}/setup_KWIVER.bat.in
-    ${KWIVER_SETUP_BATCH_FILE}
-    @ONLY
-    )
-endif()
+
+configure_file(
+  ${KWIVER_CMAKE_DIR}/setup_KWIVER.bat.in
+  ${KWIVER_SETUP_BATCH_FILE}
+  @ONLY
+  )
+
 
 # install set up script
-option( KWIVER_INSTALL_SET_UP_SCRIPT "Whether or not to install setup_KWIVER.sh" ON )
+option( KWIVER_INSTALL_SET_UP_SCRIPT "Creates a setup_KWIVER script (.sh and .bat) that will add properly add kwiver to a shell/cmd prompt" ON )
 mark_as_advanced( KWIVER_INSTALL_SET_UP_SCRIPT )
 
 if( KWIVER_INSTALL_SET_UP_SCRIPT )
@@ -216,10 +215,9 @@ if( KWIVER_INSTALL_SET_UP_SCRIPT )
 endif()
 
 if ( fletch_FOUND )
-  if(WIN32)
-    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/bin\n" )
-    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/x64/${_vcVersion}/bin\n" )
-  endif()
+  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/bin\n" )
+  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/x64/${_vcVersion}/bin\n" )
+
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export LD_LIBRARY_PATH=${fletch_DIR}/lib:$LD_LIBRARY_PATH\n" )
 endif()
 
@@ -244,10 +242,9 @@ install( FILES log4cplus.properties   DESTINATION ${CMAKE_INSTALL_PREFIX} )
 if ( KWIVER_ENABLE_LOG4CPLUS )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export VITAL_LOGGER_FACTORY=$this_dir/lib/modules/vital_log4cplus_logger\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export LOG4CPLUS_CONFIGURATION=$this_dir/log4cplus.properties\n" )
-  if(WIN32)
-    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set VITAL_LOGGER_FACTORY=vital_log4cplus_logger\n" )
-    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set LOG4CPLUS_CONFIGURATION=%~dp0/log4cplus.properties\n" )
-  endif()
+
+  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set VITAL_LOGGER_FACTORY=vital_log4cplus_logger\n" )
+  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set LOG4CPLUS_CONFIGURATION=%~dp0/log4cplus.properties\n" )
 endif()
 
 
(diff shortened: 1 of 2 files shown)


aaron-bray:
So I think we can always output the .sh file, even on windows and running msvc you may be running inside a cygwin shell and still want to use the .sh, but you will also get a bat file on windows.
This should support cmd shell users and bash shell users on windows

CMakeLists.txt
```diff
@@ -190,6 +208,18 @@ mark_as_advanced( KWIVER_INSTALL_SET_UP_SCRIPT )
 if( KWIVER_INSTALL_SET_UP_SCRIPT )
   install( PROGRAMS   ${KWIVER_SETUP_SCRIPT_FILE}
     DESTINATION ${CMAKE_INSTALL_PREFIX} )
+  if(WIN32)
+    install( PROGRAMS   ${KWIVER_SETUP_BATCH_FILE}
+      DESTINATION ${CMAKE_INSTALL_PREFIX} )
+  endif()
+endif()
+
+if ( fletch_DIR )
```
mleotta: You might want to use `fletch_FOUND` here instead.  The user can set `fletch_DIR` and it might be set but invalid.  `fletch_FOUND` indicates that `find_package` found what it was looking for.

CMakeLists.txt
```diff
@@ -190,6 +208,18 @@ mark_as_advanced( KWIVER_INSTALL_SET_UP_SCRIPT )
...
+  if(WIN32)
+    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/bin\n" )
+    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/x64/${_vcVersion}/bin\n" )
```
mleotta: Not a big deal, but you could do this in one call, right?  

    "set PATH=%PATH%;${fletch_DIR}/bin;${fletch_DIR}/x64/${_vcVersion}/bin\n"

Possibly this is two lines to make it more human readable?

aaron-bray:
Yes, that would be better

CMakeLists.txt
```diff
@@ -209,24 +243,35 @@ install( FILES log4cplus.properties   DESTINATION ${CMAKE_INSTALL_PREFIX} )
...
+
 if (KWIVER_ENABLE_PYTHON)
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export PYTHONPATH=$this_dir/${python_site_packages}:$PYTHONPATH\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "\n# additional python mudules to load, separated by ':'\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export SPROKIT_PYTHON_MODULES=kwiver.processes\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "\n# set to suppress loading python modules/processes\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "# export SPROKIT_NO_PYTHON_MODULES\n\n" )
+  if(WIN32)
+    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PYTHONPATH=%~dp0/lib/python2.7/%config%/site-packages\n" )
+    file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "\n:: additional python mudules to load, separated by ':'\n" )
```
mleotta: is the separator on Windows a semi-colon `;` instead of a colon `:`?  Or maybe a colon is used regardless of platform in this case.

aaron-bray:
I don't think find_package finds fletch though... I thought fletch_ROOT was
filled with find_package... And that wasn't getting filled out

Let me test it out

CMakeLists.txt
```diff
@@ -275,10 +320,10 @@ if ( NOT fletch_ENABLED_Boost)
   list( APPEND KWIVER_LIBRARY_DIRS "${BOOST_ROOT_ARG}/lib")
 endif()
 
-if (fletch_ROOT)
+if (fletch_DIR)
```
mleotta: same again with the `fletch_FOUND`.

CMakeLists.txt
aaron-bray: ; on windows an : on li/unix

:: is a batch comment

mleotta:
`fletch_ROOT` is no longer exported from Fletch, but it used to be.  However, several other variables are exported from Fletch, so we definitely do want to do a `find_package`.

CMakeLists.txt
mleotta: We should make the comment correct then, even if the script itself is not broken.  Someone might want to add something manually.

CMakeLists.txt
aaron-bray: I think I like this as two lines, if the fletch dir is really long its nice to have it separated out
it looks like this in the batch file

set PATH=%PATH%;C:/Programming/DIVA/builds/fletch/install/bin
set PATH=%PATH%;C:/Programming/DIVA/builds/fletch/install/x64/vc15/bin

mleotta: that's fine.  makes sense.

mleotta:
I'll test this tomorrow on my Windows machine

CMakeLists.txt
aaron-bray: Yep it is found, i think i have the find_package in the right spot now as well

aaron-bray:
Sounds good, I run it from my build root, then cd into the bin\release and I am able to successfully run 

pipeline_runner.exe -p C:\Programming\diva\kwiver\sprokit\pipelines\number_flow.pipe

CMakeLists.txt
```diff
@@ -275,10 +321,10 @@ if ( NOT fletch_ENABLED_Boost)
   list( APPEND KWIVER_LIBRARY_DIRS "${BOOST_ROOT_ARG}/lib")
 endif()
 
-if (fletch_ROOT)
+if (fletch_FOUND)
   # since we have fletch, add include and lib dirs to export
-  list( APPEND KWIVER_INCLUDE_DIRS "${fletch_ROOT}/include")
-  list( APPEND KWIVER_LIBRARY_DIRS "${fletch_ROOT}/lib")
+  list( APPEND KWIVER_INCLUDE_DIRS "${fletch_DIR}/include")
+  list( APPEND KWIVER_LIBRARY_DIRS "${fletch_DIR}/lib")
```
linus-sherrill: I believe fletch_ROOT is the correct symbol here since it really points to the fletch install area. fletch_DIR may also point there, but it does not have to. fletch_DIR is used by the find_module script to point to the directory that contains fletchConfig.cmake

CMakeLists.txt
```diff
@@ -182,6 +194,13 @@ configure_file(
...
+if(WIN32)
+  configure_file(
+    ${KWIVER_CMAKE_DIR}/setup_KWIVER.bat.in
+    ${KWIVER_SETUP_BATCH_FILE}
+    @ONLY
+    )
+endif()
 
 # install set up script
 option( KWIVER_INSTALL_SET_UP_SCRIPT "Whether or not to install setup_KWIVER.sh" ON )
```
linus-sherrill: Since this option applies to both the batch file and shell script, maybe the '.sh' could be dropped.

CMakeLists.txt
```diff
@@ -38,6 +38,18 @@ set(KWIVER_BINARY_DIR           "${CMAKE_BINARY_DIR}")
 set(KWIVER_CONFIG_FILE          "${KWIVER_BINARY_DIR}/kwiver-config.cmake")
 # Create initial setup script
 set(KWIVER_SETUP_SCRIPT_FILE    "${KWIVER_BINARY_DIR}/setup_KWIVER.sh" )
+if(WIN32)
```
linus-sherrill: It looks like we are treating windows as a special case in that we are always generating the shell script, but only generating the batch file if on windows. I think these two cases should be more symmetric. It would be less code to always generate the batch file if the conditionals around the batch file creation are removed and it was always generated. After all, the shell script is generated when built on windows. this would allow us to verify the contents of the batch file when building on Linux and maybe we would remember when we change the shell script to also change the batch file. :-)

aaron-bray:
Looks like a good case of "Those who do not learn history are doomed to
repeat it"

:-D

I can take the if blocks out

We should be using these scripts on the ctest machine, then we should find
errors quickly

CMakeLists.txt
mleotta: This is no longer true.  There was an abrupt change in Fletch making fletch_DIR point to the install area, putting fletchConfig.cmake in the install area, and removing fletch_ROOT.

CMakeLists.txt
aaron-bray: A .sh can actually be used on windows via a bash shell like cygwin or something, but I do not think a .bat can be used at all on linux.

I can take out the conditionals, it won't hurt anything on linux

if we move the ctest to use these env scripts we can ensure they are working

CMakeLists.txt
aaron-bray: Ok

CMakeLists.txt
aaron-bray: on my machine, fletch_ROOT is always empty
I am not sure how it gets set.
I believe it was removed from fletch by Johan, as fletch always puts everything in the install directory now, and nothing goes in the build directory

CMakeLists.txt
aaron-bray: ok

CMakeLists.txt
mleotta: Maybe the best compromise is the always construct the .bat but only have a the `if(WIN32)` around the install step.  That way we can check that the files is constructed properly on a Linux machine, but it doesn't clutter up the install with a file that is useless on that platform.

aaron-bray: This sounds good to me

CMakeLists.txt
```diff
@@ -38,6 +38,18 @@ set(KWIVER_BINARY_DIR           "${CMAKE_BINARY_DIR}")
...
+  set(KWIVER_SETUP_BATCH_FILE    "${KWIVER_BINARY_DIR}/setup_KWIVER.bat" )
+  if(CMAKE_CXX_COMPILER_ID MATCHES MSVC)
+   if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.10)
+      set(_vcVersion vc15)
+    elseif(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19)
+      set(_vcVersion vc14)
+    else()
```
aaron-bray: @dstoup Is there any plan for fletch to know that, if built with msvc, it can copy those pesky opencv dlls libs in to the install/bin?

If not, it seems that all users of fletch will have to write this special kind of code on windows to make sure all the dlls can be found

aaron-bray: I have to make sure this extra location gets added to the path on windows, based on this code

`file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/x64/${_vcVersion}/bin\n" )`

CMakeLists.txt
```diff
@@ -183,13 +194,31 @@ configure_file(
...
+  if(WIN32)
+    install( PROGRAMS   ${KWIVER_SETUP_BATCH_FILE}
+      DESTINATION ${CMAKE_INSTALL_PREFIX} )
+  endif()
+endif()
+
+if ( fletch_FOUND )
+  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/bin\n" )
+  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PATH=%PATH%;${fletch_DIR}/x64/${_vcVersion}/bin\n" )
+
```
aaron-bray: @dstoup Is there plans to have fletch put the opencv dlls and libs in the appropriate place so projects on windows don't have to have special code to find the appropriate location to add to the path when running?



(Sorry if this is a repeat, not sure what happened to my last conversation)

mleotta: I completely agree with this.  It's not going to happen for this release, but should be a priority for the next release.  Either Fletch needs to fix the install path of OpenCV libraries or it needs to export a `fletch_LIBRARY_DIRS` with all the paths we need to check to find all the libraries.

PR #222, opened 2017-07-14
Revert sprokit to use boost pointers

mleotta:
As sad as it makes me, I think we need to temporarily switch back to boost shared pointer.  The intent is to make this release of KWIVER functional and then @hughed2 will revert the commit as part of his branch to switch to pybind11.  Thanks to @mattdawkins for providing the patch.

### DIFF
--- a/sprokit/src/sprokit/pipeline_util/lex_processor.cxx
+++ b/sprokit/src/sprokit/pipeline_util/lex_processor.cxx
@@ -46,6 +46,8 @@
 #include <sstream>
 #include <set>
 
+#include <boost/make_shared.hpp>
+
 namespace sprokit {
 
 namespace {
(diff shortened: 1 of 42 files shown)


sprokit/src/sprokit/pipeline_util/lex_processor.cxx
```diff
@@ -637,7 +639,7 @@ lex_processor::priv::
 current_loc() const
 {
   // Get current location from the include file stack top element
-  return kwiver::vital::source_location( m_include_stack.back()->m_filename,
+  return kwiver::vital::source_location( std::make_shared< std::string >( *(m_include_stack.back()->m_filename) ),
```
mattdawkins: This is the only line where the find-replace failed, please verify that it doesn't break anything.

linus-sherrill: This should have nothing to do with the python interface. The string is used locally to report error locations. in fact, the m_filename and related changes could be reverted to std::shared_ptr.

mattdawkins: Let's leave as is then to simplify the hotfix, since David H will likely revert this patch on his branch

PR #227, opened 2017-07-17
Revert fletch_DIR to fletch_ROOT

mleotta:
fletch_ROOT was removed from Fletch and then came back.

Do not merge until after this branch lands:
https://github.com/Kitware/fletch/pull/153


mleotta:
@dstoup Here is the change that KWIVER needs after your Fletch branch lands.  It needs to be tested still.

PR #229, opened 2017-07-17
Fix location of Sprokit Python files on Windows

mleotta:
This change makes Sprokit consistently use lib/${config}/python2.7
instead of lib/python2.7/${config}.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,7 +262,7 @@ if (KWIVER_ENABLE_PYTHON)
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "\n# set to suppress loading python modules/processes\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "# export SPROKIT_NO_PYTHON_MODULES\n\n" )
 
-  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PYTHONPATH=%~dp0/lib/python2.7/%config%/site-packages\n" )
+  file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "set PYTHONPATH=%~dp0/lib/%config%/python2.7/site-packages\n" )
   file( APPEND "${KWIVER_SETUP_BATCH_FILE}" "\n:: additional python mudules to load, separated by ':'\n" )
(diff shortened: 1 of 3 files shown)


mleotta:
@aaron-bray can you test this?

aaron-bray:
I pulled this branch from the mleotta fork
(Is there an easy way to do this? I am not sure if there is a quick 'checkout' button i am missing)

anyway
a fresh build resulted in 

97% tests passed, 20 tests failed out of 577

it looks like there are a lot of errors with 
***Failed  Error regular expression found in output. Regex=[
Error: ]

which is what I got if certain pyd files were not in the right place, so maybe we are missing some kwiver or   vital pyd files for these tests...

overall though, this looks much better!

mleotta:
> On Jul 17, 2017, at 8:21 PM, aaron-bray <notifications@github.com> wrote:
> 
> I pulled this branch from the mleotta fork
> (Is there an easy way to do this? I am not sure if there is a quick 'checkout' button i am missing)
> 
I'm not aware of a button, but there are a few ways from the terminal.  Since it’s a pull request, you can checkout pull requests by number from the main repo:

    git fetch origin pull/229/head:dev/fix-sprokit-python-windows-paths

If you do this frequently you can set up your repo to always pull all the PRs whenever you fetch

    git config --add remote.origin.fetch "+refs/pull/*/head:refs/remotes/origin/pr/*”

You can checkout my topic branches (PR’ed or not) like this:

    git fetch https://github.com/mleotta/kwiver.git remote-branch-name:local-branch-name
> anyway
> a fresh build resulted in
> 
> 97% tests passed, 20 tests failed out of 577
> 
> 

That confirms what I saw.  I’ll merge it.

> it looks like there are a lot of errors with
> ***Failed Error regular expression found in output. Regex=[
> Error: ]
> 
> 

That’s actually an artifact of the testing frame work.  It greps through the test output looking for “Error:” and marks the test as failed if it finds a match.  So you should see the above message with almost any failing test, regardless of the cause.

> which is what I got if certain pyd files were not in the right place, so maybe we are missing some kwiver or vital pyd files for these tests...
> 
> 

I suspect there are still some issues with libraries not in the right place because there are errors related to not being able to load a module, but this is a step in the right direction.

PR #233, opened 2017-07-19
Fix CSV parsing.

JEHoctor:
Multiple commas should not be treated as one comma in a CSV file. Thus, trimEmpty should be set to false when running the tokenizer on a line of a CSV file.

When detected_object_set_output_csv.cxx is used to create a CSV file of detections and no image file name is supplied, the second column will be correctly created as an empty column. That is, the file will contain a double comma in each line. When the parser in detected_object_set_input_csv.cxx ignores this column, it triggers the error test on line 224. In summary, the CSV writer and reader in arrows are not compatible with one another before this fix.

### DIFF
--- a/arrows/core/detected_object_set_input_csv.cxx
+++ b/arrows/core/detected_object_set_input_csv.cxx
@@ -210,7 +210,7 @@ get_input()
   }
 
   m_input_buffer.clear();
-  kwiver::vital::tokenize( line, m_input_buffer, m_delim, true );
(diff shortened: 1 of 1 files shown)


mleotta:
This look like a bug fix that would apply to the release branch.  Please rebase this commit on the lastest release branch.  Then make a new commit that adds a summary of the fix (something like your commit message, maybe a bit shorter) to the doc/release-notes/1.1.1.txt files in the **Fixes** section.

PR #239, opened 2017-07-25
Dev/remove sprokit add plugin

linus-sherrill:
Remove old CMake function to create sprokit plugins. This has been replaced with the kwiver add plugin support in CMake. They were both mostly equivalent which makes one redundant. The sprokit version had a bespoke way of controlling dll import/export decoration. The kwiver cmake support uses builtin CMake import/export controls.


mleotta:
This seems like a reasonable change.  I'm going to merge in the release branch to bring in the fixed Travis CI testing just to verify that everything works.

PR #246, opened 2017-08-03
Visual Studio does not support the "and" keyword use "&&" instead.

mleotta:
This was causing build errors on Visual Studio


dstoup:
I would swear I fixed this issue before, but perhaps it was in a different file or repo. Change looks right to me.

mleotta:
This change was introduced in a recent change by @mattdawkins that I merged.  I didn't catch it during code review and Travis CI didn't either.  Apparently `and` is a valid keyword for GCC, but not for Visual Studio.

PR #248, opened 2017-08-03
Travis CI - use pre-built Fletch

mleotta:
This branch changes the Travis CI workflow to download a pre-built Fletch tarball instead of building it from source.  The advantages are:

- It's much faster to download instead of build.  As a result we can use a complete build of Fletch and build all of KWIVER without hitting Travis CI timeouts. 

- quicker turn-around time for getting CI results

- Failures in the Fletch build will not trigger failures of the KWIVER CI tests.

Disadvantages are:

- The Fletch build is no longer automatically updated.

- All branches of KWIVER must build against the same version of Fletch.

- Generates additional network traffic to data.kitware.com when Travis CI cache is out of date


mleotta:
This seems to be working quite well.  There is one test failure in sprokit causing Travis CI to report failure, but this seems to be a legitimate test failure that should be addressed on a separate branch.

PR #251, opened 2017-08-07
Dev/travis badge

Erotemic:
After having a conversation with @aaron-bray I found that there was indeed a travis CI build for kwiver. I added the badge to the readme so this would be easier to find.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -334,6 +334,27 @@ Sprokit
 `<sprokit/tests>`_     Sprokit unit tests
 ====================== =========================================================
 
+Testing
+========
+Continuous integration testing is provided by CDash_.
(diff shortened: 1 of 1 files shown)


mleotta:
Good Idea, but we should probably take it a step further and add a section on testing like we did on our MAP-Tk README: https://github.com/kitware/maptk#testing

There are actually two Travis CI badges of interest here.  One for the `release` branch and one for the `master` branch.  Also we want a link to the CDash dashboard https://open.cdash.org/index.php?project=Kwiver.

Also, I would make this documentation change on the `release` branch and target the PR to `release`.

Erotemic:
I've made those changes.

mleotta:
Looks good.  Thanks!

PR #254, opened 2017-08-08
Refactor track_set 

mleotta:
This branch refactors the `track_set` class using a bridge design pattern.  It separates the interface from the implementation such that derived classes for each can be used interchangeably.  This removes the need for `simple_track_set`, `simple_feature_track_set`, and `simple_object_track_set` and replaces them with a single `simple_track_set_implementation` that works for all cases.  This change also sets the stage for adding more complex `track_set_implementation` classes that will also apply across `track_set`, `feature_track_set`, and `object_track_set`.

This is an extension of PR #155 and the release notes add there are still relevant.

### DIFF
--- a/vital/types/feature_track_set.cxx
+++ b/vital/types/feature_track_set.cxx
@@ -43,6 +43,32 @@
 namespace kwiver {
 namespace vital {
 
+typedef std::unique_ptr<track_set_implementation> tsi_uptr;
+
+
+/// Default Constructor
+feature_track_set
+::feature_track_set()
+  : track_set(tsi_uptr(new simple_track_set_implementation))
+{
+}
+
+
+/// Constructor specifying the implementation
+feature_track_set
+::feature_track_set(std::unique_ptr<track_set_implementation> impl)
+  : track_set(std::move(impl))
+{
+}
+
+
+/// Constructor from a vector of tracks
+feature_track_set
+::feature_track_set(std::vector< track_sptr > const& tracks)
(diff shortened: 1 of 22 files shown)


vital/types/feature_track_set.cxx
```diff
@@ -43,6 +43,32 @@
...
+::feature_track_set()
+  : track_set(tsi_uptr(new simple_track_set_implementation))
+{
+}
+
+
+/// Constructor specifying the implementation
+feature_track_set
+::feature_track_set(std::unique_ptr<track_set_implementation> impl)
+  : track_set(std::move(impl))
```
linus-sherrill: Should this use the typedef from above?

mleotta: It certainly could, but I'm not sure that it needs too.  This signature matches the one in the header.  I only added the typedef because the first constructor required an explicit cast from raw to unique_ptr and writing it out on one line was a bit unwieldy.  Note that the typedef is local to this source file and not in the header.

mleotta:
The API is only defined once in the base class (`track_set_interface`).  The API is *implemented* in each of the derived classes.  This is based on a standard design pattern called a [Bridge Pattern](https://en.wikipedia.org/wiki/Bridge_pattern).  A standard bridge pattern actually does have a doubly-defined API, the Abstraction and Implementor classes are independent and the Abstraction members call the Implementor members.  I modified it slightly here such that there is a common abstract base class defines the API once.  Then both the Abstraction (`track_set`) and the Implementor (`track_set_implementation`) inherit that API.

PR #257, opened 2017-08-11
Frame-indexed track_set implementation

mleotta:
This is PR adds a new frame-indexed track_set implementation that caches track states belonging to each frame for much faster querying of tracks on a given frame (once the cache is built).  For algorithms like `analyze_tracks`, with repeated such calls, the code runs about 10x faster.

Right now I've put this in `arrows/core` and it must be explicitly used when constructing a `track_set`.  The original thought was that this would only be used in cases that cared about these queries because it is expensive to build the cache (e.g. in the constructor) if you are not going to use it, and many use cases of a `track_set` are to simply wrap a vector of tracks for passing into a function that immediately extracts the vector back out.  However, I've since modified the code to compute the cache lazily, the first time it is needed.  So there is no added cost if you do not call any query functions.  Now I'm wondering if this implementation should move into Vital and be the default implementation.  Building the cache may be slightly more expensive than doing a single query, but is probably worthwhile if doing two or more queries.

This PR also contains extensive unit tests to demonstrate that the frame-indexed implementation matches the original implementation.


mleotta:
Are there any objections to me merging this branch?  I can relocate this code to Vital later if we decide that is worthwhile.

mleotta:
I added release notes, and I'm going to merge after verifying that the CI tests still pass

PR #258, opened 2017-08-15
Can now load .py files in PYTHONPATH as plugins

Erotemic:
I ran into an issue where I had defined a Sprokit plugin module and put it in my PYTHONPATH, but the module was just a .py file instead of a package with an __init__ file. From a language perspective python treats these exactly the same, so Sprokit should to. 

Now instead of trying to run `os.listdir` on a file path, Sprokit checks if its a file or a directory and handles each case appropriately. 

This new feature should make prototyping pipelines much easier because you now have the option of working form a single file. 

------

While I was making this change I noticed that the "paths" returned by `_findPluginFilePaths` were not actually paths, but weird mangled half-path-synatx half-module-syntax strings. This went unnoticed because its only called from one function `_findPluginModules`, which transforms it into all module-syntax strings. I changed `_findPluginFilePaths`  to actually return relative paths like it seemed it was originally intended to do. From a functionality perspective nothing is changed by this, but the next person who reads this code may be less confused. 

Lastly, I fixed the `print` statements to be both python2/python3 compatible, fixed a few pep8 issues, and changed quotes to single quotes to be consistent.

### DIFF
--- a/sprokit/src/bindings/python/modules/loaders.py
+++ b/sprokit/src/bindings/python/modules/loaders.py
@@ -134,11 +134,12 @@ class ModuleLoader(Loader):
                 base = namespace_path
                 if base not in already_seen:
                     for ext in py_exts:
-                        mod_fpath = base + '.py'
+                        mod_fpath = base + ext
                         if os.path.isfile(mod_fpath):
                             already_seen.add(base)
-                            mod_rel_path = namespace_rel_path + '.py'
+                            mod_rel_path = namespace_rel_path + ext
                             yield mod_rel_path
+                            # Dont test remaining pyo / pyc extensions.
                             break
 
     def _findPluginModules(self, namespace):
--- a/sprokit/src/bindings/python/modules/loaders.py
+++ b/sprokit/src/bindings/python/modules/loaders.py
@@ -30,6 +30,7 @@
 
 """Facility to load plugins."""
 
+from __future__ import print_function
 import sys
 import os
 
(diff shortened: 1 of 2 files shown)


sprokit/src/bindings/python/modules/loaders.py
```diff
@@ -84,15 +84,26 @@ def _isPackage(self, path):
...
     def _findPluginFilePaths(self, namespace):
+        """
+        Searches for modules in `namespace` that are reachable from the paths
+        defined in the PYTHONPATH environment variable.
+        """
         already_seen = set()
 
+        for ext in ['.py', '.pyc']:
+            if namespace.endswith(ext):
+                print(('[WARNING] do not specify .py extension for '
```
Erotemic: I'm certainly not against adding the `__future__` line.

sprokit/src/bindings/python/modules/loaders.py
```diff
@@ -84,15 +84,26 @@ def _isPackage(self, path):
         return False
 
     def _findPluginFilePaths(self, namespace):
+        """
+        Searches for modules in `namespace` that are reachable from the paths
+        defined in the PYTHONPATH environment variable.
+        """
         already_seen = set()
 
+        for ext in ['.py', '.pyc']:
```
Erotemic: good catch, forgot about pyo's.

sprokit/src/bindings/python/modules/loaders.py
```diff
@@ -84,15 +84,26 @@ def _isPackage(self, path):
         return False
 
     def _findPluginFilePaths(self, namespace):
+        """
+        Searches for modules in `namespace` that are reachable from the paths
+        defined in the PYTHONPATH environment variable.
+        """
```
Erotemic: I added a few docstrings. I hope Google style is ok.

Purg: Given that there wasn't anything there before, this is better. I'm not sure we have a standard, but I have been using restructured/sphinx style in other projects (e.g. SMQTK).

sprokit/src/bindings/python/modules/loaders.py
```diff
@@ -115,21 +124,22 @@ def _findPluginFilePaths(self, namespace):
...
-                        yield mod_rel_path
+                base = namespace_path
+                if base not in already_seen:
+                    for ext in py_exts:
+                        mod_fpath = base + '.py'
+                        if os.path.isfile(mod_fpath):
+                            already_seen.add(base)
+                            mod_rel_path = namespace_rel_path + '.py'
+                            yield mod_rel_path
+                            break
```
Purg: This is fine for now, but try to avoid the use of breaks.

Purg:
I can't seem to find any tests for these classes. While we're in here, can we add some unit tests?

Erotemic:
I was actually considering writing some doctests, but I'm not sure how the testing works for this module. I'm most familiar with testing python projects with C++ plugins, so I'm getting used to the reverse direction. 

How does testing python code work for this project?

PR #261, opened 2017-08-22
Rename active release notes

mleotta:
The active release notes (those which have not yet been officially released) are now renamed to `release.txt` and `master.txt` and will be renamed to a version number right before the release is made.  This makes it easier to find these files and make it more clear that these are notes for unreleased changes.

The other file will be renamed to `master.txt` on a separate branch.


dstoup:
+1

PR #264, opened 2017-08-23
Fixed python3 print statements

Erotemic:
A smaller change fixing python3 syntax compatibility. I'm working on bigger changes, but I'll push this in the meantime.


vital/bindings/python/vital/types/eigen.py
```diff
@@ -419,7 +419,7 @@ def _destroy(self):
         # Not smart-pointer controlled in C++. We might not own the data we're
         # viewing.
         if self.c_pointer and self._owns_data:
-            # print "Destroying"
+            # print("Destroying")
```
Purg: Commented out code, so should probably be removed.

PR #268, opened 2017-08-24
Moving Sprokit/Vital test to py.test + symlinks

Erotemic:
Currently working on breaking this PR into several smaller ones. 

The major features in this PR are: 
* Python files can now be optionally symlinked to the build/install directories on unix systems.
* Vital python tests are now registered with `ctest`
* Sprokit tests are now installed as a submodule of `sprokit`

Things that should be done, but are not in this PR in order to keep it from growing futher:
* kwiver/sprokit/conf should be moved to kwiver/sprokit/cmake/conf
* the sprokit python bindings should be reorganized so the source directory mirrors the directory structure of the installed module (including tests). 
* sprokit, kwiver, and vital bindings should be given a `setup.py` and `pip install -e ` should be used to install them to a custom `site-packages`. 
* Maybe `sprokit.test` and `sprokit.tests` should be merged?
* There are many `kwiver_<funcname>` and `sprokit_<similar-funcname>` cmake functions that should probably be consolidated. 

Added
-----

* New option `KWIVER_SYMLINK_PYTHON` is available on unix systems. 

   If enabled pure python files will be symlinked to the build/install
   directories instead of being copied. This means you can develop existing
   python modules and fix broken python and tests and then retest without having
   to re-execute CMake. 

   This is done with the new function `sprokit_symlink_file_w_uid`, `sprokit_symlink_file_w_uid`, and `kwiver_symlink_file`. 

* a python tool to discover tests in python files.

    Uses python ast module to find functions/methods prefixed with Test/test_
    These are returned to cmake which then can set up an individual ctest
    target for each individual test. This tool is now used by both vital and sprokit.

* `pytest` version `3.2.0` is now in the vital `requirements.txt` file.
    
    This is required to run vital python tests with ctest
    Also pytest-cov and pytest-xdist plugins are added so we can run 
    with coverage and with the `--forked` parameter (allows us to test
    loading-based environment variable tests in the same py.test run)

Changes in Vital Python Tests
-----------------------------

* Vital python tests are now registered with ctest

   Previously vital tests relied on nose to be run. Now each test is registered
   with ctest and run with py.test.  The source code in the vital tests has not
   changed, so nose (an py.test) can still be used to run tests.

   NOTE: some vital python tests are broken.
   

Changes in Sprokit Python Tests
-------------------------------

* Removed need for external parameterization.
  
   Previously there were some tests functions that required arguments.  Now all
   functions are either paramaterless or paramatarizations are taken care of via
   py.test. 

    There is one hacky thing done to accommodate existing tests (while still allowing all python sources to be symlinked). There are a few tests that rely on two files in `kwiver/sprokit/tests/data/pipeline`. Previously, to access those file the python sources were configured with the path to that directory. Now I configure a text file that contains the path, and gracefully skip the test if it doesn't exist (e.g. if the source directory moved). 

   When I did this I thought many of the pipeline files in that dir were being used, but it turns out only two are, so a better way to do this in might be to just copy the required pipeline files to the python module. 

* Changed sprokit tests `__main__` to invoke `py.test`.
  
   Previously the main function used a internal introspection in `sprokit.test.test` to run a specific test. Now running the `__main__` without any args will run all tests. Specifying a testname will run the file in a similar way to how it was previously run. This change has a side benefit of removing all `import *` statements from the testing code.

* Sprokit python tests now have a `.py` suffix in the `${CTEST_BINARY_DIRECTORY}/tests/bin`. 
   
   This makes it more clear that these are python scripts and allows `py.test`
   to test them. I'm unsure if it is necessary to copy the tests to both the
   sprokit module and the tests bin directory. For now I'm doing both for an 
   easier transition.

* Removed `TEST_PROPERTY` comment fixtures in sprokit python tests

   These fixtures can be dynamically handled with pure python and py.test
   without needing to invoke non-standard testing fixtures. Examples include
   setting test timeouts and test environment variables.

CMake Enhancements
------------------

* cached external call to get python site-packages

    In `_kwiver_python_site_package_dir` there was a comment about how the result
    of the site-packages should be cached, so I went ahead and did that.  Now the
    executable is only run if the `PYTHON_EXECUTABLE` has changed.

* sprokit python tests are now configured to the `sprokit.tests` module.

  This mirrors the structure of `vital.tests` and is a more standard for python
  packages. The sources of the tests are still separated in the source tree, 
  more work needs to be done to reorganize the python source tree to look like 
  the module it will be built to. Once this is done a `setup.py` file can be added 
  (or configured to the binary directory) so the module can be installed with
  standard `pip` mechanisms.

* sprokit tests can now be run through `py.test` in addition to `ctest`
  

CMake Consolidations
--------------------

* Move sprokit python helper functions to their own file

   Moved python testing functions defined in kwiver/sprokit/tests/bindings/python/CMakeLists.txt to a new file `kwiver/sprokit/cmake/conf/sprokit-macro-python-tests.cmake` that mirrors the kwiver structure.

* registered vital tests modules in a list and call `kwiver_add_python_module` in a loop.

    There were redundant call `vital/bindings/python/vital/tests/CMakeLists.txt` to 
   `kwiver_add_python_module` with the same signature. I put them in a loop. This also made it much 
   simpler to register the testable modules with ctest.  I did the same change with some of the sprokit 
   tests (where appropriate).

*  Made intermediate cmake variables

    In several function there were long concatenations of existing variables used
    multiple times. I consolidated these into single variables (although in some
    cases these were just aliased for print-debugging / code readability purposes).  E.g. 

 ```python
  +  set(PYFILE_SRC "${path}")
  +  set(PYFILE_DST "${kwiver_python_output_path}${python_noarchdir}${python_sitepath}/${modpath}/${module}.py")
  +  # installation path for this module
  +  set(PYPKG_INSTALL_PATH "${python_install_path}/${kwiver_python_subdir}${python_sitepath}/${modpath}")
  ```

* removed duplicate function cmake func `_sprokit_python_site_package_dir`

    The functions `_kwiver_python_site_package_dir` and
    `_sprokit_python_site_package_dir` were exact duplicates so I removed the
    latter and replaced all references with the former.

* Removed cmake funcs `kwiver_add_python_module_int` and `sprokit_add_python_module_int`
   
   I replaced them with `kwiver_add_python_module` and `sprokit_add_python_module` because all the original function did was call these.

* Commented unused cmake variables

    In `CMake/utils/kwiver-utils-python.cmake`, the vars `kwiver_configure_cmake_args` and `kwiver_configure_extra_dests` seemed unused.
    It looks like the sprokit version of this function has analgous vars, but it uses them in its configure, whereas kwiver does not.

Small Fixes
-----------

* CYGWIN was misspelled as CYTWIN. 
* Added various docstrings


linus-sherrill:
This PR has conflicts that need to be resolved before it can be accepted.

linus-sherrill:
Is this still relevant or OBE?

Erotemic:
I think there might be some unmerged things in this, but it looks like most of the important stuff was merged in other branches.

PR #287, opened 2017-09-08
Remove VITAL_FOREACH

mleotta:
Replace VITAL_FOREACH with a standard range-based for loop


mattdawkins:
I have so many VITAL_FOREACH's in existing branches I'm going to need to replace

mleotta:
It's a fairly simple search and replace, and you should do it anyway, even if this branch doesn't land.  New code does not need VITAL_FOREACH.

I did something like this in Bash:

    for name in $(grep -Rl VITAL_FOREACH *); do
      perl -pi -e "s|VITAL_FOREACH\(([^,]*),(.*)|for\(\$1 :$2|g" $name
      sed -i '' '/vital_foreach/d' $name
    done

I ran a view variations of the above command to deal with slight different spacing issues.

PR #289, opened 2017-09-08
Make track_set mutable

mleotta:
This PR makes `track_set` a mutable class with member functions to `insert` and `remove` tracks.  It also updates feature tracking and loop closure algorithms to modify the `track_set` in place using these and other functions rather than recreate a `track_set` from a vector of tracks each time.  This change allows the `track_set` implementations to persist and in some cases (e.g. `frame_indexed_track_set_impl`) keep an updated cache for faster retrieval.

This PR also moves the merging capability from an external function to a member of `track_set` and uses a new `track_data_redirect` subclass of `track_data` to let a track who's states have been transferred via merge to another track keep a pointer to that track.  This bookkeeping helps resolve race conditions in track linking in which tracks are involved in multiple merges.

This branch should make typical work with track sets more efficient, however, it does not address how to handle efficiently track_sets in streaming pipelines where you need a snapshot of the tracks at a given time step.  For now I think we will need to explicitly copy or recreate the track_set when pass through a pipeline.


mleotta:
@mattdawkins Can you provided your input on this branch?  I just resolved conflict with your branch that merged with master.

PR #292, opened 2017-09-08
Dev/doc improvements

linus-sherrill:
Added missing release note on config get as vector. 
Added deprecated note on process draw_detected_object_boxes_process. All that drawing code has been moved to a drawing algo with the wrapper process. The new process `draw_detected_object_set_process` allows the actual drawing algo to be specified in the config.


mleotta:
looks okay to me 👍

PR #294, opened 2017-09-08
Explicitly added __init__ files for core vital/sprokit modules

Erotemic:
This is a PR towards the goal of making the python bindings first class citizens. The goal it to make the core python modules is the binary directory look almost exactly like the ones in the source directory. Once this is accomplished,  a `setup.py` script can be written allowing the vital/sprokit/kwiver modules to be installed via pip. Furthermore, this opens the possibility of creating and distributing wheels that can be pip installed via pypi. 

The core of this change is to simply explicitly add the `__init__.py` files to the source directory, and then replace instances of `kwiver_create_python_init` and `sprokit_create_python_init` with `kwiver_add_python_module` and `sprokit_add_python_module`. This will make it less confusing for primarily python developers (like myself) to understand the directory layout. 

In fact, most `__init__.py` files were already explicitly in the vital package. Most changes made by this PR are actually in sprokit.

---------

Currently, the parts of each python module are scattered all over the place (especially in sprokit). So they still need to be moved around, but having the `__init__` files in the directory will help make it clear what the directories that need to be moved are. 

This PR also does not change anything using the `*_create_python_plugin_init` macro. Ideally this would also be removed, but I'm don't have a clear idea of how to handle the arrows yet.


Erotemic:
I'm closing this PR, for now. Some of its functionality may be incorporated later when we restructure the python modules themselves.

PR #296, opened 2017-09-11
Added cmake option to symlink python files

Erotemic:
This is an attempt to move some of the functionality out of  #268 and into standalone components. 

This PR adds the option `KWIVER_SYMLINK_PYTHON` (defaulting to Off), which makes it so Python files can be optionally symlinked to the build/install directories on unix systems. 

By default, this PR is cause no change to existing functionality. However, by turning symlinks on, instead of having to run cmake every time you change a python file, you can simply edit it and the changes will immediately be reflected in the build files.  While this make python development easier and more dynamic, it is not always a desirable behavior, which is why it is optional and defaulted to false (also it wont work on systems without symlink support). 

#### The first commit
The first commit  adds the CMAKE helper functions `kwiver_symlink_file`, `sprokit_symlink_file`, `sprokit_symlink_file_with_uid` (eventually I think all the sprokit versions should simply be replaced with the kwiver versions, but somebody correct me if I'm wrong about that). 

It modifies the file `kwiver_add_python_module_int` and `sprokit_add_python_module_int` to use the symlink function instead of the corresponding configure option if `KWIVER_SYMLINK_PYTHON` is enabled.

Currently, there are very few configuration variables actually used in python files so the actual configure step is (almost) equivalent to a copy. Most all instances of configure variables in Python files refer to the `PYTHON_EXECUTABLE` cmake variable to setup a shebang (e.g. `#!@PYTHON_EXECUTABLE@` ) However, this is not necessary and should be replaced with `#!/usr/bin/env python`. 

The only other configuration variable that is used is a comment in the `test-pymodules.py` file 

```
# TEST_PROPERTY(ENVIRONMENT, PYTHONPATH=@CMAKE_CURRENT_SOURCE_DIR@)
def test_pythonpath():
```

This test will fail if symlinks are on in this branch. However, these comment-based macros will be replaced by `pytest` configurations once the move to `pytest` is complete (hopefully breaking PRs like this out of #268, will make that easier), and then this test will pass even with symlinks on (note because symlinks are off by default and should only be enabled when developing python code, this will not cause dashboard failures). 

#### The second commit
The second commit replaces the configured shebangs with the previously mentioned static ones. It touches a lot of files, but is a very small and repetitive change. The main change only touches 4 files.

### DIFF
--- a/CMake/utils/kwiver-utils-configuration.cmake
+++ b/CMake/utils/kwiver-utils-configuration.cmake
@@ -84,3 +84,54 @@ function(kwiver_configure_file name source dest)
       )
   endif()
 endfunction()
+
+
+###
+#
+# Mimics a `kwiver_configure_file`, but will symlink `source` to `dest`
+# directly without any configureation. This should only be used for interpreted
+# languages like python to prevent the need to re-make the project after making
+# small changes to these interpreted files.
+#
+# SeeAlso:
+#     kwiver/CMake/utils/kwiver-utils-python.cmake
+#     kwiver/sprokit/conf/sprokit-macro-configure.cmake
+#
+function (kwiver_symlink_file name source dest)
+
+  if(EXISTS ${dest} AND NOT IS_SYMLINK ${dest})
+    # If our target it not a symlink, then remove it so we can replace it
+    file(REMOVE ${dest})
+  endif()
+
+  # Need to ensure the directory exists before we create a symlink there
+  get_filename_component(dest_dir ${dest} DIRECTORY)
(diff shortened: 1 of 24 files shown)


linus-sherrill:
You are correct about the duplicate cmake support in kwiver and sprokit. The goal is to convert all code to use the kwiver_* versions. There are still some redundant files/functions in sprokit which we should migrate to the kwiver_ equivalents, as time permits.

Erotemic:
I rebased onto master and fixed merge conflicts with recent branches.

PR #297, opened 2017-09-11
Improved site-packages implementation

Erotemic:
Similar to #296, this PR is a small standalone piece of #268.

The work is based on a comment I saw in `_sprokit_python_site_package_dir`

```cmake
#		
# This is run many times and should produce the same result, which could be cached.		
# Think about it.		
```

So, I basically did that. I created an internal cache variable that stores the result of the call. If `PYTHON_EXECUTABLE` is ever changed, this cache is reset.


Erotemic:
Closing because  #315 makes this obsolete.

PR #299, opened 2017-09-11
Removing *add_python_module_int

Erotemic:
This PR is a small standalone piece of #268.
In both the sprokit and kwiver python utils (which should likely be consolidated and merged), there were two functions `{kwiver,sprokit}_add_python_module_int` that were only used internally to each file. The functions  `{kwiver,sprokit}_add_python_module` were called elsewhere, but all they did was call the `_int` functions. 

I essentially removed these do-nothing functions and renamed the `_int` functions to `{kwiver,sprokit}_add_python_module` to reduce code clutter. 

Also,  I added a TODO comment to remove these lines 
```
  file(APPEND "${init_template}"    "from pkgutil import extend_path\n")
  file(APPEND "${init_template}"    "__path__ = extend_path(__path__, __name__)\n")
```

from python plugin init files, as I don't believe they are necessary (unless someone can correct me).

### DIFF
--- a/CMake/utils/kwiver-utils-python.cmake
+++ b/CMake/utils/kwiver-utils-python.cmake
@@ -93,10 +93,28 @@ function (kwiver_add_python_library    name    modpath)
 
 endfunction ()
 
+
 ###
-# - internal implementation -
 #
-function (kwiver_add_python_module_int    path     modpath    module)
+# kwiver_add_python_module(path modpath module)
+#
+# Installs a pure-Python module into the 'modpath' and puts it into the
+# correct place in the build tree so that it may be used with any built
+# libraries in any build configuration.
+#
+# Args:
+#     path: Path to the python source (e.g. kwiver_process.py)
+#     modpath: Python module path (e.g. kwiver/processes)
+#     module: Python module name. This is the name used to import the code.
+#         (e.g. kwiver_process)
+#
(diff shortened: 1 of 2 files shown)


linus-sherrill:
Would it be possible to remove the sprokit version of the _ _add_python_module() function and use the kwiver version?

Erotemic:
I'm interested in doing that. It would make things much cleaner. However, there seems to be functionality in the sprokit version that does not exist in the kwiver version. The functionality relates to configure_uids, and so far I haven't been able to get a clear sense of what they are doing / why they are needed. 

The sprokit version has the vars `sprokit_configure_extra_dests`, `sprokit_configure_cmake_args`, which seem to be used (kwiver has them too, but they are unused). 

I could be that perhaps they aren't needed at all and we can just git rid of them,  I suspect the uids are only needed (if they are needed at all) for C++ build files and they really don't apply to python. If this is the case, then I'd be happy to consolidate it to just the kwiver version.

Purg:
``KWIVER_ENABLE_PYTHON`` is off on the travis tests, so that's not being verified as working.  We can either approve this without CI testing, or another branch should be made to enable python in travis CI.

mleotta:
We should definitely turn Python on in Travis CI testing now.   Let me try that in another branch and see what happens.

PR #303, opened 2017-09-11
Move sprokit test helpers to their own file 

Erotemic:
This is part of #268, but it is not quite standalone because it depends on PR #297.

There were several sprokit test helper functions defined in `kwiver/sprokit/tests/bindings/python/CMakeLists.txt`, that I moved into a helper file to conform the style used by the rest of the codebase. 

The actual implementation of these functions is unchanged in this PR, but in a future PR based on #268 they will change to use some of the functionality introduced in #302.

### DIFF
--- a/sprokit/conf/sprokit-macro-python-tests.cmake
+++ b/sprokit/conf/sprokit-macro-python-tests.cmake
@@ -1,0 +1,102 @@
+# Should these functions be rectified with the kwiver versions
+# in kwiver/CMake/utils/kwiver-utils-python-tests.cmake ?
+
+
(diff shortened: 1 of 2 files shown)


Erotemic:
This PR is a first step towards that. I'm trying to move as incrementally as possible, so there are no huge PRs to review.

Erotemic:
The needed PRs have been merged to make this PR standalone. 

In addition to splitting the code out, I did make a small addition to support `KWIVER_SYMLINK_PYTHON` as well as adding additional documentation to the functions.

PR #315, opened 2017-09-14
Consolidated kwiver/sprokit exact duplicate python helpers

Erotemic:
This PR consolidates kwiver/sprokit python functionality that was *exactly* duplicated. 

The variables `kwiver_python_output_path`, `kwiver_python_subdir`, `sprokit_python_output_path`, `sprokit_python_subdir`  and were being set in multiple places, and they were always being set to the same value. Additionally a function to find the python site-packages directory was being called multiple times with the same result. 

The main change is that I made a new file `CMake/kwiver-setup-python.cmake` (open to different naming suggestions), that handles creating all of these static python configuration variables at the beginning. I also added support for switching the version to python3 (replacing the `SPROKIT_ENABLE_PYTHON3` option). 

Because `_kwiver_python_site_package_dir` is no longer needed, this PR removes the need for  #297.


CMake/kwiver-setup-python.cmake
```diff
@@ -0,0 +1,118 @@
...
+  if (_prev_kwiver_python_version)
+    message(STATUS "The Python version changed; refinding the interpreter")
+    unset(PYTHON_EXECUTABLE CACHE)
+    unset(PYTHON_INCLUDE_DIR CACHE)
+    unset(PYTHON_LIBRARY CACHE)
+    unset(PYTHON_LIBRARY_DEBUG CACHE)
+  endif()
+endif()
+
+if (KWIVER_PYTHON_VERSION MATCHES "^3.*")
```
Purg: If we're using specific version of 2 and 3 here, the above KWIVER_PYTHON_VERSION should accurately reflect these values, e.g. specifically use 2.7 and 3.4 strings, checking for those specific strings here.

CMake/kwiver-setup-python.cmake
```diff
@@ -0,0 +1,118 @@
+###
+# Finds the python binaries, libraries, inclue, and site-packages paths
```
Purg: Please add some documentation about the variables that are exported or intended for use outside of this script, here. E.g. kwiver_python_subdir

CMake/kwiver-setup-python.cmake
```diff
@@ -0,0 +1,118 @@
...
+  set( python_sitename "site-packages")
+endif()
+string(STRIP "${python_site_packages}" python_site_packages)
+message(STATUS "Python site-packages to install into: ${python_site_packages}")
+message(STATUS "python_site_packages = ${python_site_packages}")
+message(STATUS "python_sitename = ${python_sitename}")
+
+
+execute_process(
+  COMMAND "${PYTHON_EXECUTABLE}" -c "import sys; print(sys.version[0:3])"
```
Purg: We're looking for the version of python found while finding / requiring a specific version above?  Should this be checked against the requested version of python?

CMake/kwiver-setup-python.cmake
Erotemic: Nothing is imported, but I'll add a message about the exports.

CMake/kwiver-setup-python.cmake
Erotemic: KWIVER_PYTHON_VERSION  is only allowed to either be the major version; either 2 or 3 (its a combo box). 

I'm open to reworking this versioning code. Its a little weird, but I did it that way to (1) reduce the number changes to the variables, (2) to agree with how its done elsewhere (e.g. fletch, opencv), and (3) to let us default to python 2 and while giving the user the choice of switching to 3.

I think its good to have to have a toggle variable that lets you switch between the two major versions (a-priori without knowing which is available), while having a separate variable to manage the minor ones (perhaps we should mark that variable as advanced?).  On the other hand, we might forgo that and have one variable with the full version. Maybe there's a different way I'm too tired to think of?

CMake/kwiver-setup-python.cmake
Erotemic: We request the major version above, this grabs the major + minor version. We could add a check to make sure they agree, pending we don't rework how the versioning is done.

Erotemic:
@Purg I added more documentation and generally cleaned things up. Overall, I think it should be much clearer as to what's going on.

Note, I found a few more duplicated sections of code (relating to boost-python and find_packages), and I removed/consolidate those as well.

PR #321, opened 2017-09-25
Building track_oracle induces vxl dependency

collinsr:
Trigger VxL dependency when building track_oracle.

Previously, building track_oracle all by itself failed since nothing triggered the VxL dependency (you had to also build arrows.) This commit allows track_oracle to build standalone (without, for example, arrows.)


track_oracle/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ else()
 endif()
 
 find_package( TinyXML REQUIRED )
+find_package( vxl REQUIRED )
```
dstoup: Is this line needed? It looks like it should be doing the find inside kwiver-depends-VXL.cmake, which you've already included.

collinsr: Indeed, both that line and the find_package( TinyXML ) above it seem to be redundant... patch revision forthcoming...

dstoup:
LGTM

PR #324, opened 2017-09-26
Add include to define free() on MacOS and others.

linus-sherrill:
Add include needed by MacOS to define free() function


linus-sherrill:
Problem fixed by another branch

PR #331, opened 2017-10-03
Fixed typo in PYTHON_VERSION varname

Erotemic:
The variable `_python_version` was not defined. I fixed the script to use `PYTHON_VERSION`. 

Looking at this again I was trying to figure out how it ever worked. Then I realized that the `fletch_MAJOR_PYTHON_VERSION` does not exist yet in fletch master. So it was working because both variables were undefined. What I should have been checking is the `PYTHON_VERSION` variable against the `KWIVER_MAJOR_PYTHON_VERSION` variable.  This patch fixes the code to do that.

### DIFF
--- a/CMake/kwiver-setup-python.cmake
+++ b/CMake/kwiver-setup-python.cmake
@@ -153,8 +153,8 @@ import sys
 print(sys.version[0:3])
 ")
 # assert that the right python version was found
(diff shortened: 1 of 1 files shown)


Erotemic:
Now that Fletch#176 is merged into master, the `fletch_PYTHON_MAJOR_VERSION` variable will be populated, and cause errors. This should be merged soon to prevent those errors.

dstoup:
LGTM ... any objections?

PR #335, opened 2017-10-03
Examples

aaron-bray:
Reorganizes the various sprokit pipeline files into the examples directory.
Pipefiles are now *.in files and will configure to reference image/videos in the source pool.
Also added the darknet pipe and associated data (~250MB) to girder and created 2 cmake projects to manage it (1 to pull the data, 1 to unzip and configure the pipefiles)
This darknet projects are excluded from the build and must be run manually.
(That will be explained in a documentation section/page dedicated to object detection (using darknet)

This compliments the documentation PR and its accompanying example C++ files/projects


linus-sherrill:
Would you have time to convert the config entries in the pipe files to the new format?
Old format:
`:name value`
New format:
`name = value`

aaron-bray:
Yeah sure, will do.

Note, I have not been able to have sprokit find the python processes for some reason.
(when modules_python gets loaded)
There are 4, print_numbers, ApplyDescriptor, ProcessImage, homography_writer

They are in a sprokit\processes\python and seem to be in a 'kwiver' namespace (if that applies to python), I think they should be here : sprokit\src\bindings\python\processes and in a 'sprokit' namespace

I did that and got only the print_number process to be found but when executed my pipeline runner just locked up. 

I was hoping to get this in, and someone with better python skills could look into these processes and see why they are not found and do not seem to be running when they are.

linus-sherrill:
Jenkins test this please

PR #336, opened 2017-10-04
Move to pybind11 v2.2

hughed2:
The first commit is a bunch of the original pybind11 conversion code plus the start of an upgrade to v2.2, the rest are smaller changes. It could probably use a bit more cleanup, but it works and passes the tests we would expect it to pass.

I want to give Jon Crall a chance to try this out before it gets merged, but putting this up now for inspection and comments.


hughed2:
This is ready to go in.

Note that the travis failure was because of a darknet issue, not from the pybind11 stuff.

PR #338, opened 2017-10-16
ReadMe images update and Darknet hash update

aaron-bray:
The zip contents were not in sync with the pipeline examples directory structure, so I uploaded a new zip to girder and this is the new hash

Image updates:
Enable darknet in cmake image options
I also do not have the test option enabled in the image as end users would need to have google test installed which they may not. 

If we want to have end users run unit test, we probably need a section describing that in full.


aaron-bray:
Is this travis ci server on an old fletch? I am not sure why it would not be able to find the darknet headers

PR #339, opened 2017-10-16
Added loggers to sprokit python modules

Erotemic:
This PR makes two changes.

1. Pure python modules in sprokit now use logging instead of print statements. 
2. Python exceptions are now printed when called from C++. 

First: 

The print statements used in python files have been converted to use the builtin python logging module. 
This logger will respect the `KWIVER_DEFAULT_LOG_LEVEL` environment variable unless the `KWIVER_PYTHON_DEFAULT_LOG_LEVEL` is set, in which case it will be overriden. 

Second: 

Previously, if Python raised an exception, when called from C++, it would fail silently. The `SPROKIT_PYTHON_IGNORE_EXCEPTION` and `SPROKIT_PYTHON_HANDLE_EXCEPTION` macros catch the exception and then call a function called `sprokit::python::python_print_exception`, but for some reason (that I don't understand), that call is not actually printing the exception. Therefore I've used the vital logger to log the error message as a warning. While this doesn't fix the underlying issue, it does indicate that something has happened.

### DIFF
--- a/sprokit/src/bindings/python/sprokit_logging.py
+++ b/sprokit/src/bindings/python/sprokit_logging.py
@@ -4,7 +4,6 @@ A thin wrapper around python logging.
 """
 from __future__ import print_function, unicode_literals, absolute_import
 import sys
-import six
 import os
 import logging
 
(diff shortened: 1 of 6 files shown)


sprokit/src/bindings/python/sprokit_logging.py
```diff
@@ -0,0 +1,113 @@
...
+    ]
+    text = '\n'.join(lines)
+    print(text)
+
+
+def exc_report(func):
+    """
+    Prints a message if an exception occurs in the decorated function.
+
+    Modules called from C++ (e.g. pybind11) are not gaurenteed to print
```
linus-sherrill: Spelling - guaranteed

sprokit/src/sprokit/python/util/python_exceptions.h
```diff
@@ -38,42 +38,48 @@ namespace python {
...
+#define SPROKIT_PYTHON_HANDLE_EXCEPTION(call)                     \
+  try                                                             \
+  {                                                               \
+    call;                                                         \
+  }                                                               \
+  catch (pybind11::error_already_set const& e)                    \
+  {                                                               \
+    auto logger = kwiver::vital::get_logger("python_exceptions"); \
+    LOG_WARN(logger, "Handle Python Exception:");                 \
+    LOG_WARN(logger, e.what());                                   \
```
linus-sherrill: This could be all on one line:

  `LOG_WARN(logger, "Handle Python Exception: " << e.what() );`

sprokit/src/sprokit/python/util/python_exceptions.h
```diff
@@ -38,42 +38,48 @@ namespace python {
...
+#define SPROKIT_PYTHON_IGNORE_EXCEPTION(call)                     \
+  try                                                             \
+  {                                                               \
+    call;                                                         \
+  }                                                               \
+  catch (pybind11::error_already_set const& e)                    \
+  {                                                               \
+    auto logger = kwiver::vital::get_logger("python_exceptions"); \
+    LOG_WARN(logger, "Ignore Python Exception:");                 \
+    LOG_WARN(logger, e.what());                                   \
```
linus-sherrill: This could all be on one line:

`LOG_WARN(logger, "Ignore Python Exception: " << e.what()); '

sprokit/src/bindings/python/sprokit_logging.py
```diff
@@ -0,0 +1,113 @@
...
+    @exc_report
+    def __init__(self, name):
+        self._logger = logging.getLogger(name)
+        self.log('debug', 'created logger for ' + name)
+
+    @exc_report
+    def log(self, level, msg):
+        if isinstance(level, six.string_types):
+            level = getattr(logging, level.upper())
+        self._logger.log(level, msg)
```
linus-sherrill: Could this be routed through the vital logger?

sprokit/src/sprokit/python/util/python_exceptions.h
Erotemic: will fix.

sprokit/src/sprokit/python/util/python_exceptions.h
Erotemic: will fix.

sprokit/src/bindings/python/sprokit_logging.py
Erotemic: This continues to be one of my most misspelled words. Will fix.

sprokit/src/bindings/python/sprokit_logging.py
Erotemic: Hmm, it might be possible to write the appropriate C++ bindings. However, I think that may add unnecessary complexity. Also, Python's logging system is very mature and well suited to logging in a python environment (e.g. it can produce line numbers in the python source files or be configured to produce colored output). We may be able to implement a special Handler class that routes to the vital logger, but my initial feeling are that this is unnecessary.

linus-sherrill: There is a use case we have encountered where our software is embedded in a larger system and the customer wants all log messages to go through that logger so they can manage and capture the log stream. This was done by adding a project specific logger back end, but it does require that all log messages be run through the main c++ logger. We do not need this support at this time, but it is worth making sure the architecture can support it.

linus-sherrill:
jenkins test this please

PR #348, opened 2017-10-30
[WIP] Crop Chips

Erotemic:
Another component for the simple tracker. The idea is to add an algorithm that can take an image and a set of bounding boxes and extract the corresponding chips. A new process crop_detections will use this algorithm to get a chip for each detection. This chip will later be passed to a feature extraction algorithm. 

This is built on top of #345


linus-sherrill:
Superseded by 384

PR #349, opened 2017-10-30
Initial version of logging guide.

linus-sherrill:
Add initial version of guidance for logging and how to use the different levels.


aaron-bray:
This looks good, 
Pretty standard definitions.
I am good for pushing this.

I think we need to have a 'logging' section in the tutorials.rst
Something that explains if you build with log4cplus, then you can go to the log4cplus.properties file an muck with options. I usually go in and change the process loading classes at the bottom to TRACE so I can see which libraries kwiver cannot load.

PR #350, opened 2017-10-31
Visual studio fixes

hughed2:
It looks like we'll need a few fixes for Visual Studio. I've started off by getting rid of some code that worked in VS14 but is broken in VS15 (and has become unnecessary anyway). Some other changes will have to be made as well. At the very minimum, C bindings are broken in VS15 (haven't checked VS14, but I assume they worked at some point) and some of the new GTest code is not windows friendly.


dstoup:
It looks like these changes will help fix the broken dashboards. Is it ready to consider, or still WIP?

hughed2:
We can add this in. There are a number of things broken in windows, the current commit will fix one but not all.

PR #361, opened 2017-11-03
[WIP] add example tracking support

linus-sherrill:
Adding example - skeletal tracker support by moving basic classes from other topic branches.

### DIFF
--- a/arrows/core/CMakeLists.txt
+++ b/arrows/core/CMakeLists.txt
@@ -8,24 +8,25 @@ set( plugin_core_headers
   close_loops_multi_method.h
   compute_ref_homography_core.h
   convert_image_bypass.h
-  detected_object_set_input_kw18.h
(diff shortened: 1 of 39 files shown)


mattdawkins:
I would prefer you don't submit this. You cherry-picked a majority of the other tracking branch. It could possibly mess up the git history of other branches that tracking branch has been merged into.

collinsr:
We'll never just land that branch in one go; we can't land a megabranch (with all of the inertia associated with its code mass) and defer refactoring until a hypothetical point in the future when enough energy has been gathered to overcome that inertia. IMHO it's better to take it step by step.

mattdawkins:
This branch has 68% of the mass of the other one and as far as I can tell duplicated functionality. Where do you draw the line on what is considered a "megabranch"?  :-1:

PR #369, opened 2017-11-08
Workaround for compiler bug in older versions of visual studio

hughed2:
I haven't checked this in non-VS 2015 yet, so make sure to doublecheck dashboards. Otherwise, it fixes the pybind11 error on old versions of Visual Studio.


dstoup:
Jenkins test this please

dstoup:
I think the kwiverPR dashboard and probably the fletch nightly are misconfigured a bit. This branch builds when the correct things are enabled but the fact that it doesn't when certain configs are jiggled is a different bug. Merging.

PR #375, opened 2017-11-09
Remove explicit setting of logger from all processes.

linus-sherrill:
The default logger name is already set by the base class and also doing
it in the derived class leads to non-standard logger names.


linus-sherrill:
Jenkins test this please

PR #376, opened 2017-11-13
process property handling

linus-sherrill:
Added process properties that are bound to the process object. Process properties were created based on the process class hierarchy, where the derived process would create the set of properties. There are cases where the derived process does not know all the properties, such as when process instrumentation is enabled. The solution is to add a set of properties that is part of the process object that can accumulate these condition based properties. Properties from the derived class are merged with the set of properties from the base class when the set of properties is requested

Add attribute to plugin factory to allow process properties to be forwarded to the created process. This allows the process registration operation to specify specific attributes for a process and to communicate those properties to the process.

### DIFF
--- a/doc/release-notes/master.txt
+++ b/doc/release-notes/master.txt
@@ -101,6 +101,16 @@ Vital
  * Added at() method to detected_object_set to provide bounds checked
    direct indexing.
 
+ * Added process properties that are bound to the process
+   object. Process properties were created based on the process class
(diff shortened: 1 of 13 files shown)


hughed2:
Just a quick FYI: I've got a PR coming in soon that checks /all/ C++ schedulers and throws an error if it has a python process. You can add a check for specific schedulers if you want, especially if you want it handled in a special way, but it's not strictly necessary.

PR #377, opened 2017-11-14
Converted Sleep() to kwiversys Delay().

linus-sherrill:
Sleep() is not universally available on all supported systems.
Converted to OS Neutral version from kwiversys.

### DIFF
--- a/examples/cpp/CMakeLists.txt
+++ b/examples/cpp/CMakeLists.txt
@@ -5,7 +5,6 @@ set(SOURCE "main.cpp"
 )
 
 add_executable(kwiver_examples ${SOURCE})
-target_link_libraries(kwiver_examples 
-                      vital_algo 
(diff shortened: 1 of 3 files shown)


aaron-bray:
Dang, you are right

We could also do something like this : std::this_thread::sleep_for(std::chrono::seconds(2));
I think that is cross system compatible, and you wouldn't need the #if's

linus-sherrill:
I used the Delay() function from SystemTools since it has support for all supported systems. You could try ...::sleep_for(). I don't have enough test systems to make sure it works on all platforms.

aaron-bray:
I think it works...
But I am fine with Delay, if we get some time we could replace the Delay
with that std sleep method

I would rather go through the same utility so if we every need to swap out
functionality, we only need to do it in one place.

I approve of this change

PR #386, opened 2017-11-15
Robustly set argv in sprokit C++ python entrypoint

Erotemic:
By default pybind11 does not set `sys.argv`, which caused my to add a workaround to `sprokit/__init__.py`. While debugging other python issues I found a better way to accomplish this by calling `PySys_SetArgvEx` immediately after the call to `Py_Initialize` in the sprokit code that registers python modules. 

While making this change I also removed `using namespace  pybind11` in favor of the `namespace py = pybind11;` pattern that is present in the pybind11 tutorials. This helps clarify what code is coming from the pybind11 headers.

### DIFF
--- a/sprokit/src/bindings/python/__init__.py
+++ b/sprokit/src/bindings/python/__init__.py
@@ -4,19 +4,6 @@ The base SPROKIT package initialization
 """
 # flake8: noqa
 from __future__ import print_function, unicode_literals, absolute_import
-import sys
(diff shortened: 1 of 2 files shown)


linus-sherrill:
This branch has developed a conflict with pybind11 name space handling.

linus-sherrill:
This branch has developed conflicts

PR #388, opened 2017-11-15
Added PYTHON_LIBRARY to setup_KWIVER.sh

Erotemic:
This is a fix for Issue #385

Previously python symbols were not available to plugins unless the main executable linked to libpython. The `python_modules` library in sprokit  had a mechanism for preloading libpython using the `PYTHON_LIBRARY` environment variable, but this would fail unless that was set. 

I simply added `PYTHON_LIBRARY` to the setup_KWIVER.sh script to point to it. 

Note, this will not fix issues on deployment systems, because the `PYTHON_LIBRARY` variable is an absolute path found by CMake. This seems like an ok temporary solution, but a better one would make the `python_modules`  not rely on the environment variable to preload the python symbols. 

Honestly, I don't even know why it needs to be cause `ldd lib/sprokit/modules_python.so` shows that it is linked to libpython2.7. Is there a way to use `dlfcn.h` to dynamically get the path to whatever libpython it is being linked to and call `dlopen( pylib, RTLD_LAZY | RTLD_GLOBAL );` on whatever that path is? Then if `PYTHON_LIBRARY` we use it, otherwise we fallback to whatever python lib we are linked to. (or maybe we just always do the latter of the two choices?)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,17 +244,22 @@ endif()
 
 
 if (KWIVER_ENABLE_PYTHON)
+  file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "# Python environment\n")
+  file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export PYTHON_LIBRARY=\"${PYTHON_LIBRARY}\"\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export PYTHONPATH=$this_dir/${python_site_packages}:$PYTHONPATH\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "\n# additional python mudules to load, separated by ':'\n" )
   file( APPEND "${KWIVER_SETUP_SCRIPT_FILE}" "export SPROKIT_PYTHON_MODULES=kwiver.processes\n" )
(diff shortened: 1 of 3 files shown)


Erotemic:
In my latest commit, I added a fair bit of code, which is motivated by these observations/ideas: 

* The `python_modules.so` file is linked to libpython
* So, we should be able to load symbols from libpython without forcing the user to specify an environment variable. 
* That being said, if the user does specify the envvar, we should still respect it. 

It turns out doing this is a bit of a pain. I was looking for a way to essentially run `ldd` at runtime to get the name of the libpython we are linking to, but I was unable to figure out how to do it. 

-------

So, I went with the next best option: 
* Use CMake to define a C preprocessor macro (-DPYTHON_LIBRARY=${PYTHON_LIBRARY}) that encodes the name of libpython in the binary. 

I needed to add a macro to transform the macro value into a literal string, but while doing this I hit an issue. The PYTHON_LIBRARY had the pattern `-linux-` in it, and `linux` was already defined by the preprocessor, so when my PYTHON_LIBRARY macro was expanded, the linux subpattern was also expanded, yielding an invalid path. 

Thus, I added in a dirty hack which saves the value of `linux`, undefines it, and then redefines it as the end of the file. 


-------

I left the setup_KWIVER.sh stuff in there as well, because I think its useful. Note: if encoding the name of libpython as a preprocessor macro is determined to be unwanted, this PR still works perfectly well without commit c0bf93783.

PR #389, opened 2017-11-17
Get rid of extra dll import/exports

hughed2:
The SPROKIT_PIPELINE_EXPORT is unnecessary in the python binding cxx files, but prevent successful Windows builds. I've removed them from the factory code that was moved over to the binding files.


linus-sherrill:
Jenkins test this please

PR #395, opened 2017-11-21
Add process timing instrumentation

linus-sherrill:
Added a new process instrumentation variant that writes the method time to a CSV file. This instrumentation type is designed to assist in doing process timing analysis. Process timing can be generated on a per-process basis by enabling process instrumentation for a process.  This is done in the pipe file as follows:

    process my_process :: image_reader
      block _instrumentation
        type = timer  # select timing instrumentation
        timer:type = cpu  # valid types are cpu and wall
      endblock

An optional config entry for "output_file" can be used to specify a non-default file name. The default name is <process-name>_timing.csv. In the above example, the file name would be "my_process_timing.csv"

### DIFF
--- a/examples/pipelines/descriptor.pipe.in
+++ b/examples/pipelines/descriptor.pipe.in
@@ -25,12 +25,6 @@ process descriptor
   algorithm = algo
 
 # ================================================================
-# global pipeline config
-#
-config _pipeline:_edge
-       capacity = 10
-
-# ================================================================
 # connections
 connect from input.image
         to   descriptor.image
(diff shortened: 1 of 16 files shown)


aaron-bray:
From a syntax and construction point of view this looks good.
And it looks to solve the problem at hand, and should be fine to merge in.

I have not really dug into the inner workings of sprokit, and how it actually traverses it's pipeline.
I would like to spin up on the schedulers and other mechanisms that make up sprokit.
Is there any kind of primer or road map that kind of introduce the concepts?
(I also have not done a deep dive into the sprokit docs yet either, but I think I might manage without my water wings)

PR #408, opened 2017-11-28
Converting boost::any to vital::any

hughed2:
For the most part this was a straightforward change. I had to do a minor update to any_cast to make sure the object has actually been instantiated, with a corresponding update to the bad_any_cast error message.


hughed2:
I didn't do anything involving demangle, and this worked fine on my linux machine, but it's close enough to my code that I think it's worth taking another quick look at those dashboard errors.

linus-sherrill:
The vital_util library may need to be added in some cases to supply the demangle() function.

linus-sherrill:
Jenkins test this please

PR #413, opened 2017-11-29
The yaml_parser in kfp examples needs the include_dir for yaml set

dstoup:


dstoup:
Looking at Roddy's branch makes me believe this branch should not merge, but rather should be included in that PR. Commenting over there.

PR #414, opened 2017-11-30
Fix copy of config entry to subblock.

linus-sherrill:
The previous fix for copying config entries did not account for dropping
one config level. Although it did copy all the config entry metadata.


collinsr:
@aaron-bray just cc'ing you in-- with this change, I can run the darknet pipline as supplied by Aaron without modification.

PR #420, opened 2017-12-04
track-oracle plumbing for KPF geometry

collinsr:
track_oracle plumbing to support KPF geometry I/O.

As a side effect, track_oracle/examples/track_reader_example can now convert back and forth between kw18 and kpf-geometry.

(With the exception of the timestamp_usecs, which hasn't yet been formally defined as KPF timestamp domain.)

All the changes are inside track_oracle except for the following commits:
5f945a1 ("Read support for KPF geometry files")
eb1bf7d ("Implement KPF EVAL packet")
daf1714 ("Minor cleanups of the KPF packet declaration")

### DIFF
--- a/CMake/kwiver-depends-kpf.cmake
+++ b/CMake/kwiver-depends-kpf.cmake
@@ -4,5 +4,8 @@ option( KWIVER_ENABLE_KPF
   "Enable kpf i/o for vital types"
   ${fletch_ENABLED_YAMLCPP}
 )
-# Mark this as advanced 
+# Mark this as advanced
 mark_as_advanced( KWIVER_ENABLE_KPF )
+if (KWIVER_ENABLE_KPF )
+  add_definitions( -DKWIVER_ENABLE_KPF )
+endif()
--- a/track_oracle/core/track_oracle_core_impl.cxx
+++ b/track_oracle/core/track_oracle_core_impl.cxx
@@ -143,6 +143,19 @@ track_oracle_core_impl
   }
 }
 
+vector< field_handle_type >
+track_oracle_core_impl
+::get_all_field_handles() const
+{
+ boost::unique_lock< boost::mutex > lock( this->api_lock );
+ vector< field_handle_type > ret;
+ for (const auto p: this->name_pool)
+ {
+   ret.push_back( p.second );
+ }
+ return ret;
+}
+
 field_handle_type
 track_oracle_core_impl
 ::unlocked_lookup_by_name( const string& name ) const
--- a/track_oracle/core/track_oracle_core_impl.cxx
+++ b/track_oracle/core/track_oracle_core_impl.cxx
@@ -248,6 +248,31 @@ track_oracle_core_impl
   return ret;
 }
 
+vector< vector< field_handle_type > >
+track_oracle_core_impl
+::fields_at_rows( const vector<oracle_entry_handle_type>& rows ) const
+{
+  boost::unique_lock< boost::mutex > lock( this->api_lock );
+  vector< vector< field_handle_type > > ret( rows.size() );
+
+  vector< oracle_entry_handle_type > sorted_rows( rows );
+  std::sort( sorted_rows.begin(), sorted_rows.end() );
+
+  for (map< field_handle_type, element_store_base* >::const_iterator i = this->element_pool.begin();
+       i != this->element_pool.end();
+       ++i)
+  {
+    auto exists_list = i->second->exists( rows );
+    for (auto j=0; j<exists_list.size(); ++j)
+    {
+      if (exists_list[j])
+      {
+        ret[j].push_back( i->first );
+      }
+    }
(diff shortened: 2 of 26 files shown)


collinsr:
After regex patch (b8afc17), Travis builds / installs; usual 4 tests fail on linux.

CMake/kwiver-depends-kpf.cmake
```diff
@@ -4,5 +4,8 @@ option( KWIVER_ENABLE_KPF
   "Enable kpf i/o for vital types"
   ${fletch_ENABLED_YAMLCPP}
 )
-# Mark this as advanced 
+# Mark this as advanced
 mark_as_advanced( KWIVER_ENABLE_KPF )
+if (KWIVER_ENABLE_KPF )
```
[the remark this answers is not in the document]
collinsr: Theoretically it could be possible to use KPF without YAML (if we add protobuf as an alternative transport). This particular patch enables the KPF-specific includes and whatnot; the find_package(YAML...) calls are in the per-directory CMakeLists.txt files with the vague idea that it made more sense to find YAML closest to where it was actually needed.

That said, if it's better / more customary to do that find_package once and localize all the searching and so forth here in the depends step, I'm happy to change it.

track_oracle/file_formats/track_kpf_geom/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  PUBLIC               track_oracle
+                       data_terms
+                       kwiver_algo_kpf
+  PRIVATE              vital_logger
+                       vul
+                       vgl
+                       track_oracle_format_base
+                       logging_map
+                       kwiversys
+                       ${YAML_CPP_LIBRARIES}
```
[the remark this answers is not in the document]
collinsr: fixed in next commit

track_oracle/core/track_oracle_core_impl.cxx
```diff
@@ -143,6 +143,19 @@ ::track_oracle_core_impl()
   }
 }
 
+vector< field_handle_type >
+track_oracle_core_impl
+::get_all_field_handles() const
+{
+ boost::unique_lock< boost::mutex > lock( this->api_lock );
```
linus-sherrill: Can this be done with std library calls?

collinsr: I strongly suspect so; however, the rest of the track_oracle singleton uses this pattern. I'm happy to submit a followup branch that de-boosts.

linus-sherrill:
Jenkins test this please

track_oracle/core/track_oracle_core_impl.cxx
dstoup: That would be good. I believe track_oracle is one of the packages that explicitly requires Boost in the CMake though, so at least it's protected. DeBoostificaction should be a separate PR  for sure.

track_oracle/example/track_reader_example.cxx
```diff
@@ -197,18 +204,38 @@ int load_tracks( const string& track_fn, track_handle_list_type& tracks, const s
...
 
     frame_handle_list_type frames = track_oracle_core::get_frames( tracks[i] );
     frame_count += frames.size();
+    vector< oracle_entry_handle_type > frame_handles( frames.size() );
+    for (auto j=0; j<frames.size(); ++j)
+    {
+      frame_handles[j] = frames[j].row;
+    }
+    /*
     for (size_t j=0; j<frames.size(); ++j)
```
dstoup: Can we ditch dead code?

track_oracle/example/track_reader_example.cxx
```diff
@@ -197,18 +204,38 @@ int load_tracks( const string& track_fn, track_handle_list_type& tracks, const s
...
+    for (auto fields_at_frame: track_oracle_core::fields_at_rows( frame_handles ))
+    {
+      for (auto field: fields_at_frame)
+      {
+        ++frame_stats[ field ];
+      }
+    }
+    if (timer.real() > 5 * 1000)
+    {
+      LOG_INFO( main_logger, "Computed stats on " << i << " of " << tracks.size() << "; " << frame_count << " frames" );
```
dstoup: Wondering if log message should be info level of something smaller, like debug/trace. Same message applies throughout, there are a lot of new INFO messages.

track_oracle/example/track_reader_example.cxx
collinsr: Yes; I promise to do this in the followup dev/kpf-track-oracle-activities branch. :)

track_oracle/example/track_reader_example.cxx
collinsr: Actually, all the "LOG_INFO( ... )" calls in the examples/ programs should probably just be regular calls to cout. Willfix in the activities branch.

PR #422, opened 2017-12-07
Fix missing trailing parentheses

Purg:
Typo unnoticed because this macro is not used anywhere yet it seems.


dstoup:
Good catch, thanks. We should merge this as soon as travis finishes.

PR #426, opened 2017-12-13
Update to readme

ankitshah009:
number_flow.pipe name changed to number_flow.pipe.in


ankitshah009:
Current readme has correct version of commands one needs to run. 
However to enable examples, one need to add -DKWIVER_ENABLE_EXAMPLES:BOOL=ON

PR #439, opened 2018-01-05
Export not needed on templates

aaron-bray:


arrows/kpf/yaml/kpf_canonical_types.h
```diff
@@ -56,7 +56,7 @@ namespace canonical
 //
 
 template< typename T >
-struct KPF_YAML_EXPORT scoped
```
dstoup: We don't call any of the struct's methods from outside the library?

aaron-bray:
Its a template, there is nothing to link to, its all dynamically type
instantiated at compile time

PR #445, opened 2018-01-09
Add kwiver applet runner excutable.

linus-sherrill:
This new executable is used to run the tool plugins.
Ported config_explorer to new plugin pattern.


tools/kwiver_applet.cxx
```diff
@@ -0,0 +1,71 @@
...
+  m_context = ctxt;
+}
+
+
+// ----------------------------------------------------------------------------
+std::ostream&
+kwiver_applet::
+outstream() const
+{
+  return *m_context->m_ostream;
```
mleotta: This would benefit from a check to make sure the pointer is not null

mleotta: Actually it needs two checks for null.  One for `m_context` and one for `m_ostream`.

tools/kwiver_applet.cxx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+// ----------------------------------------------------------------------------
+const std::string&
+kwiver_applet::
+applet_name() const
+{
+  return m_context->m_applet_name;
```
mleotta: make sure the `m_context` is not null.

tools/kwiver_tool_runner.cxx
```diff
@@ -0,0 +1,123 @@
...
+
+  // initialize the global context
+  G_context.m_ostream = &std::cout;
+  G_context.m_wtb.set_indent_string( "      " );
+
+  // Test to see if there are any args
+  if ( (argc < 2) || (std::string(argv[1]) == "help") )
+  {
+    // display help message
+    std::cout << "Usage: " << argv[0] << " <applet?  [args]" << std::endl
```
mleotta: `<applet?` --> `<applet>`

tools/kwiver_tool_runner.cxx
```diff
@@ -0,0 +1,123 @@
...
+  {
+    // Create applet based on the name provided
+    applet_factory app_fact;
+    auto applet = app_fact.create( argv[1] );
+    applet->initialize( &G_context );
+
+    // Edit the command line args
+    // TBD Need to add support for -o <file> flag to direct output to a file.
+
+    // TBD Add support for converting <applet-name> "help" into <applet-name> "--help"
```
mleotta: Another way to hand this is to support the syntax `kwiver help <applet-name>`.  If you leave off the applet name you get the list of applets.  This is what `git` does for example.  Though it also supports the equivalent of `kwiver <applet-name> --help`.  I would make the tool runner handle the first case and leave it up to each tool to implement a `--help` option that does the same thing if they want (but it would be recommend)

tools/kwiver_tool_runner.cxx
```diff
@@ -0,0 +1,123 @@
...
+  // ----------------------------------------------------------------------------
+  try
+  {
+    // Create applet based on the name provided
+    applet_factory app_fact;
+    auto applet = app_fact.create( argv[1] );
+    applet->initialize( &G_context );
+
+    // Edit the command line args
+    // TBD Need to add support for -o <file> flag to direct output to a file.
```
mleotta: If we are going to have options like this at the tool_runner level, maybe they should need to come before the applet name, like `kwiver -o logfile.txt <applet> [applet options]`.

I could change this, but currently I use `-o` to specify output locations to dump config files in the tools in MAP-Tk.

We will probably also want to have default config files installed for each tool and loaded automatically with the option of specifying an additional config file on the command line to override parameters.  I'm not sure if this should be for all tools, or left up to each tool to implement

tools/kwiver_tool_runner.cxx
```diff
@@ -0,0 +1,123 @@
...
+
+    // TBD Add support for converting <applet-name> "help" into <applet-name> "--help"
+    // This requires that all applets support --help
+
+    // create local args for tool
+    int local_argc(argc-1);
+    char** local_argv;
+    local_argv = static_cast< char** >(calloc( sizeof(char*), argc-1 ));
+
+    local_argv[0] = argv[0];
```
mleotta: Should `local_argv[0]` be `argv[0]`, `argv[1]`, or `argv[0] + " " + argv[1]`.  I could see an argument for any of these.

tools/kwiver_tool_runner.cxx
```diff
@@ -0,0 +1,123 @@
...
+    // Run the specified tool
+    return applet->run( local_argc, local_argv );
+  }
+  catch ( kwiver::vital::plugin_factory_not_found& )
+  {
+    std::cerr << "Tool \"" << argv[1] << "\" not found. Type \""
+              << argv[0] << " help\" to list available tools." << std::endl;
+
+    exit(-1);
+  }
```
mleotta: It would be useful to catch all other uncaught exceptions here as well and print a meaningful message rather than a program crash.

tools/config_explorer.h
```diff
@@ -0,0 +1,59 @@
...
+{
+public:
+  config_explorer();
+
+  virtual int run( int argc, char* argv[] );
+
+protected:
+
+private:
+  void print_help();
```
mleotta: It seems like `print_help` should be an abstract virtual function on the base class. That way each applet must implement it and this would be useful to implement `kwiver help <applet-name>`

tools/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  PRIVATE        vital
+                 vital_vpm
+                 vital_logger
+                 kwiversys
+                 kwiver_tools
+)
+
+###
+# Add applet for config
+kwiver_add_plugin( tool_config_explorer
```
mleotta: Maybe this is just and example, but it seems like it would be better to not have one plugin for each tool.  We could have one plugin for all vital tools, maybe another for all sprokit tools.

mleotta:
This looks like a really great start.  I'm quite excited about this.

tools/kwiver_applet.cxx
linus-sherrill: Done

tools/kwiver_applet.cxx
linus-sherrill: Done

tools/kwiver_tool_runner.cxx
linus-sherrill: Oops - Done

tools/kwiver_tool_runner.cxx
linus-sherrill: Done

tools/config_explorer.h
linus-sherrill: Good idea. I added a method to base class that writes usage information to a specified stream.

tools/CMakeLists.txt
linus-sherrill: Can easily add multiple applets to a plugin. I started with tis approach for porting existing tools which may have bad namespace habits.

tools/kwiver_tool_runner.cxx
```diff
@@ -0,0 +1,255 @@
...
+
+
+// ============================================================================
+class command_line_parser
+{
+public:
+  command_line_parser( int p_argc, char *** p_argv )
+    : m_argc(0)
+    , m_argv(0)
+    , m_state(0)
```
mleotta: Is there a reason that this needs to be a member variable rather than a local variable?

tools/kwiver_tool_runner.cxx
```diff
@@ -0,0 +1,255 @@
...
+  try
+  {
+    // Create applet based on the name provided
+    applet_factory app_fact;
+    auto applet = app_fact.create( options.m_applet_name );
+    tool_context->m_applet_name = options.m_applet_name;
+    applet->initialize( tool_context.get() );
+
+    // check for <applet> help
+    if ( options.m_applet_args[1] == "help" )
```
mleotta: I don't think we want to hijack the applet arguments by looking into its argument list for `help`.  The syntax should be  `kwiver help <applet>` not `kwiver <applet> help`.  This is how Git does it too.  Basically `help` is like its own simple applet.  With no arguments is prints all the other applets, when passed an applet name it prints the help for that applet.   We can also support `kwiver <applet> --help`, but that will be up to each applet to handle if they want.

tools/config_explorer.cxx
```diff
@@ -0,0 +1,238 @@
...
+
+  if ( ! arg.Parse() )
+  {
+    std::cerr << "Problem parsing arguments" << std::endl;
+    exit( 0 );
+  }
+
+  if ( opt_help )
+  {
+    usage( std::cout );
```
mleotta: should this be `usage( this->outstream() );`?

tools/applet_context.h
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @brief Applet context provided by the tool runner.
+ *
+ */
+class applet_context
+{
+public:
+
+  // current output stream set by tool runner
+  std::shared_ptr<std::ostream> m_ostream;
```
mleotta: Should we be using the vital logger here instead of an ostream?  Or maybe there is a way to configure the logger in code such that log messages are also set this ostream?  If I called this option on my tool to change the output to a file, I would expect all logger messages to go there as well.

tools/config_explorer.cxx
linus-sherrill: I believe there is a standard for usage information, although I have not been able to find it, that specifies which output stream to use (either stdout or stderr). git uses stdout. The intent of the outstream() stream is to allow the output of an applet to be captured in a file.

dstoup:
Jenkins test this please

dstoup:
Since there was previous excitement about this work, is it something we should usher in?

linus-sherrill:
I think this is the time to complete porting the tools to this plugin approach since we have forked the release branch.

mleotta:
@linus-sherrill please also add release notes in the `master.txt` release notes file to describe these changes.

linus-sherrill:
Ready for review

vital/kwiversys/CommandLineArguments.cxx
```diff
@@ -134,6 +134,19 @@ void CommandLineArguments::Initialize(int argc, char* argv[])
   this->Initialize(argc, static_cast<const char* const*>(argv));
 }
 
+// ----------------------------------------------------------------------------
+void CommandLineArguments::Initialize(const std::vector<std::string>& argv)
```
mleotta: Is this upstream in KWSys?  If not, should it be?

sprokit/src/tools/pipeline_runner_pl.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include "kwiver_applet.h"
+
+#include <string>
+#include <vector>
+
+namespace kwiver {
+namespace tools {
+
+class pipeline_runner
```
mleotta: Are we missing the code to register this applet in a sprokit plugin?

sprokit/src/processes/clusters/registration.cxx
```diff
@@ -42,7 +42,8 @@
 #include <vital/util/tokenize.h>
 
 #include <sprokit/pipeline_util/load_pipe_exception.h>
-#include <sprokit/pipeline_util/path.h>
+#include <sprokit/pipeline_util/pipeline_builder.h>
+//+#include <sprokit/pipeline_util/path.h>
```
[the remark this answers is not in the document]
linus-sherrill: Done

vital/kwiversys/CommandLineArguments.cxx
linus-sherrill: This has not yet been pushed upstream.

sprokit/src/tools/pipeline_runner_pl.h
linus-sherrill: The pipeline runner is not set up to run as a plugin yet. Still a work in progress. It looks like all sprokit tools should be converted at the same time because they share a lot of common functionality. These pipeline runner files could be moved to another branch.

mleotta:
If the pipeline runner stuff is not ready, I think it would be best to move it to a new branch so we can land the framework first.

linus-sherrill:
Jenkins test this please

mleotta:
@linus-sherrill this still seems to contain a lot of sprokit changes.  Is that expected?  I'm okay with merging this, but I don't want to leave the sprokit tools in a broken state.

linus-sherrill:
I can't say that the sprokit changes are benign, so let me redo this branch to get them out.

linus-sherrill:
Cleaned up branch to include only applet support.

mleotta:
I think we lost the release note, but I'd like to not hold this up anymore.  I'm going to merge.  Please make sure there is a release note in your follow-up branch.

PR #446, opened 2018-01-11
[WIP] initial exploratory work on modifying algorithm interface

linus-sherrill:
Attempting to define new algorithm creation and configuration semantics,

working towards an approach that is something like the following:

    auto algo = create_algorithm( config_block );
    def_config = algo->get_configuration();
    def_config->merge_config( config_block );
    algo->set_configuration( def_config );


linus-sherrill:
Still too nebulous to review

PR #449, opened 2018-01-18
bgr rgb color space conversion 

linus-sherrill:
This is an extension of #368 and addresses comments about converting color space in place.

### DIFF
--- a/arrows/ocv/image_container.h
+++ b/arrows/ocv/image_container.h
@@ -54,9 +54,10 @@ class KWIVER_ALGO_OCV_EXPORT image_container
 {
 public:
 
+  enum ColorMode{RGB,BGR};
+
   /// Constructor - from a cv::Mat
-  explicit image_container(const cv::Mat& d)
-  : data_(d) {}
+  explicit image_container(const cv::Mat& d, ColorMode cm = RGB);
 
   /// Constructor - convert kwiver image to cv::Mat
   explicit image_container(const vital::image& vital_image)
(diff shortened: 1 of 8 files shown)


mattdawkins:
I still think the default should be BGR

arrows/ocv/image_container.h
```diff
@@ -54,9 +54,10 @@ class KWIVER_ALGO_OCV_EXPORT image_container
 {
 public:
 
+  enum ColorMode{RGB,BGR};
+
   /// Constructor - from a cv::Mat
-  explicit image_container(const cv::Mat& d)
-  : data_(d) {}
+  explicit image_container(const cv::Mat& d, ColorMode cm = RGB);
```
mleotta: I agree with @mattdawkins that since there is a default color mode should probably be BGR here, because we assume by default that cv::Mat images are in BGR.

arrows/ocv/image_container.h
```diff
@@ -98,7 +99,7 @@ class KWIVER_ALGO_OCV_EXPORT image_container
   static vital::image_pixel_traits ocv_to_vital(int type);
 
   /// Convert a VITAL image to an OpenCV cv::Mat
-  static cv::Mat vital_to_ocv(const vital::image& img);
+  static cv::Mat vital_to_ocv(const vital::image& img, ColorMode cm = RGB);
```
mleotta: Likewise here the default should probably be BGR when producing a cv::Mat

arrows/ocv/image_container.h
```diff
@@ -117,7 +118,7 @@ class KWIVER_ALGO_OCV_EXPORT image_container
  *
  * \param img Image container to convert to cv::mat
  */
-KWIVER_ALGO_OCV_EXPORT cv::Mat image_container_to_ocv_matrix(const vital::image_container& img);
+KWIVER_ALGO_OCV_EXPORT cv::Mat image_container_to_ocv_matrix(const vital::image_container& img, image_container::ColorMode cm = image_container::RGB);
```
mleotta: Here as well the default should probably be BGR too

mattdawkins:
Please don't rebase this branch, I've merged it into a few others

PR #454, opened 2018-01-25
Move python libraries install location

hughed2:
Python binding libraries build at ${library_subdir}${library_subdir_suffix}. However, they install to ${library_subdir}. Previously, all libraries would install to lib/python#.#, rather then the module subdirectories. Putting the full path in ${library_subdir} insures they install to the correct location.

Also, pointing the vital bindings to the kwiver rather than the sprokit functions.


hughed2:
I see a bunch of new failing tests on Windows, but they're all ocv:image_conversion. That shouldn't be affected by this commit, which is should only be affecting how python libraries are built, plus one python file not used in those tests. Have we been seeing these tests fail elsewhere?

PR #459, opened 2018-02-04
Fix GIL lock issue for embedded pipelines, other cleanup

mattdawkins:


hughed2:
Jenkins test this please

hughed2:
LGTM

sprokit/src/sprokit/python/util/python_gil.h
```diff
@@ -36,37 +36,126 @@
...
+ */
+#define SPROKIT_SCOPED_GIL_ACQUIRE_START                                          \
+  {                                                                               \
+    pybind11::gil_scoped_acquire acquire;                                         \
+                                                                                  \
+    (void) acquire;
+
+#define SPROKIT_SCOPED_GIL_ACQUIRE_END                                            \
+  }
+
```
linus-sherrill: Having a start and an end scope macro hides the brackets from the user and the editor so it is not clear what is really going on. Somebody could assume that the scope is not really a bracketed scope but some other mechanism and put the SPROKIT_SCOPED_GIL_ACQUIRE_END call at a different block level.

It would be more obvious if the code from these macros were just in the calling code. A macro could be used to hide the pybind11 call if desired e.g. SPROKIT_SCOPED_GIL_ACQUIRE();  or something to make it look like a call.

sprokit/src/sprokit/python/util/python_gil.h
```diff
@@ -36,37 +36,126 @@
...
+    {                                                                             \
+      pybind11::gil_scoped_acquire acquire;                                       \
+                                                                                  \
+      (void) release;                                                             \
+      (void) acquire;
+
+#define SPROKIT_SCOPED_GIL_RELEASE_AND_ACQUIRE_END                                \
+    }                                                                             \
+  }
+
```
linus-sherrill: This one is a little more complicated than the previous one, but it still obfuscates the real block structure. My main concern is that the ..._END macro gets placed incorrectly with respect to the surrounding block structure. A block matching editor will not help in this case. 

One possible alternate implementation could be to include the code that is in the scope as part of the macro as follows:


    #define SPROKIT_SCOPED_GIL_RELEASE_AND_ACQUIRE(CODE)                              \
      {                                                                               \
        pybind11::gil_scoped_release release;                                         \
        {                                                                             \
          pybind11::gil_scoped_acquire acquire;                                       \
          (void) release;                                                             \
          (void) acquire;
          CODE;                                                                 \
        }                                                                             \
      }

mattdawkins:
This branch is likely superseded by https://github.com/Kitware/kwiver/pull/468

PR #464, opened 2018-02-09
Shorten pythread_per_process_scheduler name

mleotta:
This is to avoid the Windows 248 character limit on filenames.

### DIFF
--- a/sprokit/src/bindings/python/schedulers/CMakeLists.txt
+++ b/sprokit/src/bindings/python/schedulers/CMakeLists.txt
@@ -5,4 +5,4 @@ sprokit_create_python_plugin_init(sprokit/schedulers)
 
 sprokit_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/pythread_per_process_scheduler.py
   sprokit/schedulers
-  pythread_per_process_scheduler)
+  pythread_per_process)


mleotta:
This change fixed my build, but I haven't tried using the pythread_per_process_scheduler, so I don't know if there are any other side effects or run-time issues.

sprokit/src/bindings/python/schedulers/CMakeLists.txt
```diff
@@ -5,4 +5,4 @@ sprokit_create_python_plugin_init(sprokit/schedulers)
 
 sprokit_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/pythread_per_process_scheduler.py
   sprokit/schedulers
-  pythread_per_process_scheduler)
```
linus-sherrill: You could shorten the name even more since it is largely irrelevant.
"python scheduler", "ppp_sched"

mleotta: removing "_scheduler" was enough to get the file path under the 248 limit  (this name string is used twice in the path).  The "scheduler" part seemed redundant because it is also in a "schedulers" subdirectory, which shows up in the path as well.

Anyway, it works from a compilation point of view, but I was unclear if this name was used elsewhere and if any other code needed to change to make it work at run time.

aaron-bray:
Looks good to me, any little bit helps.
I have not run any kwiver python in a long time.
I am not sure if this will impact stuff, maybe David Hughes might have an idea?

PR #466, opened 2018-02-09
Cerr requires include iostream.

dstoup:


linus-sherrill:
A very similar fix was merged as MR#462.  Pull master and see if there is still a problem.

dstoup:
Fixed in master

PR #468, opened 2018-02-09
Fix issues with GIL deadlocking

KyleFromKitware:
All of our GIL deadlock issues are being caused by what I suspect is a bug in pybind11. I have filed a bug report here: https://github.com/pybind/pybind11/issues/1273


sprokit/src/bindings/python/modules/registration.cxx
```diff
@@ -105,13 +115,24 @@ register_factories(kwiver::vital::plugin_loader& vpm)
...
+      // Let pybind11 initialize threads and set up its internal data structures
+      pybind11::gil_scoped_acquire acquire;
+      (void)acquire;
+    }
+  }
 
-  (void)gil;
+  _load_python_library_symbols();
 
+  SPROKIT_PYTHON_GIL_SCOPED_ACQUIRE_BEGIN
```
mattdawkins: per Linus' comment on https://github.com/Kitware/kwiver/pull/459 (which I abandoned) we might want to swap these macros out so they don't have an explicit start/end, I think it's fine either way. It looks like everywhere they're used you might be able to just use the gil_scoped_acquire in non-macro form to simplify the code.

KyleFromKitware: I decided to go with the macros because I don't want to duplicate the `(void)acquire;` warning suppression everywhere, but I am also ok with doing it either way.

linus-sherrill: You can still wrap the warning suppression in a macro. I just think that it is easy for a future maintainer to get the ..._END macro in the wrong scope, either by inserting more complicated logic into the scoped block or not understanding the semantics of the ..._END macro. Most syntax aware editors will not show the implied scope created by the ..._BEGIN and ..._END macros. A macro something like what follows could be used to acquire the GIL in a { ... } scope.

    #define SPROKIT_PYTHON_GIL_SCOPED_ACQUIRE \
       pybind11::gil_scoped_acquire acquire;  (void)acquire

sprokit/src/bindings/python/modules/registration.cxx
```diff
@@ -70,6 +70,16 @@ static bool is_suppressed();
 static void _load_python_library_symbols();
 
 
+class scoped_save_thread
```
mattdawkins: Also minor, but you might want to move this into the python_util folder

KyleFromKitware: This class is just to work around the (possible) bug in pybind11, and I made it a class instead of just calling `PyEval_InitThreads()` outright so that it would still be called even if `gil_scoped_acquire` throws an exception. It will not, and should not, be used anywhere but here.

KyleFromKitware: An alternative would be:

```c++
try
{
  gil_scoped_acquire acquire;
}
catch( ... )
{
  PyEval_SaveThread();
  throw;
}
PyEval_SaveThread();
```

mattdawkins: Maybe a comment saying this then

sprokit/src/bindings/python/modules/registration.cxx
KyleFromKitware: Is the warning suppression strictly necessary? Getting rid of it would eliminate the need for macros entirely and simplify this a lot.

PR #477, opened 2018-02-16
Avoid running try_compile with each configure by using cache variables.

billhoffman:


dstoup:
I think this is a good branch and I'm inclined to merge. I would like to hear though if there is any good reason the USE* variables which are now getting passed into try_compile directly should not go into the cache. They were previously just getting pushed to parent_scope.

dstoup:
Jenkins test this please

PR #478, opened 2018-02-16
Force pybind11 to use the PyGILState_* API to prevent deadlocks

KyleFromKitware:
Note: this depends on Kitware/fletch#326.


KyleFromKitware:
Closed because I am going to take a new approach that doesn't involve patching pybind11.

PR #487, opened 2018-02-21
Use video frame rate in video_input_process

KyleFromKitware:


sprokit/processes/core/video_input_process.cxx
```diff
@@ -225,6 +239,7 @@ ::_step()
     push_to_port_using_trait( timestamp, ts );
     push_to_port_using_trait( image, frame );
     push_to_port_using_trait( metadata, metadata );
+    push_to_port_using_trait( frame_rate, d->m_video_reader->frame_rate() );
```
linus-sherrill: Not sure what the use case is for a port with the frame rate? Presumably it is constant?

mattdawkins: Not necessarily, but in this case it's so that a downsampler after the source can enter a target frame-rate (e.g 1hz)

vital/algo/video_input.h
```diff
@@ -288,6 +288,16 @@ class VITAL_ALGO_EXPORT video_input
...
+  /**
+   * \brief Get frame rate from the video.
+   *
+   * If frame rate is not supported, return -1.
+   *
+   * \return Frame rate.
+   */
+  virtual double frame_rate();
+
+
```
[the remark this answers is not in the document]
KyleFromKitware: OK, I've added this capability and set it for `vidl_ffmpeg_video_input`. The others can be implemented later on, but they're not important for what we're doing right now (see Dawkins' comment about the downsampler below.)

linus-sherrill:
jenkins test this please

sprokit/processes/core/video_input_process.cxx
```diff
@@ -195,7 +200,17 @@ ::_step()
       if ( ! d->m_video_traits.capability( kwiver::vital::algo::video_input::HAS_FRAME_TIME ) )
       {
         // create an internal time standard
-        d->m_frame_time = d->m_frame_number * d->m_config_frame_time;
+        double frame_rate = d->m_video_reader->frame_rate();
+        if( ! d->m_video_traits.capability( kwiver::vital::algo::video_input::HAS_FRAME_RATE ) ||
+            frame_rate <= 0.0 || d->m_has_config_frame_time )
+        {
+          d->m_frame_time = d->m_frame_number * d->m_config_frame_time;
```
mleotta: This seem like something the video source should be doing, not the sprokit process.  I would rather have the video source be in charge of figuring out the time for each frame.

linus-sherrill:
Have all comments been addressed? Also now have a merge conflict.

sprokit/processes/core/video_input_process.cxx
KyleFromKitware: It looks to me like the frame time configuration is supposed to be a process-level override. If it was happening at the video source level, it would have to be duplicated among all video source implementations.

Also, the old version was doing this in the process class too.

KyleFromKitware:
Merge conflict has been addressed. Waiting to hear Matt's comments on my response.

sprokit/processes/core/video_input_process.cxx
mleotta: okay, I understand now that this is supposed to be an option to override the time from the video source.  I guess that's okay.

linus-sherrill:
Jenkins test this please

linus-sherrill:
This request has been around for a while. Is it ready to be merged?

KyleFromKitware:
Yes, I think so.

linus-sherrill:
Jenkins test this please

PR #489, opened 2018-02-22
Respect PYTHON_VERSION_MAJOR and fix potential windows issues

Erotemic:
Due to this thread https://github.com/caffe2/caffe2/issues/1982 I realized there may be an issue with my `_pycmd` command on windows. Just in case there is I modified the places where it was being used to use one-liner commands instead of the multi-line commands. 

I also added logic so `KWIVER_PYTHON_MAJOR_VERSION` respects the variable `PYTHON_VERSION_MAJOR` if a program passes it on on the command line via `-D` (like the VIAME super build)

### DIFF
--- a/CMake/kwiver-setup-python.cmake
+++ b/CMake/kwiver-setup-python.cmake
@@ -85,7 +85,16 @@ endfunction()
 ###
 # Python major version user option
 #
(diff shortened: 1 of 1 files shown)


Erotemic:
this functionality was in the python 3 branch

PR #490, opened 2018-02-23
Support for Python 2 and 3

Erotemic:
This PR fixes issues that prevented KWIVER from being used with python3. 

After applying this patch all python tests pass except for 4:

```
The following tests FAILED:
	364 - test-python-export_-simple_pipeline (SEGFAULT)
	371 - test-python-pymodules-load (Failed)
	373 - test-python-pymodules-extra_modules (Failed)
	374 - test-python-pymodules-pythonpath (Failed)
```

@hughed2 Are these tests known to fail in python2 as well? or is there still a bit more work to be done?

### DIFF
--- a/CMake/kwiver-setup-python.cmake
+++ b/CMake/kwiver-setup-python.cmake
@@ -85,7 +85,16 @@ endfunction()
 ###
 # Python major version user option
 #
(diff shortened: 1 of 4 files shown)


hughed2:
A number of tests started getting seg faults recently, I'm not entirely sure why, or at what commit that started happening.

PR #494, opened 2018-03-02
Replace boost LCM with a local implementation.

linus-sherrill:
Remove other boost constructs.
Cleanup environment access code. Removed get_envvar() function since it has been replaced with kwiversys function.

### DIFF
--- a/sprokit/src/sprokit/pipeline/CMakeLists.txt
+++ b/sprokit/src/sprokit/pipeline/CMakeLists.txt
@@ -110,9 +110,7 @@ target_link_libraries(sprokit_pipeline
                       vital_logger
                       ${Boost_CHRONO_LIBRARY}
                       ${Boost_SYSTEM_LIBRARY}
-  PRIVATE             ${Boost_FILESYSTEM_LIBRARY}
-                      ${Boost_DATE_TIME_LIBRARY}
-                      ${Boost_THREAD_LIBRARY}
+  PRIVATE             ${Boost_THREAD_LIBRARY}
                       ${CMAKE_DL_LIBS}
                       ${CMAKE_THREAD_LIBS_INIT}
                       vital_vpm kwiversys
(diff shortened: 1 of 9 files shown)


sprokit/src/sprokit/pipeline/pipeline.cxx
```diff
@@ -61,8 +60,52 @@
...
+ * @param a First number
+ * @param b Second number
+ *
+ * @return GCD
+ */
+template <typename T>
+T
+gcd( T a, T b )
 {
+  for ( ; ; )
```
mleotta: Is this guaranteed to return?  I haven't thought about it enough to understand why or why not, but it looks like the type of code that could have a corner case that would lead to an infinite loop.   Is there any documentation on what this algorithm is called,  and how it works?  Maybe a link to a webpage or something?  I know boost have several different algorithms for this that it was selecting between.

sprokit/src/sprokit/pipeline/CMakeLists.txt
```diff
@@ -110,9 +110,7 @@ target_link_libraries(sprokit_pipeline
                       vital_logger
                       ${Boost_CHRONO_LIBRARY}
                       ${Boost_SYSTEM_LIBRARY}
-  PRIVATE             ${Boost_FILESYSTEM_LIBRARY}
-                      ${Boost_DATE_TIME_LIBRARY}
-                      ${Boost_THREAD_LIBRARY}
+  PRIVATE             ${Boost_THREAD_LIBRARY}
```
mleotta: This is not really a comment for this branch, but since I see it, what do we still need Boost Thread for?  I was under the impression that std::thread is a complete replacement for boost::thread.

sprokit/src/sprokit/pipeline/pipeline.cxx
linus-sherrill: I believe it will always converge. After each iteration, the values of a and b are reduced.
I can use Euclid's algorithm from Knuth so that the alg has a reference.

dstoup:
Jenkins test this please

PR #496, opened 2018-03-09
Disallow python symlinks on windows

hughed2:
cmake -E create_symlink doesn't really work on Windows. Ordinarily, we
could default to just turning KWIVER_SYMLINK_PYTHON off, and all would
be well. But this doesn't work for viame, which doesn't run through the
normal kwiver configuration.

### DIFF
--- a/CMake/utils/kwiver-utils-python.cmake
+++ b/CMake/utils/kwiver-utils-python.cmake
@@ -144,7 +144,7 @@ function (kwiver_add_python_module path     modpath    module)
   set(pypkg_install_path "${python_install_path}/${kwiver_python_subdir}/${python_sitename}/${modpath}")
 
   # copy and configure the source file into the binary directory
-  if (KWIVER_SYMLINK_PYTHON)
+  if (KWIVER_SYMLINK_PYTHON AND NOT WIN32)
     kwiver_symlink_file("python${python_arch}-${safe_modpath}-${module}"
       "${pyfile_src}"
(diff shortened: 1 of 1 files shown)


CMake/utils/kwiver-utils-python.cmake
```diff
@@ -144,7 +144,7 @@ function (kwiver_add_python_module path     modpath    module)
   set(pypkg_install_path "${python_install_path}/${kwiver_python_subdir}/${python_sitename}/${modpath}")
 
   # copy and configure the source file into the binary directory
-  if (KWIVER_SYMLINK_PYTHON)
+  if (KWIVER_SYMLINK_PYTHON AND NOT WIN32)
```
mattdawkins: Shouldn't you not be able to turn on KWIVER_SYMLINK_PYTHON if not windows?

We wouldn't want the user to be able to turn on SYMLINKs if it doesn't work (and then it unknowingly fails)

mattdawkins: if windows*

hughed2: You're actually not supposed to be able to. KWIVER_SYMLINK_PYTHON is accidentally set on Windows without user intervention.

I'm actually going to just close this--I'm testing a more robust fix anyway, and it's going to be in soon enough that I don't think there's a lot of value to do this more quick and dirty fix.

PR #497, opened 2018-03-14
Creation of new class for RPC camera model

chetnieter:
Introduces support for RPC camera model.


chetnieter:
@mleotta At this point this MR just includes the refactoring the creates a base camera (camera) and renames the perspective camera (camera_perspective). This involves lots of changes so I figure I would go ahead and put them out for review.  A have couple open questions:

1. I had to put in quite a few dynamic_cast and dynamic_pointer_cast for cases where a camera_map is used. I didn't really add any guards around these to catch null pointers from failed casts. I didn't do this  initially since I wasn't sure what if anything should be done in these cases.

2. I didn't rename the outward facing api's for the c and python bindings, I just replace camera with camera_perspective under the hood. Should I go ahead and change the names?

arrows/core/transform.cxx
```diff
@@ -193,7 +194,9 @@ necker_reverse(vital::camera_map_sptr& cameras,
   vital::rotation_d Rz180(vital::vector_4d(0.0, 0.0, 1.0, 0.0));
   for(cam_map_t::value_type& p : cams)
   {
-    vital::simple_camera* flipped = new vital::simple_camera(*p.second);
```
mleotta: this code was like from before I knew about `make_shared`, so it could be reworked to avoid the raw pointer and `new` call.

vital/types/camera.h
```diff
@@ -40,242 +40,38 @@
...
 #include <vital/types/vector.h>
-#include <vital/types/similarity.h>
-#include <vital/logger/logger.h>
 
 
 namespace kwiver {
 namespace vital {
 
-/// forward declaration of camera class
+/// forward declaration of perspective camera class
```
mleotta: This one should not say perspective

vital/types/camera.h
```diff
@@ -40,242 +40,38 @@
...
 class camera;
 /// typedef for a camera shared pointer
 typedef std::shared_ptr< camera > camera_sptr;
 
 
 // ------------------------------------------------------------------
-/// An abstract representation of camera
+/// An abstract representation of perspective camera
```
mleotta: This one should not say perspective

vital/types/camera.h
```diff
@@ -40,242 +40,38 @@
...
+ * The base class of cameras.
  */
 class VITAL_EXPORT camera
 {
 public:
   /// Destructor
   virtual ~camera() = default;
 
-  /// Create a clone of this camera object
-  virtual camera_sptr clone() const = 0;
```
mleotta: We probably still want to keep `clone` in the base class so we can deep copy cameras without needing to know type

arrows/core/transform.cxx
chetnieter: Since the flipped camera is written back to the camera map is there any reason for a copy to be made or can we just grab a pointer and change the camera in place.

chetnieter:
@mleotta I think I am at a good point to merge into master. I still need add a gdal arrow for importing an image but the rpc_camera is done and I have a helper method to create one from vital metadata. Both the rpc_camera and the helper methods have testing in place. It looks like the CI checks are failing to find fletch.

mleotta:
Thanks.  I'll try to find time to look at it soon, but will probably be a bit distracted preparing for the CORE3D site visit next week.  Of course, you can get feedback from others too.

I've also sent out a call for help in trying to fix KWIVER master.  We need master to build or the CI is useless.

arrows/core/transform.cxx
mleotta: I think it's safer to make a copy.  The way this function is often used is to keep a pointer to the original camera map and then pass a pointer to that camera map into this function, which returns a new reversed camera map its place.  We then often use both the original cameras and the reverse cameras to compare and see which is better.  If we change the cameras in place it will corrupt our copy of the original cameras.   Of course we could work around this by requiring the user to explicitly clone the camera map before passing to this function, but I think it's easier to just do the copy in this function until we have a reason to do otherwise.

vital/io/camera_from_metadata.cxx
```diff
@@ -0,0 +1,169 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * \file
+ * \brief Implementation of file IO functions for a \ref kwiver::vital::camera
```
mleotta: comment needs to be updated

vital/io/camera_from_metadata.cxx
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * File format is the KRTD file.
```
mleotta: same here

vital/io/camera_from_metadata.h
```diff
@@ -0,0 +1,66 @@
...
+ */
+Eigen::VectorXd
+VITAL_EXPORT string_to_vector( std::string const& s );
+
+/// Produce RPC camera from metadata
+/**
+ * \param file_path   The path to the file to read in.
+ * \return A new camera object representing the contents of the read-in file.
+ */
+camera_rpc_sptr
```
mleotta: I think the return type here should be a base class `camera_sptr`.  For now this just converts RPC metadata, but we should later port over the code from MAP-Tk for perspective cameras from metadata and put it in this same function.  The function would have to inspect the available metadata tags and determine which camera to create.  We don't need to do that on this branch, but using the base camera as a return type now will allow for this change later without needing to change the API.

chetnieter:
@mleotta Fixed the latest issues you pointed out. Builds for me on both mac (clang) and linux (gcc). Ran all the tests with camera in their name (ctest -R camera). They all pass on Linux. One fails on Mac (ceres:optimize_cameras.noisy_cameras) but it also fails for master.

vital/types/camera_rpc.h
```diff
@@ -0,0 +1,183 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * \file
+ * \brief Header for \link kwiver::vital::camera_perspective camera_perspective \endlink and
```
mleotta: Update names in comment

vital/types/camera_rpc.h
```diff
@@ -0,0 +1,183 @@
...
+#include <vital/types/vector.h>
+#include <vital/logger/logger.h>
+
+
+namespace kwiver {
+namespace vital {
+
+typedef Eigen::Matrix< double, 4, 20 > rpc_matrix;
+
+/// forward declaration of perspective camera class
```
mleotta: update comment

vital/types/camera_rpc.h
```diff
@@ -0,0 +1,183 @@
...
+class camera_rpc;
+/// typedef for a camera_rpc shared pointer
+typedef std::shared_ptr< camera_rpc > camera_rpc_sptr;
+
+
+// ------------------------------------------------------------------
+/// An abstract representation of perspective camera
```
mleotta: update comment

vital/types/camera_rpc.h
```diff
@@ -0,0 +1,183 @@
...
+  camera_rpc();
+
+  kwiver::vital::logger_handle_t m_logger;
+
+};
+
+
+/// A representation of a camera
+/**
+ * Contains camera location, orientation, and intrinsics
```
mleotta: update comment

vital/types/camera_rpc.h
```diff
@@ -0,0 +1,183 @@
...
+    world_scale_( base.world_scale() ),
+    world_offset_( base.world_offset() ),
+    image_scale_( base.image_scale() ),
+    image_offset_( base.image_offset() ),
+    rpc_coeffs_( base.rpc_coeffs() )
+  { }
+
+  /// Create a clone of this camera object
+  virtual camera_sptr clone() const
+  { return camera_sptr( new simple_camera_rpc( *this ) ); }
```
mleotta: it would be better to use `std::make_shared<simple_camera_rpc>( *this )` here.  I know `new` is used all over the place for making cameras, but we should probably use make_shared at least on the new uses (and maybe update the rest later).

chetnieter:
@mleotta Finished the rebase and I addressed the issues you raised two day ago. I believe we are ready to merge.

mleotta:
@chetnieter Can you add release notes to `doc/release-notes/master.txt` noting the refactoring of the camera class and the addition of the new RPC camera?  I will merge after the release notes are added.

PR #502, opened 2018-03-28
Fix the fletch_DIR used by travis

dstoup:
The MR is a response to a change in fletch which relocated its config file to a more sensible location.


dstoup:
Jenkins test this please

dstoup:
This patch fixes the issue which it was intended. It sets the correct fletch_DIR. There are still build issues though which is why things still fail.

PR #506, opened 2018-03-30
Fix include paths issue introduced by 0f2404d

sankhesh:
Fix syntax issue in `foreach`


sankhesh:
@mleotta @linus-sherrill Minor change that should be merged into release and then release into master.

PR #514, opened 2018-04-11
_USE_MATH_DEFINES needs to come first so we have a definition of M_PI…

dstoup:
_USE_MATH_DEFINES needs to come first so we have a definition of M_PI on Windows

### DIFF
--- a/arrows/core/tests/test_interpolate_camera.cxx
+++ b/arrows/core/tests/test_interpolate_camera.cxx
@@ -1,5 +1,5 @@
 /*ckwg +29
- * Copyright 2014-2017 by Kitware, Inc.
+ * Copyright 2014-2018 by Kitware, Inc.
  * All rights reserved.
  *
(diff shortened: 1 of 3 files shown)


mleotta:
My preference would be to avoid using these non-standard definitions like `M_PI` altogether.  They may have slightly different definitions (e.g. different levels of precision) across platforms/compilers.  There are no standard definitions of numerical constants, like pi, in C++.   It would be better to just define some `constexpr` values for use in KWIVER.  Maybe we need a math_constants.h in vital.

dstoup:
@mleotta do you propose we do that as part of this branch, or can we merge this one to get Windows building and replace those constants in a follow-up branch?

mleotta:
It's fine to do it in a separate branch.  I understand the need to get the dashboards green first.

dstoup:
Jenkins test this please

PR #520, opened 2018-04-16
Simply the kwiver-config generation.

dstoup:
The branch fixes the odd entries in kwiver-config-install.cmake and kwiver-config.cmake. Currently the Eigen3_INCLUDE_DIR and the fletch_INCLUDE_DIR are munged together so eigen cannot be found. Also the KWIVER_LIBRARIES values all have a trailing /.

This branch is currently up for evaluation only. I didn't really finish it yet. It needs to have more of the stuff that the suspect function is doing, removed. I didn't want to go too far down that road if I was missing something.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -325,39 +325,20 @@ if (fletch_FOUND)
   list( APPEND KWIVER_LIBRARY_DIRS "${fletch_ROOT}/lib")
 endif()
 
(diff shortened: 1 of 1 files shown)


dstoup:
Since it looks ok. I will see how much of that function can go away and if
i can write a test to demonstrate the fix. I will tackle that in the
morning.

PR #525, opened 2018-04-17
Remove old deprecated python file

mattdawkins:
This is not needed after all the pybind stuff landed, but had some changes in a recent large branch which just landed.


dstoup:
Jenkins test this please

PR #526, opened 2018-04-18
Fix failing process introspection test.

linus-sherrill:
- The process that failed did so because of its design. It
  created ports dynamically.

- Mark this process as not testable.

- Add comments about why perfectly good processes can fail some of these
  tests and how to mark them as not testable.


dstoup:
Looks good to me and already sees the test passing.

PR #533, opened 2018-04-27
Ffmpeg arrow

vovythevov:
Available features: 
 - Opening/Closing a video
 - Extracting frame information

@mleotta


arrows/ffmpeg/tests/decode_barcode.h
```diff
@@ -0,0 +1,76 @@
...
+// Ignore 8 pixels on either side of the barcode
+static int bc_buffer = 8;
+// Barcode lines two pixels wide and 4 pixels high
+static int bc_width = 2;
+static int bc_height = 4;
+static int bit_depth = 256;
+static int bc_area = bc_width*bc_height;
+
+// Decode barcodes from test frame images
+uint32_t decode_barcode(kwiver::vital::image_container_sptr img_ct)
```
mleotta: We should probably share a single copy of this code, and possibly the test data, between this and the VXL tests.  Note the VXL barcode reader is in the master branch now, you may need to merge in master to see it.  We could put the common code in core/tests for now, but really we should create a better place for common tests, data, and utility functions.

arrows/ffmpeg/ffmpeg_video_input.h
```diff
@@ -0,0 +1,92 @@
...
+#include <vital/algo/video_input.h>
+
+#include <arrows/ffmpeg/kwiver_algo_ffmpeg_export.h>
+
+
+namespace kwiver {
+namespace arrows {
+namespace ffmpeg {
+
+/// Video input using VXL vidl ffmpeg services.
```
mleotta: update comments to remove VXL references

arrows/ffmpeg/ffmpeg_video_input.h
```diff
@@ -0,0 +1,92 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * \file
+ * \brief \todo
+ */
+
+#ifndef KWIVER_ARROWS_FFMPEG_ffmpeg_video_input_H
```
mleotta: use all upper case in include guard

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,772 @@
...
+
+#include <kwiversys/SystemTools.hxx>
+
+#include <mutex>
+#include <memory>
+#include <vector>
+#include <sstream>
+
+extern "C" {
+#if FFMPEG_IN_SEVERAL_DIRECTORIES
```
mleotta: We don't need many old versions of FFMPEG like VXL does.  The main focus is the version in Fletch.  I'm not sure how long ago this include path change happened, but if it was several years ago it's likely we don't need to support the old way.

arrows/ffmpeg/ffmpeg_init.h
```diff
@@ -0,0 +1,43 @@
...
+* CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+* OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+* OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef ffmpeg_init_h_
+#define ffmpeg_init_h_
+
+//: Initialize the ffmpeg codecs.
+//
```
linus-sherrill: This doesn't look like doxygen format documentation.

arrows/ffmpeg/ffmpeg_init.cxx
```diff
@@ -0,0 +1,55 @@
...
+#endif
+}
+
+//--------------------------------------------------------------------------------
+
+void ffmpeg_init()
+{
+  static bool initialized = false;
+  if ( ! initialized ) {
+    av_register_all();
```
linus-sherrill: I'm not sure how this will be used, but are there any race conditions if two threads try to initialize at the same time?

arrows/ffmpeg/ffmpeg_video_input.h
```diff
@@ -0,0 +1,92 @@
...
+/// Video input using VXL vidl ffmpeg services.
+// ----------------------------------------------------------------
+/**
+ * This class implements a video input algorithm using the VXL vidl
+ * ffmpeg video services.
+ *
+ */
+class KWIVER_ALGO_FFMPEG_EXPORT ffmpeg_video_input
+  : public vital::algorithm_impl < ffmpeg_video_input, vital::algo::video_input >
+{
```
linus-sherrill: Since this is derived from video_input it should support algorithm capabilities.

CMake/FindFFMPEG.cmake
```diff
@@ -0,0 +1,89 @@
...
+#
+# Sets
+#   FFMPEG_FOUND.  If false, don't try to use ffmpeg
+#   FFMPEG_FOUND_SEVERAL. If true, there are several directories with headers (not only ../ffmpeg/)
+#   FFMPEG_INCLUDE_DIR
+#   FFMPEG_LIBRARIES
+
+set( FFMPEG_FOUND "NO" )
+
+find_path( FFMPEG_INCLUDE1_DIR ffmpeg/avcodec.h
```
dstoup: Since this is kwiver code, should we add some priming of paths from fletch_DIR? We probably want to give preference to that package to ensure they continue to work together.

CMake/kwiver-depends-ffmpeg.cmake
```diff
@@ -0,0 +1,10 @@
+# Optional find and confgure VXL dependency
+
+option( KWIVER_ENABLE_FFMPEG
+  "Enable FFMPEG dependent code and plugins (Arrows)"
+  ${fletch_ENABLED_FFMPEG}
+  )
+
+if( KWIVER_ENABLE_FFMPEG )
+  find_package( FFMPEG REQUIRED )
```
dstoup: Do we care about minimum version of FFmpeg at this point?

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,834 @@
...
+      - static_cast<int>(this->f_frame_number_offset));
+  }
+
+  // ==================================================================
+  /*
+  * @brief Convert a ffmpeg pixel type to kwiver::vital::image_pixel_traits
+  *
+  * @return \b A kwiver::vital::image_pixel_traits
+  */
+  pixel_format pixel_format_from_ffmpeg(AVPixelFormat ffmpeg_format) const
```
mleotta: I suggest simplifying this quite a bit and removing this function and the `pixel_format` enum.  The VXL vidl library was designed to be able to store these various video frame encodings in raw form.  Most of these are packed formats with different resolutions in the different color channels that `vital_image` cannot support.  In our case we should always ask FFMPEG to convert the image into RGB24 for us.  The only exceptions would be
- `RGBA` - create a 4 channel image 
- `GRAY8` - create a 1 channel image
- `MONOWHITE` and `MONOBLACK` - create a 1 channel boolean image

Note that it is quite rare to find videos encoded with these.  Most videos use some variant of YUV.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,834 @@
...
+    return this->f_frame && this->f_frame->data[0];
+  }
+
+  // ==================================================================
+  /*
+  * @brief Return the current frame number
+  *
+  * @return \b Current frame number.
+  */
+  unsigned int frame_number() const
```
mleotta: same comment as above.  We should keep a frame counter (how many times you advanced) and return that.  This gets tricky when dealing with seeking.  It may require a novel solution to make seeking reliably frame accurate.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,834 @@
...
+  {
+    int width = enc->width;
+    int height = enc->height;
+
+    // If the pixel format is not recognized by then convert the data into RGB_24
+    ffmpeg_video_input::priv::pixel_format format = d->pixel_format_from_ffmpeg(enc->pix_fmt);
+    if (format == ffmpeg_video_input::priv::PIXEL_FORMAT_UNKNOWN)
+    {
+      int size = width * height * 3;
+      if (!d->current_image_memory || size != d->current_image_memory->size())
```
mleotta: We probably don't want to reuse the same image memory for decoding each frame.  This is great for efficiency, but will corrupt any image that an outside user is still holding on to.  At this point we should just allocate a new image for each new frame.  You can probably just make the image directly and let it allocate it's own memory and access the memory instead of needing to allocate the memory and building an image object around it.

If this memory allocation becomes a bottleneck, I recommend creating a special image memory allocator that builds a pool of image image memory objects such that the memory gets returned to the pool instead of deallocated when the last shared point to an image goes out of scope.  However, this optimization is complex and should be put off until it is needed.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,834 @@
...
+      // Copy the image into contiguous memory.
+      else
+      {
+        if (!d->current_image_memory)
+        {
+          int size = avpicture_get_size(enc->pix_fmt, width, height);
+          d->current_image_memory = vital::image_memory_sptr(new vital::image_memory(size));
+        }
+        avpicture_fill(&test_frame, (uint8_t*)d->current_image_memory->data(), enc->pix_fmt, width, height);
+        av_picture_copy(&test_frame, (AVPicture*)d->f_frame, enc->pix_fmt, width, height);
```
mleotta: This case is okay, but only in those cases where the data is in a recognized format and only needs too be copied: RGB24, RGBA, GRAY8, MONOWHITE, MONOBLACK

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,834 @@
...
+          width, height, 3,
+          3, 3 * width, 3 * width*height//,
+          //image_pixel_from_traits_macro from fmt ?
+        );
+      d->current_image = std::make_shared<vital::simple_image_container>(vital::simple_image_container(image));
+      }
+      // Copy the image into contiguous memory.
+      else
+      {
+        if (!d->current_image_memory)
```
mleotta: I would just store `d->current_image` and not `d->current_image_memory`, and always create a new image for each new frame.  The image memory is kind of redundant to store since it's inside the image.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,834 @@
...
+  {
+    return this->f_pts * av_q2d(this->f_video_stream->time_base);
+  }
+
+  // ==================================================================
+  /*
+  * @brief Returns the double value to convert from a stream time base to
+  *  a frame number
+  */
+  double stream_time_base_to_frame() const
```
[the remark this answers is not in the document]
vovythevov: I think we should improve this in a separate branch that improves the seeking in itself.

vovythevov:
@chetnieter: If you want to take a look :)

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,749 @@
...
+    kwiver::vital::simple_metadata_map());
+}
+
+
+// ------------------------------------------------------------------
+bool
+ffmpeg_video_input
+::end_of_video() const
+{
+  return false;
```
chetnieter: Is there a way to tell from the FFmpeg API if we are at the end of the video? It looks like the old VXL implementation set a flag to mark it at the end if the decoding failed but I am not sure if that is the correct approach.

arrows/ffmpeg/tests/test_open_video.cxx
```diff
@@ -0,0 +1,167 @@
+/*ckwg +29
```
chetnieter: Maybe rename file to `test_video_input_ffmpeg.cxx` since the tests cover more than just opening the video.

arrows/ffmpeg/ffmpeg_video_input.cxx
chetnieter: Please take a look at the VXL implementation and reproduce here since the `end_of_video()` method is used in multiple places in both maptk and kwiver.

arrows/ffmpeg/tests/test_video_input_ffmpeg.cxx
```diff
@@ -0,0 +1,209 @@
...
+  }
+  // Test various invalid seeks past end of video
+  std::vector<kwiver::vital::timestamp::frame_t> in_valid_seeks =
+  { -3, -1, 0, 51, 55 };
+  for (auto requested_frame : in_valid_seeks)
+  {
+    EXPECT_FALSE(input.seek_frame(ts, requested_frame));
+    EXPECT_NE(requested_frame, ts.get_frame());
+  }
+}
```
chetnieter: Please reproduce the "is_good" and "read_video" tests from the VXL video reader [tests](https://github.com/Kitware/kwiver/blob/master/arrows/vxl/tests/test_vidl_ffmpeg_video_input.cxx).

vovythevov:
@chetnieter: Bump :)

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,826 @@
...
+    return true;
+  }
+
+  // ==================================================================
+  /*
+  * @brief Advance to the next frame (but don't acquire an image).
+  *
+  * @return \b true if video was valid and we found a frame.
+  */
+  virtual bool advance()
```
chetnieter: Is there a reason this method in the private data class is virtual? It is causing compilation to fail with clang on MacOS.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,749 @@
...
+        break;
+      }
+      case AV_PIX_FMT_MONOWHITE:
+      case AV_PIX_FMT_MONOBLACK:
+      {
+        num_pixels = 1;
+        pixel_trait = vital::image_pixel_traits_of<bool>();
+        direct_copy = true;
+        break;
+      }
```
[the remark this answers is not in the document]
chetnieter: The lack of a `default` is producing a warning. "warning: 167 enumeration values not handled in switch: ...". Could we add a default block that just sets `direct_copy = false` to get rid of these compiler warnings.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,826 @@
...
+  virtual bool advance()
+  {
+    // Quick return if the file isn't open.
+    if (!this->is_opened())
+    {
+      this->frame_advanced = 0;
+      return false;
+    }
+
+    // \todo - num_frames not implemented
```
chetnieter: I think this should just be removed since `num_frames` is not implemented. It probably makes sense to add a TODO or better yet open an issue to remind us to improved how the number of frames is reported but this comment is confusing since `num_frames` is not available.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,826 @@
...
+    av_freep(&d->f_frame);
+  }
+  d->f_frame = nullptr;
+
+  if (d->f_video_encoding && d->f_video_encoding->opaque)
+  {
+    av_freep(&d->f_video_encoding->opaque);
+  }
+
+  //d->num_frames_ = -2;
```
chetnieter: Remove this is it is no longer relevant nor is it like to be used in the future.

chetnieter:
Found a few new minor issues but other than that things look good. I noticed there are some outstanding issues raised by others, notable @linus-sherrill and @dstoup that haven't been resolved. Given the funding situation I think it is ok we push dealing with those after merging but I think we should either open a issue or mark with TODO's so they don't get forgotten.

arrows/ffmpeg/ffmpeg_video_input.cxx
linus-sherrill: If it is the private class it seems unlikely that it will be used as a base class. I would agree with removing the virtual.

vovythevov:
@chetnieter: Thanks for the feedback !
I updated my code to reflect your comments.
As for @linus-sherrill remark's, I rebased the code which and added all the new necessary methods. Is that what I needed to do ?
As for @dstoup remark's, I did not want to change the FindFFMPEG as it doesn't seem that other Find* prime their paths with Fletch variables.

vovythevov:
@mleotta: I pushed the algorithm capabilities commit.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,832 @@
...
+      }
+      default:
+      {
+        direct_copy = false;
+      }
+    }
+    if (direct_copy)
+    {
+      int size = avpicture_get_size(enc->pix_fmt, width, height);
+      d->current_image_memory = vital::image_memory_sptr(new vital::image_memory(size));
```
mleotta: it would be safer to use ` std::make_shared<vital::image_memory>(size);` here.

Longer term we might want to explore recycling image memory instead of allocating it for each frame.  However, that is a bit tricky and should certainly wait until later.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,832 @@
...
+
+      AVPicture frame;
+      avpicture_fill(&frame, (uint8_t*)d->current_image_memory->data(), enc->pix_fmt, width, height);
+      av_picture_copy(&frame, (AVPicture*)d->f_frame, enc->pix_fmt, width, height);
+    }
+    else
+    {
+      int size = width * height * num_pixels;
+      d->current_image_memory = vital::image_memory_sptr(new vital::image_memory(size));
```
mleotta: it would be safer to use ` std::make_shared<vital::image_memory>(size);` here.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,832 @@
...
+  }
+
+  AVCodecContext* enc = d->f_format_context->streams[d->f_video_index]->codec;
+
+  // If we have not already converted this frame, try to convert it
+  if (!d->current_image_memory && d->f_frame->data[0] != 0)
+  {
+    int width = enc->width;
+    int height = enc->height;
+    int num_pixels = 3;
```
mleotta: `num_pixels` is a bit misleading for this variable name.  I would call it `num_channels` or `depth`.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,832 @@
...
+      {
+        num_pixels = 4;
+        direct_copy = true;
+        break;
+      }
+      case AV_PIX_FMT_MONOWHITE:
+      case AV_PIX_FMT_MONOBLACK:
+      {
+        num_pixels = 1;
+        pixel_trait = vital::image_pixel_traits_of<bool>();
```
mleotta: I'm not sure if mono images are actually stored as an array of `bool` or or not.  It would be interesting to test.  I don't know that we have any test data though.

arrows/ffmpeg/ffmpeg_video_input.h
```diff
@@ -0,0 +1,106 @@
...
+
+  virtual void open( std::string video_name );
+  virtual void close();
+
+  virtual bool end_of_video() const;
+  virtual bool good() const;
+
+  virtual bool seekable() const;
+  virtual size_t num_frames() const
+    {
```
mleotta: remove extra indentation here.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,832 @@
...
+{
+  if (!this->good())
+  {
+    return {};
+  }
+
+  // We don't always have all components of a timestamp, so start with
+  // an invalid TS and add the data we have.
+  kwiver::vital::timestamp ts;
+  ts.set_frame(d->frame_number() + d->f_frame_number_offset + 1);
```
mleotta: Do not use this `frame_number()` function to compute the current frame.   It is not reliable.  We should keep an internal counter indicating the last frame we stepped to and return that instead.

arrows/ffmpeg/ffmpeg_video_input.cxx
mleotta: Again, I would remove this function.  It has been shown to be broken in VXL and we don't use it in `vidl_ffmpeg_video_input`.  We remember the frame number in a member variable instead.

CMake/kwiver-depends-ffmpeg.cmake
mleotta: This change could be done in a follow-up branch.

arrows/ffmpeg/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+# Build / Install plugin containing ffmpeg algorithm implementations + support
+# structures
+
+include_directories(${FFMPEG_INCLUDE_DIR})
+
+if (NOT FFMPEG_FOUND_SEVERAL)
+  message( FATAL_ERROR "FFMPEG headers all in the same directories, "
```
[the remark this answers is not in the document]
mleotta: This change could be done in a follow-up branch.

arrows/ffmpeg/ffmpeg_video_input.cxx
vovythevov: @mleotta: Do you want to solve this before landing the branch ?

arrows/ffmpeg/ffmpeg_video_input.h
vovythevov: Solved with https://github.com/Kitware/kwiver/pull/533/commits/a1384a87ec522d79109bf6abf29f1515c2ff40a1

arrows/ffmpeg/ffmpeg_video_input.cxx
vovythevov: Good catch ! Solved with https://github.com/Kitware/kwiver/pull/533/commits/a1384a87ec522d79109bf6abf29f1515c2ff40a1

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -0,0 +1,832 @@
...
+  kwiver::vital::image_container_sptr current_image;
+
+  // local state
+  int frame_advanced; // This is a boolean check value really
+  bool end_of_video;
+  size_t number_of_frames;
+  bool number_of_frames_valid;
+
+  // ==================================================================
+  /*
```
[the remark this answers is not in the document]
vovythevov: https://github.com/Kitware/kwiver/pull/533/commits/a1384a87ec522d79109bf6abf29f1515c2ff40a1

arrows/ffmpeg/ffmpeg_video_input.cxx
vovythevov: Solved with https://github.com/Kitware/kwiver/pull/533/commits/6983d5a90f5ebbe5c0bd184b73c33d002f35482d

arrows/ffmpeg/ffmpeg_video_input.cxx
mleotta: Let's land the branch as is and the follow up with improvements

PR #535, opened 2018-05-03
add vital/exceptions.h back to CMakeLists.txt

mleotta:
This header was accidentally removed from CMakeLists when moving the
exceptions subdirectory to its own library.  This means exceptions.h
is no longer installed and it breaks projects building agains installed
KWIVER.


mleotta:
Can someone give me a +1 on this change.  I need it to fix the MAP-Tk Travis-CI build.  All of the dashboards here are failing due to known test failures.

PR #537, opened 2018-05-07
Fix KWA writer example pipeline.

linus-sherrill:
Update pipe file to new syntax. Fix CMake file configure to use @ONLY.


aaron-bray:
This is a good change. Safer with the @ONLY

PR #516, opened 2018-04-13
Fixed broken CSV detection reader

russell-david:
In cases where a CSV file of detections was not sorted by image, the reader would fail to correctly match the detections to the appropriate images. I fixed this in a similar fashion to how Matt D. did it for HABCAM inputs, by pre-computing a map from image names to a set of detections.

PR #539, opened 2018-05-08
[WIP] Fixed broken CSV detection reader (#516)

linus-sherrill:
This PR is a second pass at modifying the detected object reader, CSV format.

* Improved the process of extracting detections per frame
* improved commenting and naming
* Only create map when needed. Should address CI issues
* Only create map when needed. Should address conflicts
* Tidied up from the last commit
* Reverted last three commits and hopefully fixed CI errors
* Changed the logger
* Changed to the correct logger

### DIFF
--- a/arrows/core/detected_object_set_input_csv.cxx
+++ b/arrows/core/detected_object_set_input_csv.cxx
@@ -73,7 +73,8 @@ public:
 
   bool get_input();
   void add_detection();
-
+  void add_detection( const std::vector< std::string >& parsed_line);
+  void read_all();
 
   // -------------------------------------
   detected_object_set_input_csv* m_parent;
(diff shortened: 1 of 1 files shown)


linus-sherrill:
These changes introduce a semantic change in the abstract
class. Specifically the read_set() method. Originally the `image_name`
was an output parameter. This change to the CVS format reader makes it
an input/output parameter with special handling if the string is
empty. This semantic change would need to be propagated to all other
implementations of this reader. Currently only the CSV reader works
this way.

dstoup:
The core:detected_object_io.stream_io test still fails on Linux

dstoup:
... aside from the Windows build failures.

mattdawkins:
FYI we have already propogated this change to the same CSV readers in VIAME (habcam and noaa csv)

PR #540, opened 2018-05-09
Updating ocv arrow detection output filename to include chip coords

hughed2:
If the string format in the config has more than one input, it now tries to include top left x, top left y, width, and height before the id. So we can do something like `detection_%05d` to only produce id, as before, or we can do `detection_%05d_%fx%f+%f+%f` to include the coordinates.


linus-sherrill:
I looked through the source file and the file name pattern is not documented anywhere. Would you include the file name pattern options in the config  item description so it is clear what the expected patterns are. Putting unexpected format specifiers (such as %s) could result in unexpected behaviour.

PR #541, opened 2018-05-10
Remove unused boost_system_library links

dstoup:
Boost_SYSTEM_LIBRARY is linked throughout kwiver but never used. Some of those links are public which pollutes the link lines of consumers. This branch removes them all.


dstoup:
This branch needs to be redone, if at all. There is a tangle of deps brought in by Boost_Filesystem which is pulling in Boost_System.

PR #553, opened 2018-07-05
add docker feature

Cookt2:


docker/dockerfile
```diff
@@ -0,0 +1,63 @@
...
+&& tar -xvf cmake-3.11.3.tar.gz \
+&& cd cmake-3.11.3 \
+&& ./configure \
+&& make -j8 -k \
+
+#
+# Building Fletch
+#
+
+&& mkdir /fletch /fletch/build /fletch/build/rel /fletch/build/deb \
```
dstoup: Doesn't look like we use /fletch/build/deb

docker/dockerfile
```diff
@@ -0,0 +1,63 @@
...
+&& git checkout release \ 
+&& cd /fletch/build/rel \ 
+&& cmake -DCMAKE_BUILD_TYPE=Release -Dfletch_ENABLE_ALL_PACKAGES=ON -Dfletch_ENABLE_PYTHON=ON ../../src \
+&& make -j8 -k \
+
+#
+# Building Kwiver
+#
+
+&& mkdir /kwiver /kwiver/src /kwiver/build /kwiver/build/release /kwiver/build/debug \
```
dstoup: Also looks like we don't use /kwiver/build/debug

docker/dockerfile
dstoup: Also, you can do mkdir in 1 step with mkdir -p /fletch/build/rel

docker/dockerfile
dstoup: There might be some debate, but I contend that we don't really need to use the fletch build source or build products anymore, once this is done. You should probably set the fletch_BUILD_INSTALL_PREFIX to a directory which will persist and rm -rf /fletch/build/rel  /fletch/src so the image size doesn't explode.

PR #554, opened 2018-07-05
Add in non-maximum suppression process

hughed2:


hughed2:
Switching to use an algo

PR #563, opened 2018-07-19
Removing use of auto keyword with Eigen data types.

chetnieter:
Removing a few more cases of where the auto keyword was being used with Eigen data types. The use of the auto keyword with Eigen data types is known to be problematic.

https://eigen.tuxfamily.org/dox/TopicPitfalls.html


dstoup:
Jenkins test this please

chetnieter:
@mleotta I am happy to drop the changes if we are confident they are not needed. The auto/Eigen issue has tripped me up multiple times and as we learned can cause cross-platform failures which can be hard to track down so I guess I was being overly cautious.

PR #564, opened 2018-07-19
Resolve ambiguous type when including pybind11

aaron-bray:
pybind11 seems to make its own alias for size_t in its own namespace
They both resolve to the same thing, but msvc gets confused for some reason

Here is the definition in question

https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/common.h#L298


aaron-bray:
Note I only get this with python bindings turned on for kwiver (and fletch python/pybind support)

linus-sherrill:
There looks like no downside to namespacing the type.

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #566, opened 2018-07-20
Replace sprokit python init with kwiver

dstoup:
This branch is made of 3 commits. The first, fixes a configuration issue when sprokit is not enabled and python is. Vital/types was using a sprokit version of the create_python_init function when a kwiver version is available.

Considering the desire to reduce maintenance on things like this, the second commit replaces all uses of the sprokit_create_python_init with kwiver_create_python_init.

The third commit removes the sprokit version altogether so no one else is tempted to use it.

I made this PR in 3 commits in the event someone has an objection about the wholesale replacement/removal of sprokit_create_python_init.


dstoup:
@Erotemic , can you have a look too since you wrote some bits of sprokit_create_python_init?

Erotemic:
The sprokit version is slightly different from the kwiver version

The kwiver version:
```cmake
function (kwiver_create_python_init    modpath)
  _kwiver_create_safe_modpath("${modpath}" safe_modpath)

  set(init_template "${CMAKE_CURRENT_BINARY_DIR}/${safe_modpath}.__init__.py")

  if (NOT copyright_header)
    set(copyright_header "# Generated by kwiver")
  endif ()

  file(WRITE "${init_template}"      "${copyright_header}\n\n")

  foreach (module IN LISTS ARGN)
    file(APPEND "${init_template}"      "from .${module} import *\n")
  endforeach ()

  kwiver_add_python_module("${init_template}"
    "${modpath}"
    __init__)
endfunction ()
```

And the sprokit version:

```cmake
###
# Creates a default __init__.py file for a core package in the build directory.
#
function (sprokit_create_python_init    modpath)
  _kwiver_create_safe_modpath("${modpath}" safe_modpath)

  set(noarch_suffix)
  if (python_noarch)
    set(noarch_suffix "u")
  endif ()

  set(init_template "${CMAKE_CURRENT_BINARY_DIR}/${noarch_suffix}${safe_modpath}.__init__.py")

  if (NOT copyright_header)
    set(copyright_header "# Generated by sprokit")
  endif ()

  file(WRITE "${init_template}"
    "${copyright_header}\n\n")

  file(APPEND "${init_template}"
    "from __future__ import absolute_import\n\n")

  foreach (module IN LISTS ARGN)
    file(APPEND "${init_template}"
      "from .${module} import *\n")
  endforeach ()

  sprokit_add_python_module("${init_template}"
    "${modpath}"
    __init__)
endfunction ()
```

I think the lines:
```cmake
  file(APPEND "${init_template}"
    "from __future__ import absolute_import\n\n")
```

should probably be incorporated into the kwiver version. 

The other major difference is the noarch stuff, which I still don't quite understand.

dstoup:
@Erotemic I've added the from __future__ import calls. I will wait for some clarification as to why we should be prefixing the patch with a u in noarch scenarios. Anyone else with insight?

Purg:
```python
from .${module} import *
```

*twitch*

Erotemic:
@Purg Perhaps we can avoid the `import *` by explicitly including an `__init__.py` file in the source dir that lists all of the compiled modules as well as the public members that we expect to be exposed? Doing this would be more explicit and remove the need to programatically create the `__init__.py` file at build time. 

That may look something (the following is non-exhaustive) like this: 

```python
from scheduler_factory import (add_scheduler, create_scheduler,
                               is_scheduler_module_loaded,
                               mark_scheduler_module_as_loaded, types,
                               description, default_type)
from modules import load_known_modules
from process_factory import (is_process_module_loaded,
                             mark_process_module_as_loaded, add_process,
                             create_process, description, types,)
from datum import (Datum, DatumType, new, new_int, new_float, new_string, new_image_container,
                   new_descriptor_set, new_detected_object_set, new_track_set,
                   new_double_vector, new_string_vector, datum_from_capsule,
                   empty, flush, complete, error)
from edge import Edge, Edges, EdgeData, EdgeDatum
from pipeline import Pipeline
from process import ProcessNames, ProcessTypes, ProcessProperties, PortFlags,
                    Ports, PortFrequency, PortAddr, PortAddrs, Connection
from utils import name_thread
from config import Config, empty_config
from stamp import new_stamp, incremented_stamp
from process_cluster import cluster_from_process
```

We could use dynamic introspection in https://github.com/Erotemic/mkinit to autogenerate a base `__init__.py` file that is equivalent to the current behavior. The only reason I haven't done this is because I don't have kwiver built atm.

dstoup:
I would like to consider improvements to the __init__.py generation as a folllow up branch. I think this one should stay focused on consolidating functions which will make the improvement easier anyway.

dstoup:
Jenkins test this please

Purg:
I think we're on the same page that we should prefer a more explicit specification of the ``__init__.py`` file to be checked into the repository.  Our ``__init__.py`` files should be relatively small anyway since they should probably only import important and common symbols from submodules (in this case the pybind11 `.so` libraries) and not literally everything.

dstoup:
Jenkins test this please

dstoup:
Dashboards make me think something happened to the nightly fletch. Don't merge until we can get a clean build.

PR #579, opened 2018-07-27
Adding two new detection processes

hughed2:
Two new processes for Dawkins.

1) merge_detection_sets: Pretty trivial process, just takes two detected_object_sets and outputs a merged detected_object_set
2) create_detection_grid: Takes in an image as input, with a detection width and height in configs. Spits out a grid of detections spanning the image. All detections are the same width and height, so the last row and column may overlap with the previous row and column.


mattdawkins:
The merge process should be it's own process, although it would still be slightly better if the create_grid one was an implementation of the detector base vital class instead of its own process.

hughed2:
detect_features? Sure, I'll move it.

mattdawkins:
chip generator needs a step parameter that isn't the width/height

merger process shouldn't need a custom output port name (I'd just call it detected_object_set as opposed to detected_object_set_out so that it is consistent with everything else that produces detections)

mattdawkins:
I have a commit on viame/master you can cherry-pick for the later part, didn't do the first

arrows/core/create_detection_grid.cxx
```diff
@@ -0,0 +1,180 @@
...
+}
+
+
+vital::detected_object_set_sptr
+create_detection_grid::
+detect( vital::image_container_sptr image_data) const
+{
+  vital::detected_object_set_sptr grid(new vital::detected_object_set());
+  size_t img_width = image_data->width();
+  size_t img_height = image_data->height();
```
linus-sherrill: Could these be const?

arrows/core/create_detection_grid.cxx
```diff
@@ -0,0 +1,180 @@
...
+      vital::bounding_box<double> bbox(i, j, i+d_->width-1, j+d_->height-1);
+      vital::detected_object_sptr det_obj(new vital::detected_object(bbox));
+      grid->add(det_obj);
+    }
+  }
+
+  // Now get the bottom row
+  for (int i = 0; i + d_->width < img_width ; i+=d_->width)
+  {
+    vital::bounding_box<double> bbox(i, img_height - d_->height, i+d_->width-1, img_height-1);
```
linus-sherrill: What if `img_height - d_->height` is zero or negative?

arrows/core/create_detection_grid.cxx
```diff
@@ -0,0 +1,180 @@
...
+    vital::detected_object_sptr det_obj(new vital::detected_object(bbox));
+    grid->add(det_obj);
+  }
+
+  // Now get the bottom column
+  for (int j = 0; j + d_->height < img_height ; j+=d_->height)
+  {
+    vital::bounding_box<double> bbox(img_width - d_->width, j , img_width-1, j+d_->height-1);
+    vital::detected_object_sptr det_obj(new vital::detected_object(bbox));
```
linus-sherrill: What if `img_width - d_->width ` is zero or negative?

arrows/core/create_detection_grid.cxx
```diff
@@ -0,0 +1,180 @@
...
+    grid->add(det_obj);
+  }
+
+  // Our last special case, the bottom right
+  vital::bounding_box<double> bbox(img_width - d_->width, img_height - d_->height , img_width-1, img_height-1);
+  vital::detected_object_sptr det_obj(new vital::detected_object(bbox));
```
linus-sherrill: What if `img_width - d_->width` or ` img_height - d_->height` are zero or negative?

arrows/core/create_detection_grid.cxx
hughed2: There's a check for that above.

arrows/core/create_detection_grid.cxx
hughed2: Again, there's a check for that above.

arrows/core/create_detection_grid.cxx
linus-sherrill: Thanks, Now I see it.

PR #586, opened 2018-08-09
Speed up SRNN_tracking

neal-siekierski:
These changes serve primarily to speed up the tracking done in `arrows/pytorch/SRNN_tracking.py`. The main algorithmic changes towards that end are as follows:
- Use iterators instead of indices for iteration where possible, bypassing a O(N) item look-up cost
- Maintain the set of active tracks, removing the need to check all tracks for their active status
- Only use the active tracks in the construction of the similarity matrix passed to the linear sum assignment problem solver

Note that the second change has been made in a way that breaks the backwards compatibility of the `track` and `track_set` classes in `arrows/pytorch/track.py`. Track activity is now managed at the `track_set`, not the `track` level, and `track_set.get_track` has been renamed to `track_set.__getitem__`, with the old O(N) `track_set.__getitem__` removed.

I have also done a small amount of refactoring and clean-up, along with a couple of bugfixes. The main one that could theoretically change behavior is the one for `track.add_new_track_state`, which previously could lead to overlapping track IDs being created if any track state's confidence was less than the supplied threshold. I've changed it to not skip an ID when a track state is skipped.

I've marked this as WIP mainly because I'm currently having a separate issue involving deactivated tracks and as such haven't been able to validate some of the changed code that only runs on active tracks.

**Edit:** The track activity issue was due to the interval before marking as inactive being less than a step. I've now been able to run it without obvious errors.

**TODO**:
- [X] Validate new code that only runs on active tracks

### DIFF
--- a/arrows/pytorch/SRNN_matching.py
+++ b/arrows/pytorch/SRNN_matching.py
@@ -17,9 +17,6 @@ class TargetRNNDataLoader(data.Dataset):
         self._track_search_threshold = track_search_threshold
         self._rnnType = rnnType
 
-        self._tracks_num = len(track_set)
-        self._track_states_num = len(track_state_list)
-
         self._data_list = self._make_dataset()
 
     def _make_dataset(self):
--- a/arrows/pytorch/SRNN_matching.py
+++ b/arrows/pytorch/SRNN_matching.py
@@ -31,10 +31,9 @@ class TargetRNNDataLoader(data.Dataset):
             else:
                 _rnnType = RnnType.Target_RNN_AIM
 
-            # only process active and un-unpdated track
+            # only process active and un-updated tracks
             if cur_track.active_flag and not cur_track.updated_flag and (_rnnType is self._rnnType):
                 for ts, track_state in enumerate(self._track_state_list):
-
                     # distance between the two bbox's x instead of center
                     dis = abs(cur_track[-1].bbox[0] - track_state.bbox[0])
 
@@ -127,7 +126,7 @@ class SRNN_matching(object):
         self._est_similarity(AIM_V_data_loader, RnnType.Target_RNN_AIM_V)
         self._est_similarity(AIM_data_loader, RnnType.Target_RNN_AIM)
 
-        # obtain the dict: similarity row idx->track_id
+        # obtain the list mapping: similarity row idx->track_id
         track_idx_list = [track.id for track in track_set]
 
         return self._similarity_mat.cpu().numpy(), track_idx_list
--- a/arrows/pytorch/SRNN_matching.py
+++ b/arrows/pytorch/SRNN_matching.py
@@ -25,9 +25,7 @@ class TargetRNNDataLoader(data.Dataset):
     def _make_dataset(self):
         _data_list = []
 
-        for t in range(self._tracks_num):
-            cur_track = self._track_set[t]
-
+        for t, cur_track in enumerate(self._track_set):
             if len(cur_track) < g_config.timeStep:
                 # if the track does not have enough track_state, we will duplicate to time-step and use Target_RNN_AIM_V
                 _rnnType = RnnType.Target_RNN_AIM_V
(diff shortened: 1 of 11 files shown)


neal-siekierski:
@dongshuhao Could you review this? I'd add you as a reviewer directly but I don't have the permissions for that.

arrows/pytorch/SRNN_matching.py
```diff
@@ -123,8 +123,9 @@ def __call__(self, track_set, track_state_list, track_search_threshold):
             TargetRNNDataLoader(track_set, track_state_list, track_search_threshold, RnnType.Target_RNN_AIM),
             batch_size=self._batch_size, shuffle=False, **kwargs)
 
-        self._est_similarity(AIM_V_data_loader, RnnType.Target_RNN_AIM_V)
-        self._est_similarity(AIM_data_loader, RnnType.Target_RNN_AIM)
+        with torch.no_grad():
```
dongshuhao: @neal-siekierski I did not understand your point here. Why do we need to re-enable gradient? Since we only do the forward pass, we do not need any gradient information.

arrows/pytorch/SRNN_matching.py
```diff
@@ -25,9 +25,7 @@ def __init__(self, track_set, track_state_list, track_search_threshold, rnnType)
     def _make_dataset(self):
         _data_list = []
 
-        for t in range(self._tracks_num):
-            cur_track = self._track_set[t]
-
+        for t, cur_track in enumerate(self._track_set):
```
dongshuhao: @neal-siekierski Good point. Since it was my first python coding work, I was still in the mindset of C++.

arrows/pytorch/SRNN_matching.py
neal-siekierski: Gradient computation is enabled by default, and none of the code that I've seen explicitly enables it. Moreover, the emphasis in the documentation is on *locally* disabling gradient computation: https://pytorch.org/docs/stable/autograd.html#locally-disabling-gradient-computation. While I don't think it matters for my current use-case, where indeed only the forward pass is needed, potential future code elsewhere that tries to do training after this method has run would otherwise fail because gradient computation is disabled.

dongshuhao: @neal-siekierski For this tracking code, we only allow forward pass. Not sure we really would have the requirements that training during tracking. At this moment, the training part is a separate project. From my point of view, when we deploy a model, we need to disable gradient in order to have a better performance in terms of forwarding speed.

neal-siekierski: The `with`-block makes sure that gradient calculation is disabled when running inference in `_est_similarity`. The only difference behavior-wise is that, once inference is done, the flag that indicates whether to do gradient calculation is restored to its previous state. So that still does what we're looking for performance-wise, right?

(Also, looking at the docs again, I see that `torch.no_grad()` can also be used as a decorator. Using it on the definition of `_est_similarity` directly rather than in a `with`-block elsewhere would be clearer than what I have currently, I think.)

arrows/pytorch/SRNN_tracking.py
```diff
@@ -71,29 +71,26 @@
...
+   #try:
+       #msg = '[{}] {}'.format(threading.current_thread(), msg)
+   ##    with open('database/Logs/SRNN_Tracking_Log', 'a') as f:
+   ##        f.write(str(msg) + '\n')
+   #except Exception as ex:
+       #with open('database/Logs/SRNN_Tracking_Error_Log', 'a') as f:
+           #f.write('Error durring print! Attempting to report\n')
+           #f.write(repr(ex) + '\n')
+           #f.write(traceback.format_exc() + '\n')
+           #raise
```
mattdawkins: It would be nice if we can somehow parametize these, unfortunately they're required for running the pipeline in the query GUI for now for a semi-unknown reason (the stdout pipe is NULL and writing to it causes an error). Fixing the later would probably be a better solution

mattdawkins:
You should be able to add reviewers now (I added you to the list), default settings should probably be changed to allow anyone to request a specific reviewer

arrows/pytorch/SRNN_tracking.py
neal-siekierski: This particular change is already in `viame/master`. Do we want to revert it? (Also @dongshuhao)

PR #590, opened 2018-08-16
Report all python processes in a pipeline.

linus-sherrill:
When scheduler that does not support python processes gets a pipeline with python processes, it now reports all the python processes.


dstoup:
Clang analysis build didn't fail. I needed to kill it.

PR #600, opened 2018-09-06
Major restructuring of serialization code

linus-sherrill:
- Removing support for multi-data-item serializers
- Changing serializer algo parameters to single element
- New serialized message format which is more introsepctable.
- Messages are assembled by the serializer from single data elements
- Changing [de]serializer process port syntax and semantics.


dstoup:
I approved but just realized that there is a build failure on Windows. It looks like it's rerunning so perhaps that's the fix.

linus-sherrill:
Jenkins, test this please

linus-sherrill:
Jenkins test this please

PR #605, opened 2018-09-07
WIP: Add Resnet descriptor CPU support

neal-siekierski:
In terms of externally functionality, this PR furthers the changes of 1d964fba0561dd405c1bc5b191ccbc6c78510c0a to additionally cover `resnet_descriptors`, adding the ability to specify `gpu_list = None` in a pipeline file to use the CPU instead.

Internally, this PR cleans up some whitespace, changes how the CPU-only mode is specified, and creates a new module `parse_gpu_list` that holds some shared code related to turning the configuration strings into lists and devices.  CPU-only mode is specified by an empty list of GPU indices instead of a separate boolean parameter, which simplifies argument passing somewhat and exposes a similar interface to the pipeline files.

**TODO:**
- [ ] Potentially pick a better name for the `parse_gpu_list` module
- [ ] Verify that the changes to `pytorch_resnet_f_extractor` don't break GPU functionality

### DIFF
--- a/arrows/pytorch/CMakeLists.txt
+++ b/arrows/pytorch/CMakeLists.txt
@@ -72,4 +72,8 @@ if( KWIVER_ENABLE_PYTHON )
   kwiver_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/iou_tracking.py
     kwiver/arrows/pytorch
     iou_tracking)
+
+  kwiver_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/parse_gpu_list.py
+    kwiver/arrows/pytorch
+    parse_gpu_list)
 endif()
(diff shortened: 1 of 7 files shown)


neal-siekierski:
Also, I haven't done anything with `pytorch_alexnet_f_extractor`.  It doesn't appear to be used anywhere in KWIVER or VIAME and is very nearly a duplicate of `pytorch_resnet_f_extractor`. Can it be removed? (Also @mattdawkins)

dongshuhao:
@neal-siekierski pytorch_alexnet_f_extractor and pytorch_resnet_f_extractor were not added by me. Therefore, I am not very sure about the real purpose of these two codes. @mattdawkins maybe a better resource for this PR?

PR #606, opened 2018-09-13
Dev/enhanced documentation

kfieldho:
This PR begins a process to upgrade our "narrative" documentation.  It re-organizes the top level sections a bit to pull key components up in the TOC tree, it changes the default "RTD" theme to use a Kitware color scheme and substantially adds to the Sprokit documentation based on the documentation created for the DIVA project.   It also fixes a few bugs and mis-features along the way.


dstoup:
Looks good to me.

doc/manuals/sprokit/getting-started.rst
```diff
@@ -1,234 +1,485 @@
...
+To test this out we'll  use a simple process called ``numbers`` which simply
+generates numbers on a Sprokit port.  We'll also use a simple Python process
+that prints the number  called ``kw_print_number_process`` the code for which
+can be seen in
 [sprokit/processes/python/kw_print_number_process.py](sprokit/processes/python/kw_print_number_process.py).
+
 As usual, we can lean about this process with the following command::
 
-    $ plugin_explorer --process --type kw_print_number_process -d --config
+	plugin_explorer --process --type kw_print_number_process -d --config
```
linus-sherrill: I think the command whould be
 `plugin_explorer --proc kw_print_number_process -d`
or 
 `plugin_explorer --proc print -d`
The second may get more processes since it is matching on 'print'. The --config flag is no longer supported.

aaron-bray:
Here is what it looks like on RTD
https://kwiver.readthedocs.io/en/dev-enhanced-documentation/

That color scheme looks very nice!
We have an issue with doxygen on RTD with links not showing up
Is the doxygen content available when you build it locally?

Here is the full RTD build log
https://readthedocs.org/api/v2/build/7789595.txt
I did not see anything obvious like breathe running out of memory...

Purg:
Using a fresh python environment with only ``breathe`` and its dependencies installed (``pip install -r doc/manual/requirements.txt`` causes an error when trying to run ``make html`` (``Theme error: sphinx_rtd_theme is no longer a hard dependency since version 1.4.0. Please install it manually.(pip install sphinx_rtd_theme)``).  Either I missed something or "sphinx_rtd_theme" should be added to the ``requirements.txt`` file.

Purg:
Same with ``livereload``.

kfieldho:
@aaron-bray As best that I can tell the Doxygen content does show up in my build, but I might not be looking in the right place.  Can you provide specific examples that you noticed?

@Purg  Which requirements.txt are you looking at?

Purg:
@kfieldho  I was looking at the ``doc/manual/requirements.txt``.  In chatting with @aaron-bray it is not clear if we should add local user deps to that ``requirements.txt`` file, if we should stay with the current semantic dependency listing in the ``README`` file, or if we should have a separate ``requirements.txt`` file for local user installation.

Purg:
I also am not seeing doxygen content being build via a simple ``make html`` invocation (I get the ``doxygenclass: Cannot find file`` warnings as can be seen on the [Vital common structures page here](https://kwiver.readthedocs.io/en/dev-enhanced-documentation/vital/common.html).  I attempted to build the XML content specifically via a ``make xml`` before running ``make html`` from a clean slate but I still get the same error.

doc/manuals/_templates/layout.html
```diff
@@ -0,0 +1,15 @@
+{% extends "!layout.html" %}
+  {% block footer %} {{ super() }}
+
+  <style>
```
Purg: Would a comment here noting that this is for changing the bars to Kitware colors be helpful?

doc/manuals/conf.py
```diff
@@ -66,17 +66,17 @@
 
 # General information about the project.
 project = u'KWIVER'
-copyright = u'2017, Kitware, Inc.'
+copyright = u'2018, Kitware, Inc.'
```
Purg: Do we want the current year or a date range like in our header files?

doc/manuals/sprokit/distributed_pipeline.rst
```diff
@@ -1,50 +1,58 @@
...
+implementations for various serialization protocols.  KWIVER supports JSON
+based serialization and binary serialization. For binary serialization,
+Google’s Protocol Buffers are used. While JSON based serialization makes
+reading and debugging types like detected_object_set easy, binary serialization
+is used to serialize data heavy elements like images.    As with other KWIVER
+arrows, providing new implementations supporting other protocols is
+straightforward.
+
+Constructing a Serialization Algorithm
+''''''''''''''''''''''''''''''''''''''
```
Purg: Sphinx section guideline puts ``^^^`` as the next section tier ([reference](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections))

doc/manuals/sprokit/distributed_pipeline.rst
```diff
@@ -187,7 +205,13 @@ data_serializer algorithms can be displayed with the following command
 
 
 Transport Processes
-===================
+-------------------
+
+KWIVER’s transport strategy is structured as Sprokit end-caps (the needs and
+requirements for the transport implementations are somewhat dependent on
+Sprokit’s stream processing implementation and so don’t lend themselves to
+implementation as KWIVER arrows). The current implementation focuses on a
```
Purg: The sentence that starts with ``The current implementation...`` reads a little odd to me.  Would either of the following sound better?
- ``The current implementation focuses on one-to-many and many-to-one topologies for VITAL data types.``
- ``The current implementation focuses on one-to-many and many-to-one topologies for any one VITAL data type.``

doc/manuals/introduction.rst
```diff
@@ -1,12 +1,47 @@
...
 Introduction
 ============
 
-The KWIVER toolkit is a collection of software tools designed to
-tackle challenging image and video analysis problems and other related
-challenges. Recently started by Kitware’s Computer Vision and
-Scientific Visualization teams, KWIVER is an ongoing effort to
-transition technology developed over multiple years to the open source
-domain to further research, collaboration, and product development.
+KWIVER is a fully featured toolkit for developing Computer Vision *Systems*,
```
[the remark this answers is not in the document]
kfieldho: Shouldn't capitalize "computer vision systems", really.

doc/manuals/_templates/layout.html
kfieldho: Yup

kfieldho:
@aaron-bray @Purg @linus-sherrill Could you guys take a quick look at this branch so that we can get it merged?  I've made the changes that I think are reasonable within the scope of what it's trying to do.

PR #608, opened 2018-09-13
Support a connect_host configure parameter 

kfieldho:
Allows connections to hosts other than local_host


aaron-bray:
Looks good to me

PR #610, opened 2018-09-17
add image width and height into the camera class

MatthieuZins:
- add image width and height accessors in the base camera class
- for perspective cameras, image width and height are stored inside the intrinsics
- for rpc cameras, they are stored as attributes


vital/types/camera_intrinsics.h
```diff
@@ -145,13 +151,26 @@ class VITAL_EXPORT simple_camera_intrinsics :
                            const vector_2d& principal_point,
                            const double aspect_ratio=1.0,
                            const double skew=0.0,
+                           const unsigned int image_width=0,
```
mleotta: I would put these new width and height parameters after `dist_coeffs` to avoid breaking existing uses of this constructor that use distortion coefficients.

vital/tests/rpc_reader.h
```diff
@@ -79,6 +80,8 @@ kwiver::vital::simple_camera_rpc read_rpc( std::string filename )
           case 7:
             image_offset( word_idx ) = value;
             break;
+          case 8:
+            image_dimension( word_idx ) = value;
```
mleotta: we might want a static cast here since `value` is `double` but `image_dimension` contains `int`

vital/tests/rpc_reader.h
```diff
@@ -44,6 +44,7 @@ kwiver::vital::simple_camera_rpc read_rpc( std::string filename )
   kwiver::vital::vector_3d world_offset;
   kwiver::vital::vector_2d image_scale;
   kwiver::vital::vector_2d image_offset;
+  kwiver::vital::vector_2i image_dimension;
```
mleotta: We should initialize this vector to (0,0) because existing rpc files will not have the additional line for width and height.

vital/types/camera_intrinsics.h
```diff
@@ -137,6 +141,8 @@ class VITAL_EXPORT simple_camera_intrinsics :
     principal_point_(0.0, 0.0),
     aspect_ratio_(1.0),
     skew_(0.0),
+    image_width_(1),
```
mleotta: Why are these initialized to 1?  Should it be 0?

vital/types/camera_intrinsics.h
```diff
@@ -145,13 +151,26 @@ class VITAL_EXPORT simple_camera_intrinsics :
...
     aspect_ratio_(aspect_ratio),
     skew_(skew),
+    image_width_(image_width),
+    image_height_(image_height),
     dist_coeffs_(dist_coeffs)
-  {}
+  {
+    if (image_width_ == 0)
+    {
+      image_width_ = static_cast<unsigned int>(std::ceil(principal_point(0)*2));
```
mleotta: While this is usually the best way to guess the image width and height, I'm wondering if we should guess at all.  Maybe we should just leave the width and height at 0 to indicate that it is unknown?  As it is now, there is no way of knowing if this is the true dimensions or just a guess.  I'm open for discussion on this.

vital/types/camera_rpc.h
```diff
@@ -143,13 +147,24 @@ class VITAL_EXPORT simple_camera_rpc :
...
     image_scale_( image_scale ),
     image_offset_( image_offset ),
-    rpc_coeffs_( rpc_coeffs )
+    rpc_coeffs_( rpc_coeffs ),
+    image_width_( image_width ),
+    image_height_( image_height )
   {
+    if (image_width_ == 0)
+    {
+      image_width_ = static_cast<unsigned int>(std::ceil(image_offset_[0]*2));
```
mleotta: See other comment on guessing image dimension in the perspective camera.  If we decide to make changes there, we should do the same here.

vital/types/camera_intrinsics.h
MatthieuZins: Using 0 as the default value, meaning that they are unknown, seems better. I will do the changes for the rpc camera too.

vital/types/camera_intrinsics.h
MatthieuZins: It's done.

vital/tests/rpc_reader.h
MatthieuZins: done

vital/tests/rpc_reader.h
MatthieuZins: done

vital/tests/rpc_reader.h
```diff
@@ -79,6 +80,9 @@ kwiver::vital::simple_camera_rpc read_rpc( std::string filename )
           case 7:
             image_offset( word_idx ) = value;
             break;
+          case 8:
+            image_dimension( word_idx ) = static_cast<int>(value);
```
chetnieter: Sorry to comment late but this is not going to be executed and is not needed unless an additional line is added to the `rpc_data.dat` file in the `data` directory. We probably should add data to `rpc_data.dat` and then expand the ```clone``` test in `test_camera_rpc.cxx` to check the dimensions.

mleotta: +1,  I assume we can make up any dimensions for those rpc files right?  Maybe some files should not have the dimensions and we should test that those read in a zeros.

chetnieter: Yes - we can make up any dimensions. Right now for these tests I only have one set of rpc data but we could add more.

chetnieter: Changes look good to me.

mleotta:
This looks good, the mac and windows errors do not appear to be related to these changes.

PR #611, opened 2018-09-19
Change implementation of texture coordinates on corners

MatthieuZins:
This PR adds a new texture coordinates mode: one coordinate per face corner.

### DIFF
--- a/vital/types/mesh.cxx
+++ b/vital/types/mesh.cxx
@@ -417,16 +417,15 @@ mesh
     }
     else if (other.has_tex_coords() == TEX_COORD_ON_CORNER)
     {
-      tex = std::vector<vector_2d>(2*num_e, vector_2d(0,0));
-    }
-    else if (other.has_tex_coords() == TEX_COORD_ON_FACE_CORNER)
-    {
-      unsigned int nb_face_corners = 0;
-      for (unsigned int f = 0; f < this->num_faces(); ++f)
+      unsigned int nb_corners = faces_->regularity() * num_faces();
+      if (nb_corners == 0)
       {
-          nb_face_corners += this->faces().num_verts(f);
+        for (unsigned int f = 0; f < this->num_faces(); ++f)
+        {
+            nb_corners += this->faces().num_verts(f);
+        }
       }
-      tex = std::vector<vector_2d>(nb_face_corners, vector_2d(0, 0));
+      tex = std::vector<vector_2d>(nb_corners, vector_2d(0, 0));
     }
 
     tex.insert(tex.end(), other.tex_coords().begin(), other.tex_coords().end());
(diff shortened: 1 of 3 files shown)


vital/types/mesh.cxx
```diff
@@ -419,6 +419,15 @@ ::merge(const mesh& other)
     {
       tex = std::vector<vector_2d>(2*num_e, vector_2d(0,0));
     }
+    else if (other.has_tex_coords() == TEX_COORD_ON_FACE_CORNER)
+    {
+      int nb_face_corners = 0;
+      for (int f=0; f < this->num_faces(); ++f)
```
dstoup: Very minor nit. Put spaces between operators and operands. Same comment applies for all cases below.

mleotta:
I'm missing some conceptual understanding here.  How is a "face corner" different from a "corner"?  The difference between vertices and corners is that multiple corners of faces will share a common vertex in the mesh, but you may need separate texture coordinates per corner if your texture space does not have the same mesh topology.

Why can you not use the existing corners (other than there is a missing implementation of finding which faces are valid)?

MatthieuZins:
I agree that they are similar, but the number of "corners" is defined as two times the number of edges. It's ok when the mesh is watertight because each edge is adjacent to 2 faces. But it doesn't correspond to the number of "face corners" when the mesh has boundary edges.

mleotta:
I would argue that defining the number of corners as two time the number of edges is bug then and should be corrected rather than defining a new type of texture coordinates to work around it.  I don't see any valid application that would need to distinguish between "corner" and "face corner".  It may still be helpful over estimate the number of corners to use edge indexing to quickly navigate between face corners, but then just set the invalid corners to (-1, -1) or some other invalid state and not write them out to files.  However, I haven't looked at this code in detail in a while, so I'm not certain.

MatthieuZins:
Ok, then I see two options:

1) Define the number of corners as the number of "valid" half edges (i.e. the number of half edges with a valid adjacent face), or directly as "face corners" similarly to the new TEX_COORD_ON_FACE_CORNER that I added.

2) Keep the over-estimated number of corners and set invalid corners to (-1, -1).


On the one hand, the first solution does not permit indexing with half-edges, but if the mesh has regular faces it is possible to quickly navigate through texture coordinates using the face index. Also, if we disconnect corners from half-edges, we don't need to build the edge graph.

On the other hand, the second solution requires more steps when we input/output texture coordinates because they are generally indexed like this:

          face 0           face 1          face 2   
     vt00 vt01 vt02    vt10 vt11 vt12    vt20 vt21 vt22  ...

We would have to find the corresponding half-edge of each texture coordinate, rearrange the vector and add/remove the invalid values.

Do you have a preference?

mleotta:
Maybe a hybrid.  I think approach number 1 works best when you have a regular mesh (e.g. all triangles or all quads).  That is the common case, we should probably handle the way you have proposed in this branch.  However, for irregular meshes, it would be more efficient to use half edge indexes for texture coordinates.  We could do it both ways depending on whether you have regular mesh or not.  This is just an implementation detail.  Either way the number of texture coordinates written to a file should be the same, which is why I don't think it should have a new enum value.

mleotta:
Actually, I think it would be good to talk this over.  It just occurred to me that there is a case for a third type texture coordinates, so I'd like to discuss this.

MatthieuZins:
Sounds good, tell me when you have some time next week.

MatthieuZins:
Based on our discussion, I replaced TEX_COORD_ON_CORNER with my implementation.

mleotta:
Jenkins test this please.

Cookt2:
Jenkins test this please

Cookt2:
Note that the exact string must be "Jenkins test this please". I'm going to add a few other variations to the trigger; to make it a bit more user-friendly.

PR #612, opened 2018-09-19
Provide required preprocessor flag for new MSVC compiler

aaron-bray:
MSVC updated how the handle extended byte alignments (in other words, Align > alignof(max_align_t)

Our use of Eigen in shared_pts's allocated with make_shared triggers this extended alignment error and we must provide a compiler flag to resolve the build error.


aaron-bray:
I have posted my email on the eigen forums.
I do not know when/if they will do anything to address this
We should really get this in ASAP
They only way to get MSVC 2017 to work with kwiver is to already have a non updated version, or  uninstall the updated version and scour the web for an install for the non-latest version of 2017
Or to just use 2015...

linus-sherrill:
jenkins test this please

PR #613, opened 2018-09-20
Change the way plugin loader handles duplicate factories.

linus-sherrill:
The test for a duplicate factory has been narrowed so that only
factories with the same interface-type, concrete-type, and plugin-name
are considered the same.

The code to check for duplicates has been moved from the add_factory()
method to the default add_fact_hook() method. This allows derived
loaders to change this behaviour if desired.


dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #620, opened 2018-10-02
Replace the target level definitions with a global def.

dstoup:
Several test targets were not included in this definition when applied at the target level. This branch applies the definition more broadly, which is the right thing here since VS will fail if anything is missing it.


aaron-bray:
Looks good!

PR #622, opened 2018-10-02
Remove environment call

aaron-bray:
This call does not do what I was hoping it would do.
To get google tests to compile in MSVC with out setting environment variables prior to starting MSVC we need to modify the way we use Google Tests

More on that later, for now, revert to the same behavior we had previous to this change


dstoup:
Jenkins test this please

dstoup:
Windows is having build issues. If we are refactoring this whole approach, is there any reason not to just revert the original MR?

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Seems magically better. Approved and merging.

PR #625, opened 2018-10-04
disable 4 ocv_SURF tests for mac

Cookt2:
This PR creates an Apple specific test directory for /arrows/ocv/tests. The reason for this is so that the broken tests only get disabled on Apple, instead of all platforms. 

This allows our Mac dashboards to report more useful information while these tests are broken.

### DIFF
--- a/arrows/ocv/CMakeLists.txt
+++ b/arrows/ocv/CMakeLists.txt
@@ -109,6 +109,16 @@ algorithms_create_plugin( kwiver_algo_ocv
   register_algorithms.cxx
   )
 
+# Mac needs a seperate test directory
 if (KWIVER_ENABLE_TESTS)
-  add_subdirectory(tests)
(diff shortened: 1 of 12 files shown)


mleotta:
I'm not sure that we want to disable failing tests.  Otherwise we will forget that they are broken and never fix them.

linus-sherrill:
Do we know how to fix the tests?

mleotta:
We need to fix the tests.  It's likely due to no properly building or linking to opencv_contrib, which is where SURF lives.

Even if we can't fix the tests, this is not the right way to handle disabling failing tests.  It looks like you are copying the working tests into a sub directory.   This means we need to maintain two copies of the same code.  That is bad idea.

dstoup:
Since the test fix is close to landing, I am closing the MR.

PR #626, opened 2018-10-05
Dev/obj track set pybind imports

as6520:
Fix pybind include and simple test case for object track set and state

### DIFF
--- a/vital/bindings/python/vital/tests/test_object_track_set.py
+++ b/vital/bindings/python/vital/tests/test_object_track_set.py
@@ -1,0 +1,82 @@
+"""
+ckwg +31
+Copyright 2016 by Kitware, Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+ * Redistributions of source code must retain the above copyright notice,
+   this list of conditions and the following disclaimer.
+
+ * Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+ * Neither name of Kitware, Inc. nor the names of any contributors may be used
+   to endorse or promote products derived from this software without specific
+   prior written permission.
+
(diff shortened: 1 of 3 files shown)


vital/bindings/python/vital/tests/test_object_track_state.py
```diff
@@ -0,0 +1,101 @@
...
+        nose.tools.assert_equal(ts.time_usec, 14691234578)
+
+
+    def test_detection(self):
+        do = self._create_detected_object()
+        ts = ObjectTrackState(0, 0, do)
+        nose.tools.assert_equal(ts.detection, do)
+
+
+'''
```
Purg: Are these commented tests relevant?  If not we should remove this.

vital/bindings/python/vital/tests/test_object_track_set.py
```diff
@@ -0,0 +1,82 @@
...
+import nose.tools
+import numpy
+
+from six.moves import range
+
+from vital.types import ObjectTrackSet, ObjectTrackState, BoundingBox, \
+        DetectedObjectType, DetectedObject, Track
+
+
+class TestObjectTrackState (unittest.TestCase):
```
Purg: I think this should probably be ``TestObjectTrackSet`` instead of ``State``.

vital/bindings/python/vital/tests/test_object_track_state.py
```diff
@@ -0,0 +1,101 @@
...
+from vital.types import ObjectTrackState, BoundingBox, DetectedObjectType, \
+        DetectedObject
+
+
+class TestObjectTrackState (unittest.TestCase):
+
+    def _create_detected_object(self):
+        bbox = BoundingBox(10, 10, 20, 20)
+        dot  = DetectedObjectType("test", 0.4)
+        do = DetectedObject(bbox, 0.4, dot)
```
Purg: Missing ``return`` from this function, which currently returns ``None`` by default.  The tests here involving detected objects are erroneously passing because they are adding and testing the return of ``None``.

PR #628, opened 2018-10-09
Added local implementation optional data type. (deboostify)

linus-sherrill:
This is a replacement for boost::optional and will stand as a
placeholder until we can use std::optional

### DIFF
--- a/sprokit/src/sprokit/pipeline_util/pipe_declaration_types.h
+++ b/sprokit/src/sprokit/pipeline_util/pipe_declaration_types.h
@@ -36,7 +36,6 @@
 
 #include <sprokit/pipeline/process.h>
 
-#include <boost/optional.hpp>
 #include <boost/variant.hpp>
 
 #include <string>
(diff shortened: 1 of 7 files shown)


sprokit/src/sprokit/pipeline_util/pipe_declaration_types.h
```diff
@@ -36,7 +36,6 @@
 
 #include <sprokit/pipeline/process.h>
 
-#include <boost/optional.hpp>
 #include <boost/variant.hpp>
```
dstoup: Completely unrelated to this branch, but it would be great to tackle boost/variant as well to get boost out of this header.

linus-sherrill: That's on my list to do, but it is all wrapped up in the boost::visitor. Just need to settle on a visitor approach.

dstoup:
Jenkins test this please

dstoup:
rerunning to get a better output from the CentOS dashboard since the early C++ 11 support seems important here. That dashboard shows the sprokit __init__.py error frequently which suggests I should finish moving that code out to vital and get it out of sprokit altogether.

PR #632, opened 2018-10-12
Cleanup timestamp handling in vidl_ffmpeg source

mattdawkins:
Timestamps were previously not being populated to downstream processes

### DIFF
--- a/arrows/vxl/vidl_ffmpeg_video_input.cxx
+++ b/arrows/vxl/vidl_ffmpeg_video_input.cxx
@@ -388,10 +388,6 @@ public:
     meta_ts = 0;
     d_have_abs_frame_time = true;
 
-    while ( ( meta_ts == 0.0 )
-            && d_video_stream.advance()
-            && (( c_time_scan_frame_limit == 0) || frame_count-- ));
-
     return true;
   } // start_at_0_time
 
@@ -407,10 +403,6 @@ public:
     meta_ts = current_time_ms.time_since_epoch().count();
 
     d_have_abs_frame_time = true;
-
-    while ( ( meta_ts == 0.0 )
-            && d_video_stream.advance()
-            && (( c_time_scan_frame_limit == 0) || frame_count-- ));
 
     return true;
   } // current_time
--- a/arrows/vxl/vidl_ffmpeg_video_input.cxx
+++ b/arrows/vxl/vidl_ffmpeg_video_input.cxx
@@ -63,7 +63,7 @@ namespace kwiver {
 namespace arrows {
 namespace vxl {
 
-// -------------------------------------------------------------------------------------------------
+// ---------------------------------------------------------------------------------------
 // Private implementation class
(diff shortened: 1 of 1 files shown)


dstoup:
Jenkins test this please

arrows/vxl/vidl_ffmpeg_video_input.cxx
```diff
@@ -373,8 +380,42 @@ class vidl_ffmpeg_video_input::priv
...
+  bool presentation_time()
+  {
+    int frame_count( c_time_scan_frame_limit );
+
+    meta_ts = 0;
+    d_have_abs_frame_time = true;
+
+    while ( ( meta_ts == 0.0 )
+            && d_video_stream.advance()
+            && (( c_time_scan_frame_limit == 0) || frame_count-- ));
```
KyleFromKitware: What exactly does `presentation_time()` represent? Does it just mean "start at timestamp 0"? What's the purpose of this bit which advances `d_video_stream`?

mattdawkins: Yes, start at 0.

Not sure about the bit that advances d_video_stream, I copied it from other sections thinking it was handling scanning to a certain frame in the video but need to check (maybe it's redundant)

mleotta: You don't want that while loop here.  If you look at the other cases, it's actually a do-while loop.  It's scanning the first `c_time_scan_frame_limit` frames looking for metadata with a timestamp.  If you don't need to scan, then please don't.

I thought "presentation time" means to use the FFMPEG presentation time variable, which doesn't always start at zero?

arrows/vxl/vidl_ffmpeg_video_input.cxx
```diff
@@ -373,8 +380,42 @@ class vidl_ffmpeg_video_input::priv
...
+    using namespace std::chrono;
+    auto current_time = system_clock::now();
+    auto current_time_ms = time_point_cast<microseconds>(current_time);
+
+    meta_ts = current_time_ms.time_since_epoch().count();
+
+    d_have_abs_frame_time = true;
 
-// ------------------------------------------------------------------
+    while ( ( meta_ts == 0.0 )
```
mleotta: This loop is not doing anything except slowing down your program when you open a video.

mattdawkins: I'll remove it

arrows/vxl/vidl_ffmpeg_video_input.cxx
mattdawkins: I thought it was a relative time starting from 0, but I'm not well versed in mpg encoding. If you want I can change "presentation" to "start_at_0" to clean up any ambiguity?

mattdawkins:
Jenkins test this please

PR #636, opened 2018-10-25
Cleanup code

linus-sherrill:
- Improve documentation for pipeline include command
- Use basic throw for pipeline errors. The error location should focus
  on the location in the file being processes not the source code
- Replace string trim functions in the lexer to use the vital version


dstoup:
I'm assuming the Linux test failure is noise. Rerunning to verify

dstoup:
Jenkins test this please

dstoup:
Branch looks fine. Curious about the move back from VITAL_THROW though. I thought that was what we were supposed to use. Otherwise, approved.

PR #646, opened 2018-10-31
Add compute mesh depth map

MatthieuZins:
Main changes:
- new algorithm in core: compute_mesh_depth_map
- new type of camera: camera_affine
- a method to approximate an affine camera from a RPC camera

Currently, the 3 types of cameras are at the same "hierarchy level" which means that I have to check manually the camera type to be able to call their depth method in compute_mesh_depth_map(). It would be better to have a common base class allowing depth computation for camera_perspective and camera_affine. @mleotta Any thoughts on that?


vital/types/camera_rpc.h
```diff
@@ -102,13 +103,17 @@ class VITAL_EXPORT camera_rpc : public camera
   /// Project a 2D image back to a 3D point in space
   virtual vector_3d back_project( const vector_2d& image_pt, double elev ) const;
 
+  /// Get an affine camera from the first order Taylor approximation
```
mleotta: This is a very interesting idea.  I'd be curious to know how well this performs.  Specifically, as you move away from the center point, how quickly does the approximation error increase.  It would be interesting to write a test case that uses one of the really RPC models from the RPC tests, picks an approximation point and 
 1) confirms that the projections are the same at the selected point
 2) computes the error (`(rpc.project(pt) - affine.project(pt)).norm()`) as you move the 3D point away from the selected point.

An alternative algorithm for approximating the affine camera would find the least squares solution over a volume, instead of at a single point.  That is,
 1) pick a bounding volume in world space
 2) sample a uniform 3D grid of points in this volume, say 100 x 100 x 10 (or maybe 10 x 10 x 10 is enough)
 3) project each grid point into the image space with the RPC model
 4) find a least squares estimate of the affine camera that maps the 3D points to the corresponding 2D points

This approach should distribute the error more evenly across that AOI.  However, I think your Taylor approximation solution has value too, so I wouldn't get rid of that.  It would be nice to compare the approaches.

vital/types/camera_affine.h
```diff
@@ -0,0 +1,181 @@
+/*ckwg +29
+ * Copyright 2013-2018 by Kitware, Inc.
```
mleotta: Probably should be `Copyright 2018 by Kitware, SAS`

vital/types/camera_affine.cxx
```diff
@@ -0,0 +1,152 @@
+#include "camera_affine.h"
```
mleotta: Add copyright and license comment block

vital/types/camera_affine.cxx
```diff
@@ -0,0 +1,152 @@
...
+  ray_dir_ = rvec;
+}
+
+
+/// Construct an affine camera from a matrix and the image dimension
+simple_camera_affine::simple_camera_affine(const matrix_3x4d &camera_matrix, unsigned int image_width,
+                                           unsigned int image_height)
+  : view_distance_(0.0), image_width_(image_width), image_height_(image_height)
+{
+  ray_dir_ = {0.0, 0.0, 1.0};   // default ray
```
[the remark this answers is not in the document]
MatthieuZins: Actually, set_matrix() does that. I just needed to initialize ray with a default value because its sign is chosen in set_matrix() by assuming that the new and old ray directions should not differ by more than 90°.
Maybe I should change the default value to [0, 0, -1] so that ray will have a negative z component ? (this is probably more likely for satellite images)

arrows/core/compute_mesh_depth_map.cxx
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/// Helper function to compute barycentric coordinates of p w.r.t the triangle abc
+kwiver::vital::vector_3d barycentric_coordinates(const kwiver::vital::vector_2d& p,
+                                                 const kwiver::vital::vector_2d& a,
+                                                 const kwiver::vital::vector_2d& b,
+                                                 const kwiver::vital::vector_2d& c)
+{
+  kwiver::vital::matrix_3x3d abc;
+  abc << a, b, c, 1, 1, 1;
```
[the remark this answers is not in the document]
MatthieuZins: Actually `a`, `b` and `c` are `vector_2d`

arrows/core/compute_mesh_depth_map.cxx
```diff
@@ -0,0 +1,178 @@
...
+    {
+      for (int u = u_min; u <= u_max; ++u)
+      {
+        vital::vector_2d p(u, v);
+
+        // Handle pixels on triangle boundaries. Assignment rules:
+        //  - if the pixel center is inside the triangle
+        //  - if the pixel is not alread assigned and if the pixel intersects the triangle
+        bool pixel_belongs_to_triangle = is_point_inside_triangle(p, a_uv, b_uv, c_uv);
+        if (! pixel_belongs_to_triangle && std::abs(z_buffer(u, v) - std::numeric_limits<double>::max()) < 1e-5)
```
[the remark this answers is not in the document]
MatthieuZins: I changed to infinity() and even found a function isinf() to do the check

vital/types/camera_rpc.h
MatthieuZins: I did some experiments with the 2 methods. 
Here are the projection errors for a grid of 3d points (50x50 with a constant elevation), interpolated over the entire image. It was obtained for an image of AOI4 (~1.6km x 1.5km)

heatmap

cam_affine

Blue is Taylor approx. and orange is the least-square method.

You're right, the error is better distributed over the image and the global mean error is also slighty lower.

Taylor approx. mean error = 0.0850
Least-square method mean error = 0.0749

mleotta: Nice work.  Are these errors in units of pixels?  If so then either of these seems like a pretty good approximation.

mleotta:
@MatthieuZins, @chetnieter has an immediate need for the depth map rendering code for perspective cameras on another project.  Can you factor our the depth map rendering into another branch and temporarily remove the affine case from it?  That way we can land that first and then rebase the affine camera work on top.

I'd like to talk to you next week about possibly renaming the existing `camera_perspective` to `camera_projective` and using it for both perspective and affine cameras.  We would use different camera intrinsics classes to get either perspective or affine.  This means we would pretend that affine cameras have a location in space.  We are basically doing that here anyway by adding variables for a ray direction and distance.  Using a psuedo camera center is another way to encode the same thing.  The advantage is that there will be a lot more commonality between our affine and perspective cameras and it will allow us to convert between camera types more easily and render affine cameras in our GUI tools.

That said, we might be able to delay the affine approximation for now by looking at other ways to use the RPC directly.  Could we render a height map instead of depth map for RPCs?  I think that might accomplish the same goals, but we should talk about how you are using the resulting depth map.

mleotta:
FYI, here is another very useful function that I wrote years ago:
https://github.com/vxl/vxl/blob/master/contrib/brl/bbas/imesh/algo/imesh_render.h#L30

Not only does it use the triangle scan iterator to avoid testing containment at each pixel, it also removes some of the redundant calculations for interpolating depth at each pixel.  Finally, it renders an integer label into a second integer image if the depth image pixel was updated.  When called over all triangles, passing triangle ids into this function, you end up with an index image.  You can then iterate over the index image and efficiently look up which triangle contributed to each pixel without further depth testing.

vital/types/camera_rpc.h
MatthieuZins: Yes, the errors are in pixels.

MatthieuZins:
Thanks for the links, I will take a look. This can be good optimizations for my code. 
The index image seems nice. Actually, I did it also for Core3d, but it was removed when we started using an occlusion mesh different from the mesh we texture, because the check on triangle id was not possible.

I am available for discussion this week. Just tell me when you want on hangout.

mleotta:
@MatthieuZins we are cleaning up stale pull requests.  I'm going to close this one since the rendering has been done on other branches.  The affine camera work is still relevant, but we should probably handle this another way.  If you still need something from this branch please cherry-pick the relevant parts and push a new PR.

PR #648, opened 2018-11-02
Reintroduce descriptor averaging

neal-siekierski:
This PR reintroduces descriptor averaging (as implemented in 70c689e) and additionally changes it slightly to ensure that no tracks are dropped. For "non-rolling" mode this means treating the first detection specially and always outputting it. An additional change gives sensible behavior to rolling-window mode, generating output for every input detection.

Note that when integrating this with Viame, to avoid changing behavior where not desired, users of `average_track_descriptors` should set `interval` to 1.

Validation status:
- First two commits on top of older VIAME / KWIVER have been run to produce reasonable results
- All the changes compile on top of current `viame/master` (f1db431e1)

### DIFF
--- a/arrows/core/average_track_descriptors.cxx
+++ b/arrows/core/average_track_descriptors.cxx
@@ -121,10 +121,9 @@ average_track_descriptors
 
         if( ots && ots->detection && ots->detection->descriptor() )
         {
-          auto avg_it = d->m_history.find( track->id() );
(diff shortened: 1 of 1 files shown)


linus-sherrill:
Jenkins test this please

mattdawkins:
Merging now, would have liked to see a defintion of what "rolling" is in the code though (in a get/set_configuration block for the configuration)

PR #650, opened 2018-11-05
Factor out render_mesh_depth_map for perspective camera

MatthieuZins:
@mleotta @chetnieter Here is a branch with depth map rendering code for perspective camera.

### DIFF
--- a/arrows/core/render_mesh_depth_map.cxx
+++ b/arrows/core/render_mesh_depth_map.cxx
@@ -110,7 +110,8 @@ vital::image_container_sptr render_mesh_depth_map(vital::mesh_sptr mesh, vital::
 
   if (mesh->faces().regularity() == 3)
   {
-    rasterize_depth(dynamic_cast< const vital::mesh_regular_face_array<3>& >(mesh->faces()), points_2d, points_depth, z_buffer);
+    auto const& triangles = static_cast< const vital::mesh_regular_face_array<3>& >(mesh->faces());
+    rasterize_depth(triangles, points_2d, points_depth, z_buffer);
   }
   else
   {
--- a/arrows/core/render_mesh_depth_map.cxx
+++ b/arrows/core/render_mesh_depth_map.cxx
@@ -108,7 +108,14 @@ vital::image_container_sptr render_mesh_depth_map(vital::mesh_sptr mesh, vital::
     }
   }
 
-  rasterize_depth(mesh, points_2d, points_depth, z_buffer);
+  if (mesh->faces().regularity() == 3)
+  {
+    rasterize_depth(dynamic_cast< const vital::mesh_regular_face_array<3>& >(mesh->faces()), points_2d, points_depth, z_buffer);
+  }
+  else
+  {
+    LOG_ERROR(vital::get_logger("arrows.core.render_mesh_depth_map" ), "The mesh has to be triangular.");
+  }
 
   // Re-compute the real depth
   for (size_t j = 0; j < z_buffer.height(); ++j)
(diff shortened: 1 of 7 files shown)


arrows/core/render_mesh_depth_map.cxx
```diff
@@ -0,0 +1,203 @@
...
+
+    for (int v = v_min; v <= v_max; ++v)
+    {
+      for (int u = u_min; u <= u_max; ++u)
+      {
+        vital::vector_2d p(u, v);
+
+        // Handle pixels on triangle boundaries. Assignment rules:
+        //  - if the pixel center is inside the triangle
+        //  - if the pixel is not alread assigned and if the pixel intersects the triangle
```
mleotta: What is the motivation for this special case of point near the boundary if the pixel is not already assigned?  This could give different rendering results depending on the order in which you process triangles.

also, typo in comment: "alread" --> "already"

arrows/core/render_mesh_depth_map.cxx
```diff
@@ -0,0 +1,203 @@
...
+}
+
+
+void rasterize_depth(vital::mesh_sptr mesh, const std::vector<vital::vector_2d> &image_points,
+                     const std::vector<double> &depth, vital::image_of<double>& image)
+{
+  int width = image.width();
+  int height = image.height();
+  // Write faces on z_buffer with depth test
+  vital::mesh_face_array& faces = dynamic_cast< vital::mesh_face_array& >(mesh->faces());
```
mleotta: We need to check that the mesh faces are triangles.  You could add that check here, but it might be better to just require triangle indices as input to the function.  This function could take `mesh_regular_face_array<3> const& triangles` instead of a mesh.  This forces the caller to pass in the correct thing.

arrows/core/render_mesh_depth_map.cxx
MatthieuZins: The motivation is to include pixels which partially intersect with the triangle. Without this, the pixels where the triangle corners are located might be empty.

The result depends on the order in which I process the triangles in some cases, but it is quite rare (the pixel is intersected by more than one triangle and the center of this pixel is outside each of these triangles).

mleotta:
Please add a brief release note to `doc/release-notes/master.txt` describing this new function.

I think we can merge this soon, after a few minor tweaks, and optimize the implementation in a later branch.

MatthieuZins:
I pushed a new version with the requested changes.

mleotta:
Jenkins test this please

mleotta:
Looks good, but the test is failing on the [Mac Dashboard build](https://open.cdash.org/testDetails.php?test=703321374&build=5614609).  The log shows this:
```
unknown file: Failure
C++ exception with description "std::bad_cast" thrown in the test body.
[  FAILED  ] render_mesh_depth_map.perspective_camera (17 ms)
```

MatthieuZins:
The error is quite strange. It comes from a dynamic_cast which fails only on Mac.
I will need to investigate more deeply.

collinsr:
I remember coming across some weird dynamic_cast problems on my Mac.
I'm not sure this is exactly the problem I was having, but it sounds
familiar:

https://stackoverflow.com/questions/49316779/dynamic-cast-dynamic-cast-fails-with-dylib-on-osx-xcode

FWIW, there's an entry in the company status update from November 7, 2016

"ITK

- Started LightObject::SafeDownCast implementation to work around
buggy dynamic_cast on Mac OSX. This is critical for plugin-apps like
Slicer and for the Python wheel packages. (Matt, Mayeul)"

Hope that helps
-Roddy

arrows/core/render_mesh_depth_map.cxx
```diff
@@ -0,0 +1,213 @@
...
+}
+
+namespace kwiver {
+namespace arrows {
+
+
+vital::image_container_sptr render_mesh_depth_map(vital::mesh_sptr mesh, vital::camera_perspective_sptr camera)
+{
+  unsigned int nb_vertices = mesh->num_verts();
+  vital::mesh_vertex_array<3>& vertices = dynamic_cast< vital::mesh_vertex_array<3>& >(mesh->vertices());
```
mleotta: You could try this here to see if it helps with the cast failure on the dashboard
```c++
vital::mesh_vertex_array<3>& vertices = mesh->template vertices<3>();
```

arrows/core/render_mesh_depth_map.cxx
```diff
@@ -0,0 +1,213 @@
...
+  {
+    for (size_t i = 0; i < z_buffer.width(); ++i)
+    {
+      z_buffer(i, j) = std::numeric_limits<double>::infinity();
+    }
+  }
+
+  if (mesh->faces().regularity() == 3)
+  {
+    rasterize_depth(dynamic_cast< const vital::mesh_regular_face_array<3>& >(mesh->faces()), points_2d, points_depth, z_buffer);
```
mleotta: You might consider capturing the cast face array into a variable before passing into the function.
```c++
auto const& triangles = dynamic_cast< vital::mesh_regular_face_array<3>const& >(mesh->faces());
rasterize_depth(triangles, points_2d, points_depth, z_buffer);
```

It might help with the dashboard failures and it's easier to read.

arrows/core/render_mesh_depth_map.cxx
MatthieuZins: The bad cast is not caused by this line, but by the dynamic_cast on faces (line 113)
`dynamic_cast< const vital::mesh_regular_face_array<3>& >(mesh->faces())`

arrows/core/render_mesh_depth_map.cxx
MatthieuZins: I tried but it doesn't solve the problem.

mleotta:
@dstoup This dynamic cast issue may be related to other dynamic cast issues we are seeing on Mac.  I wonder if the `-flat_namespace` linker flag mentioned in that post would help in both of these cases.

MatthieuZins:
Thanks for the link Roddy.
I tried adding the flag `-flat_namespace` but it doesn't solve the problem.

The solution I found is to add `VITAL_EXPORT` in the declaration of the class `mesh_regular_face_array`. This is working on the Mac we have in our office.

MatthieuZins:
Jenkins test this please

mleotta:
Jenkins test this please

mleotta:
And now the Windows build fails :( but I think you are on to something.  However, `VITAL_EXPORT` can't be used on classes that are defined entirely in the header.

Here is the easy fix: Just use `static_cast` instead of `dynamic_cast` and remove the `VITAL_EXPORT`.  In this case, I think it is sufficient to assume that if  `mesh->faces().regularity() == 3` then your type is `vital::mesh_vertex_array<3>`.  Also this make sense since we are not handling the cases of a failed dynamic cast anyway.

If we wanted to keep the dynamic cast we would probably need to explicitly instantiate  `vital::mesh_vertex_array<3>` in `mesh.cxx` and then export it.  Or we would need to look at what `VITAL_EXPORT` evaluates to on Mac and do something more customized to use that attribute on Mac without using the declspec stuff on Windows.

MatthieuZins:
I changed to `static_cast`. It should be good.

mleotta:
Jenkins test this please

PR #652, opened 2018-11-06
Convert merge detections process to N way merge.

linus-sherrill:
Add support to merger to support dynamic input ports so this
process can merge an indeterminate number of inputs.


dstoup:
looks reasonable to me.

PR #657, opened 2018-11-07
Added image throttle process.

kfieldho:
Cherry picked and cleaned up from squadx/master after DARPA public release permission

### DIFF
--- a/sprokit/processes/core/throttle_image_stream_process.cxx
+++ b/sprokit/processes/core/throttle_image_stream_process.cxx
@@ -1,0 +1,174 @@
+/*ckwg +29
+ * Copyright 2016 by Kitware, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  * Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ *  * Neither name of Kitware, Inc. nor the names of any contributors may be used
+ *    to endorse or promote products derived from this software without specific
+ *    prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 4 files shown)


sprokit/processes/core/throttle_image_stream_process.cxx
```diff
@@ -0,0 +1,174 @@
...
+// ------------------------------------------------------------------
+void
+throttle_image_stream_process::
+_step()
+{
+  kwiver::vital::timestamp frame_time = grab_input_using_trait( timestamp );
+
+  if( !d->m_initialized )
+  {
+    vital::image_container_sptr input = grab_from_port_using_trait( image );
```
linus-sherrill: This line could be pulled up to go with the grab ror the timestamp, since it is done in all three cases below.

sprokit/processes/core/throttle_image_stream_process.cxx
```diff
@@ -0,0 +1,174 @@
...
+    d->m_next_time = frame_time.get_time_seconds() + d->m_tstep;
+    d->m_initialized = true;
+    LOG_TRACE( logger(), "Initializing frame time: " <<
+               frame_time.get_time_seconds() << " seconds");
+    push_to_port_using_trait( image, input );
+    push_to_port_using_trait( timestamp, frame_time );
+    return;
```
linus-sherrill: Return could be done at the bottom since there are no unhandled cases.

sprokit/processes/core/throttle_image_stream_process.cxx
```diff
@@ -0,0 +1,174 @@
...
+               frame_time.get_time_seconds() << " to port");
+    vital::image_container_sptr input = grab_from_port_using_trait( image );
+    push_to_port_using_trait( image, input );
+    push_to_port_using_trait( timestamp, frame_time );
+    d->m_next_time += d->m_tstep;
+    return;
+  }
+  else
+  {
+    LOG_TRACE( logger(), "Received image with frame time " +
```
linus-sherrill: This could use stream operators to construct the log message. Such as 

    << frame_time.get_time_seconds() 
rather than 

    std::to_string(frame_time.get_time_seconds())

mattdawkins:
I'm against this branch landing long-term, not because of any issues with the code, but because:

https://github.com/Kitware/kwiver/blob/viame/downsampler/sprokit/processes/core/downsample_process.cxx

should deprecate it shortly. The above includes this functionality and more (can operate off of frames or times, can optionally replace frame IDs in the process as linus points out might be a good option, works off of arbitrary types instead of just images) and is staged for integration right after the 3 viame-related branches that are up now land.

One option might be to land this without too much effort (so there's a history of it) then delete it as a part of the viame downsampler integration.

sprokit/processes/core/throttle_image_stream_process.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include "kwiver_processes_export.h"
+
+#include <vital/config/config_block.h>
+
+namespace kwiver {
+
+// ----------------------------------------------------------------
+/**
+ * @brief Image object detector process.
```
Purg: Inaccurate description.

linus-sherrill:
Discontinued in favor of a better implementation

PR #660, opened 2018-11-09
Remove scoring directory.

linus-sherrill:
This scoring support is not used.


dstoup:
Is this scoring code something replace by TrackOracle or just not needed?

Branch looks good, pending dashboards.

PR #661, opened 2018-11-09
Convert sprokit tools to applet format

linus-sherrill:
Convert the three sprokit tools to the kwiver applet architecture. Tools do not now compete in the global command namespace and are discoverable due to the plugin approach.


dstoup:
I am seeing the same build error on this clang dashboard. I will double check the machine against master just to see if there is an endemic problem. There are 3 more clang builds staged up for this branch, I am going to kill the oldest 2 to free up cycles.

dstoup:
It looks like an issue with this branch on this CentOS 7 machine (titan, which is on the domain). It doesn't seem to happen on the other CentOS 7 build though. I don't yet know if it's a difference of configuration of something about the systems. I will investigate that further.

dstoup:
Jenkins test this please

dstoup:
The CentOS build is failing because of the Qt5 update.

linus-sherrill:
Jenkins test this please

linus-sherrill:
Jenkins, test this please

linus-sherrill:
Jenkins, test this please

dstoup:
Just took a quick look at the failing Windows tests. Those are the ones that previously failed on the others. They usually run and pass.

dstoup:
It looks like we're finally getting somewhere!

As long as others are fine with the changes (which I think we've already covered), I would say merge!

PR #665, opened 2018-11-13
Porting kwiver applets to use new command option parsing package.

linus-sherrill:
Use header only cxxopts  package for parsing command line parameters

### DIFF
--- a/arrows/core/applets/render_mesh.cxx
+++ b/arrows/core/applets/render_mesh.cxx
@@ -146,7 +146,7 @@ add_command_options()
     ( "output-image", "Output image file name", cxxopts::value<std::string>() )
     ;
 
-    m_cmd_options->parse_positional({"mesh-file", "camera-file", "output-image", "extra"});
+    m_cmd_options->parse_positional({"mesh-file", "camera-file", "output-image"});
 }
 
 
--- a/arrows/core/applets/render_mesh.cxx
+++ b/arrows/core/applets/render_mesh.cxx
@@ -52,8 +52,6 @@
 #include <arrows/core/mesh_operations.h>
 #include <arrows/core/render_mesh_depth_map.h>
 
-#include <kwiversys/CommandLineArguments.hxx>
-
 namespace kwiver {
 namespace arrows {
 namespace core {
@@ -61,13 +59,9 @@ namespace core {
 namespace {
 
 // Global options
-bool        opt_help( false );
-std::string opt_config;         // config file name
 std::string opt_out_config;     // output config file name
-int opt_width = 1920;           // image width
-int opt_height = 1080;          // image height
-bool opt_height_map = false;    // render a height map instead of a depth map
-bool opt_byte_img = false;      // render as a byte image with stretched range
+int opt_width;
+int opt_height;
 
 // ------------------------------------------------------------------
 static kwiver::vital::config_block_sptr default_config()
@@ -84,11 +78,7 @@ static kwiver::vital::config_block_sptr default_config()
   return config;
 }
 
-
-
-}
-
-typedef kwiversys::CommandLineArguments argT;
+} // end namespace
 
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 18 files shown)


linus-sherrill:
Jenkins test this please

dstoup:
I don't yet understand the regex error coming from the clang dashboards, but I have seen the ffmpeg test error from the CentOS build a lot lately. I was looking into a fix yesterday without much success (or active dedication actually)

linus-sherrill:
Jenkins test this please

tools/config_explorer.cxx
```diff
@@ -34,81 +34,61 @@
...
+                                ) );
+
+ m_cmd_options->positional_help( "\n  config-file  - name of configuration file." );
+
+  m_cmd_options->add_options()
+    ( "h,help", "Display applet usage" )
+    ( "ds", "Display detailed application search path" )
+    ( "dc", "Display detailed config contents" )
+    ( "I,path", "Add directory to config search path", cxxopts::value<std::vector<std::string>>() )
+    ( "a,application", "Application name", cxxopts::value<std::string>() )
```
mleotta: Maybe this is outside the scope of this branch, but should `application` default to "kwiver" and `version` default to the version string in `version.h`?

I'm planning to port over some default config files from MAP-Tk soon.  These will get installed to something like `/usr/share/kwiver/[version]/config/` and the `read_config_file` variant that takes these values is designed to add that to the config search path.  This provides a place to store and look for default configuration for applets and algorithms without having to hard code it.

arrows/core/applets/render_mesh.cxx
```diff
@@ -129,78 +119,84 @@ stretch_to_byte_image(kwiver::vital::image_of<T> const& in_image)
...
+    ( "height-map",    "Render a height map instead of a depth map" )
+    ( "byte",          "Render as a byte image with scaled range" )
+
+    // positional parameters
+    ( "mesh-file",    "Mesh file name", cxxopts::value<std::string>() )
+    ( "camera-file",  "Camera file name", cxxopts::value<std::string>() )
+    ( "output-image", "Output image file name", cxxopts::value<std::string>() )
     ;
+
+    m_cmd_options->parse_positional({"mesh-file", "camera-file", "output-image", "extra"});
```
mleotta: What is "extra" here?  I see all the tools have it now, but I don't see where it is used.

tools/kwiver_applet.h
```diff
@@ -33,6 +33,8 @@
 
 #include <tools/kwiver_tools_applet_export.h>
```
mleotta: Maybe this is for a later branch, but we should move all framework code into `vital` or `vital/applet`.  The `tools` directory should only contain cxx files that get built into executables.   We should never have `tools` on the include path.

tools/config_explorer.cxx
linus-sherrill: I think that the default app and version should be with the actual applet rather than with the config explorer. Also the applet name could be used as the application name. Could add a method to the base applet to ``read_applet_config( ... ) `` that could supply the app name and version string.

tools/kwiver_applet.h
linus-sherrill: This can be resolved in a later PR.

linus-sherrill:
Jenkins test this please

arrows/core/applets/render_mesh.cxx
linus-sherrill: It would be a place to look for extra positional parameters beyond what precedes it in the list. It can be removed.

linus-sherrill:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
We are getting much closer now. At least we're done with cryptic build errors and Windows failures. Hopefully all of these tests are failing for the same reason.

dstoup:
Jenkins test this please

Cookt2:
Jenkins test this please

linus-sherrill:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #666, opened 2018-11-13
Add logging controls to setup script.

linus-sherrill:
Converted print to log call to better control output during
plugin loading process.


sprokit/src/bindings/python/sprokit_logging.py
```diff
@@ -107,6 +107,9 @@ def getLogger(name):
         >>> from sprokit import sprokit_logging
         >>> logger = sprokit_logging.getLogger(__name__)
         >>> logger.info('Hello World')
```
dstoup: Not part of this branch, but can you fix tabs while in here?

PR #667, opened 2018-11-14
Update depth map rendering

MatthieuZins:
@mleotta Please have a look.

Main changes:
- add a triangle scan iterator type
- update render_mesh_depth_map to use it
- add render_mesh_height_map
- update tests

### DIFF
--- a/vital/types/triangle_scan_iterator.h
+++ b/vital/types/triangle_scan_iterator.h
@@ -1,0 +1,76 @@
+/*ckwg +29
+ * Copyright 2018 by Kitware, SAS.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  * Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ *  * Neither name of Kitware, Inc. nor the names of any contributors may be used
+ *    to endorse or promote products derived from this software without specific
+ *    prior written permission.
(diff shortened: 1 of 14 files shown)


dstoup:
Jenkins test this please

dstoup:
There are several new warnings from this branch, see the dashboard run here:
https://open.cdash.org/viewBuildError.php?type=1&onlydeltap&buildid=5628783

It's the warnings block with:
warning: comparison between signed and unsigned integer expressions

MatthieuZins:
This should be fixed.

vital/types/triangle_scan_iterator.h
```diff
@@ -0,0 +1,76 @@
...
+class VITAL_EXPORT triangle_scan_iterator
+{
+public:
+  triangle_scan_iterator(const vector_2d& pt1, const vector_2d& pt2, const vector_2d& pt3) :
+    a(pt1), b(pt2), c(pt3)
+  {
+    reset();
+  }
+
+  void reset();
```
mleotta: This is not at all required for this branch, but it would be nice to make this act more like a real iterator in the future.  I've never been a fan of the VXL API for this class.  It would be nice if you could write:
```c++
for ( auto row : triangle_scan(pt1, pt2, pt3) )
{
  int y = row.y;
  int x_start = std::max(0, row.begin());
  int x_end = std::min(img_width, row.end());
  for ( int x = x_start; x < x_end; ++x )
  {
    // do stuff with x, y
  }
}
```
or optionally do the image range clipping in the iterator itself
```c++
for ( auto row : triangle_scan(pt1, pt2, pt3, img_bounds) )
{
  int y = row.y;
  for ( x : row )
  {
    // do stuff with x, y
  }
}
```
... but let's save that idea for a rainy day

MatthieuZins:
I updated the code with the requested changes.

MatthieuZins:
I made the changes.

mleotta:
Jenkins test this please

PR #668, opened 2018-11-14
Add mesh parameterization

MatthieuZins:
This adds a mesh parameterization algo interface in vital and an implementation in core.
@mleotta Please have a look.


arrows/core/parameterize_mesh.cxx
```diff
@@ -0,0 +1,294 @@
...
+}
+
+
+// Get this algorithm's \link vital::config_block configuration block \endlink
+vital::config_block_sptr
+parameterize_mesh::get_configuration() const
+{
+  vital::config_block_sptr config =
+      vital::algo::parameterize_mesh::get_configuration();
+  config->set_value("resolution", d_->resolution,
```
mleotta: I don't think that resolution really matters here.  The UV coordinates are still between 0 and 1 and should be unaffected by this parameter.  This parameter only scales the width and height return values.  I suggest removing those return values anyway (see other related comment).

arrows/core/parameterize_mesh.cxx
```diff
@@ -0,0 +1,294 @@
...
+  vector_2d norm(width, height);
+  for (unsigned int i = 0; i < triangles.size(); ++i)
+  {
+    tcoords[i * 3 + 0] = triangles[i].a.cwiseQuotient(norm);
+    tcoords[i * 3 + 1] = triangles[i].b.cwiseQuotient(norm);
+    tcoords[i * 3 + 2] = triangles[i].c.cwiseQuotient(norm);
+  }
+  mesh->set_tex_coords(tcoords);
+
+  return std::make_pair(width, height);
```
mleotta: I don't think you need to return this.  The texture coordinates should be normalized to a square space between 0 and 1 in each axis.  If later you want to figure out what size texture you need (e.g. 512x512, 1024x1024, etc.) to obtain a certain resolution, you can easily pick one or more triangles and compare the ratio of the square root of their areas in both world and texture space to figure out the size needed to achieve a certain resolution.

arrows/core/parameterize_mesh.cxx
MatthieuZins: I am using the resolution to have the triangles in pixel units and to be able to set margins (in pixels) between them.

arrows/core/parameterize_mesh.cxx
```diff
@@ -0,0 +1,294 @@
...
+      vector_2d a(0.0, 0.0);
+      vector_2d b(AB.norm(), 0.0);
+      double proj = AC.dot(AB) / AB.norm();
+      vector_2d c(proj, (AC - proj * AB.normalized()).norm());
+      // scale B and C by the resolution
+      b /= d_->resolution;
+      c /= d_->resolution;
+
+      total_area += 0.5 * (b(0) * c(1));
+      unsigned int w = static_cast<unsigned int>(std::floor(std::max(b(0), c(0)))) + 1;
```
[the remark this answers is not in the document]
MatthieuZins: Yes, you're right. I forgot that AB was the longest edge when I added this line.

arrows/core/parameterize_mesh.cxx
```diff
@@ -0,0 +1,294 @@
...
+    vector_3d pt1 = vertices[faces(f, 0)];
+    vector_3d pt2 = vertices[faces(f, 1)];
+    vector_3d pt3 = vertices[faces(f, 2)];
+
+    // triangle edges
+    vector_3d pt1pt2 = pt2 - pt1;
+    vector_3d pt1pt3 = pt3 - pt1;
+    vector_3d pt2pt3 = pt3 - pt2;
+
+    if (pt1pt2.norm() < 1e-5 || pt1pt3.norm() < 1e-5 || pt2pt3.norm() < 1e-5 || (pt1pt2.cross(pt1pt3)).norm() < 1e-5)
```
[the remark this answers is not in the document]
MatthieuZins: This check was done to ignore very thin triangles which would not appear in the atlas because they are smaller than a pixel at the given resolution. But, yes, to be completely invariant to the scale of the mesh we should remove this.

MatthieuZins:
I renamed the algorithm in `uv_unwrap_mesh` and I tried to remove all references to absolute dimensions. For that, I use a parameter called `spacing` which is a percentage of the texture size and which defines how much free space is let between triangles.

mleotta:
Thanks for the changes.  It looks like the old files with the old naming are still there, can you remove those?  Also, it's recommend to add some brief notes in the `doc/release-notes/master.txt`.  You should probably add one sentence describing the new abstract algorithm in Vital and one or two sentences describing the implementation in Core.  Before updating the release notes you will need to either rebase your branch on master or merge in master because the release notes have recently been reset since a release of KWIVER was made.

MatthieuZins:
I rebased on master. It should be ready to be merged.

arrows/core/uv_unwrap_mesh.cxx
```diff
@@ -0,0 +1,279 @@
...
+
+    // triangle edges
+    vector_3d pt1pt2 = pt2 - pt1;
+    vector_3d pt1pt3 = pt3 - pt1;
+    vector_3d pt2pt3 = pt3 - pt2;
+
+    // find the longest edge and assign it to AB, C is the other point
+    vector_3d AB, AC;
+    int longest_edge;
+    if (pt1pt2.norm() >= pt1pt3.norm() && pt1pt2.norm() >= pt2pt3.norm())
```
mleotta: Optimization suggestion for future work (not to hold up merging this branch):

use `.squaredNorm()` instead of `.norm()` because it avoids the square root and you only need to compare relative values.  Also, you might want to precompute and store these norm values in variables so you don't have to recompute the same value several times.

mleotta:
Jenkins test this please

PR #671, opened 2018-11-15
Fix release 1.3.0

mleotta:
We may need to re-tag the 1.3.0 release with these changes.


dstoup:
The CentOS build failure is intermittent and not related to this PR. Looks good to me.

PR #672, opened 2018-11-15
Fix windows plugin issues

mleotta:
These changes build on the fixes to the 1.3.0 release.  Merge #671 first.  If we are going to revise the 1.3.0 release anyway, maybe we should include these fixes too?


doc/release-notes/release.txt
```diff
@@ -1,97 +1,14 @@
...
- * Fixed a problem where ports that were not supplied data in the
-   input_adapter_process were incorrectly reported.
-
- * Fix and issue in the feature_detect_extract_SURF arrow. The arrow was using a dynamicCast
-   which causes a segmentation fault on the Mac. On the Mac, symbols exported from an image loaded with
-   RTLD_GLOBAL mode are not available to other images unless built with -flat_namespace.
+ * Fixed issues with plugin paths in setup_KWIVER.bat
 
+ * Fixed issue with loading logger plugins on Windows due to a semicolon
+   that gets appended to environment variables
```
Cookt2: Should these bug fixes be included into 1.3.0? I think we want to hold off on including a release.txt. I can add it in after the release is made.

Cookt2: I see, that makes sense. I created PR #674 with these changes.

PR #673, opened 2018-11-15
Change location for logger plugins.

linus-sherrill:
This separates loading of the logger back-ends and the other
plugins.


dstoup:
Sorry, missed Matt's valid comments. Retracting approval ....

PR #675, opened 2018-11-16
Mesh Triangulation

mleotta:
Add functions to subdivide mesh faces into triangles.  Also provide initialization list constructors for mesh types to make it easier to construct a mesh in code (e.g. in tests).


mleotta:
@MatthieuZins @chetnieter this branch is a prerequisite for my branch that adds a tool (kwiver applet) for mesh depth/height rendering.

mleotta:
Good point.  I'll add a comment about the requirement for convex faces.

It would be nice to add a more advanced version of this function in the future that does take the geometry into account to handle non-convex faces, but I don't think we have a specific need for that right now.

PR #677, opened 2018-11-19
improve render mesh

mleotta:
This branch make some improvements to mesh rendering.  Including some code refactoring, optimization, and partial handling of points behind the camera.  For now it just skips all triangle that have a vertex behind the camera, which is better than rendering garbage, but it really needs to clip and partially render those triangles instead.

### DIFF
--- a/arrows/core/render_mesh_depth_map.cxx
+++ b/arrows/core/render_mesh_depth_map.cxx
@@ -1,5 +1,5 @@
 /*ckwg +29
- * Copyright 2018 by Kitware, SAS.
+ * Copyright 2018 by Kitware, SAS., 2018 by Kitware, Inc.
  * All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


arrows/core/render_mesh_depth_map.cxx
```diff
@@ -91,6 +96,7 @@ vital::image_container_sptr render_mesh_depth_map(vital::mesh_sptr mesh, vital::
 }
 
 
+/// This function renders a height map of a triangular mesh
 vital::image_container_sptr render_mesh_height_map(vital::mesh_sptr mesh, vital::camera_sptr camera)
 {
   vital::image_of<double> height_map(camera->image_width(), camera->image_height(), 1);
```
MatthieuZins: You could initialize `height_map` with `transform_image` as you did in `render_mesh_depth_map`.

PR #678, opened 2018-11-26
Fix windows linking issues in new processes

mattdawkins:


dstoup:
Jenkins test this please

dstoup:
While ZMQ is disabled by default, the dashboard run will still make sure nothing else broke. I can run a local test on my Windows machine since I've seen zmq related link errors in the past.

mattdawkins:
This issue happens regardless of the ZMQ enable due to the other item in this folder which doesn't depend on ZMQ

PR #679, opened 2018-11-26
Fix problem with config block merge.

linus-sherrill:
The source location for the entry was not correctly copied to the
destination block when doing a merge.


dstoup:
The 'vital:config_block.merge_config' test seems to be a new failure.

PR #683, opened 2018-11-30
Add functions to generate a texture for perspective camera

MatthieuZins:
This is a first version for generating a texture for a perspective camera. 
The main function is `generate_texture`.
@mleotta Please have a look and tell me what you think about it.
Also, I don't have any test for this code because it would require several data (mesh, images, cameras, ...)

### DIFF
--- a/arrows/core/generate_texture.h
+++ b/arrows/core/generate_texture.h
@@ -57,7 +57,6 @@ namespace kwiver {
 namespace arrows {
 namespace core {
 
-
 template <class T>
 void dilate_atlas(vital::image& atlas, vital::image_of<char>& mask, int nb_iter);
 
@@ -111,40 +110,30 @@ double bilinear_interp_safe(vital::image const& img, double x, double y, int d=0
     return bilinear_interp_safe(vital::image_of<float>(img), x, y, d);
   else if (img.pixel_traits() == vital::image_pixel_traits_of<double>())
     return bilinear_interp_safe(vital::image_of<double>(img), x, y, d);
+  else
+    return 0.0;
 }
 
 
 /// Base functor used to fuse multiple images. It adjusts the contribution of each image.
-struct image_fusion_method
+struct KWIVER_ALGO_CORE_EXPORT image_fusion_method
 {
+  virtual ~image_fusion_method() {}
   virtual void operator ()(std::vector<double>& scores) const = 0;
 };
 
 
 /// This functor sets the highest score to 1 and the other to 0
-struct select_max_score : image_fusion_method
+struct KWIVER_ALGO_CORE_EXPORT select_max_score : image_fusion_method
 {
-  virtual void operator ()(std::vector<double>& scores) const
-  {
-    auto max_it = std::max_element(scores.begin(), scores.end());
-    if (*max_it > 0)
-    {
-      std::fill(scores.begin(), scores.end(), 0.0);
-      *max_it = 1.0;
-    }
-  }
+  void operator ()(std::vector<double>& scores) const;
 };
 
 
 /// This functor normalizes the scores
-struct normalize_scores : image_fusion_method
+struct KWIVER_ALGO_CORE_EXPORT normalize_scores : image_fusion_method
 {
-  virtual void operator ()(std::vector<double>& scores) const
-  {
-    double sum = std::accumulate(scores.begin(), scores.end(), 0.0);
-    if (sum > 0)
-      std::for_each(scores.begin(), scores.end(), [&sum](double& v) { v /= sum; });
-  }
+  void operator ()(std::vector<double>& scores) const;
 };
 
 
--- a/arrows/core/generate_texture.h
+++ b/arrows/core/generate_texture.h
@@ -346,7 +346,7 @@ generate_texture(vital::mesh_sptr mesh, std::vector<vital::camera_perspective_sp
     vital::vector_3d b3 = v3 - v1;
     double area_3d = a3.cross(b3).norm();
 
-    if (!std::isinf(area_2d) && !std::isinf(area_3d))
+    if (area_2d > 0 && area_3d > 0 && !std::isinf(area_2d) && !std::isinf(area_3d))
     {
       scale = static_cast<size_t>(std::ceil(sqrt(area_3d / area_2d) / resolution));
       break;
--- a/arrows/core/generate_texture.h
+++ b/arrows/core/generate_texture.h
@@ -95,7 +95,6 @@ inline double bilinear_interp_safe(vital::image const& img, double x, double y, 
 
 
 /// Base functor used to fuse multiple images. It adjusts the contribution of each image.
-template <class T>
 struct image_fusion_method
 {
   virtual void operator ()(std::vector<double>& scores) const = 0;
@@ -103,8 +102,7 @@ struct image_fusion_method
 
 
 /// This functor sets the highest score to 1 and the other to 0
-template <class T>
-struct select_max_score : image_fusion_method<T>
+struct select_max_score : image_fusion_method
 {
   virtual void operator ()(std::vector<double>& scores) const
   {
@@ -119,8 +117,7 @@ struct select_max_score : image_fusion_method<T>
 
 
 /// This functor normalizes the scores
-template <class T>
-struct normalize_scores : image_fusion_method<T>
+struct normalize_scores : image_fusion_method
 {
   virtual void operator ()(std::vector<double>& scores) const
   {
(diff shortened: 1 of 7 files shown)


arrows/core/render_mesh_depth_map.h
```diff
@@ -216,6 +216,35 @@ void render_triangle(const vital::vector_2d& v1, const vital::vector_2d& v2, con
...
+/**
+ * \param v1 [in] 2D triangle point
+ * \param v2 [in] 2D triangle point
+ * \param v3 [in] 2D triangle point
+ * \param label [in] value to fill the triangle with
+ * \param img [out] image in which the triangle is rendered
+ */
+template<class T>
+void render_triangle(vital::vector_2d const& v1, vital::vector_2d const& v2,
+                     vital::vector_2d const& v3, T const& label, vital::image& img)
```
mleotta: I'd recommend that `img` be `vital::image_of<T>` as it is in the other render functions.  This forces the caller to make sure they have the correct image type and may make the loop faster by avoiding a cast at each pixel.

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,427 @@
...
+
+
+/// This function generates a texture from a set of images and maps it on the mesh
+/**
+ * \param mesh [in/out] the mesh to texture.
+ * \param cameras [in] a list of cameras perspective
+ * \param images [in] a list of images
+ * \param resolution [in] resolution of the texture (mesh unit/pixel)
+ */
+template <class T, int N>
```
mleotta: `N` should not be a template parameter here.  `N` is the number of color channels, which we can either get from the input images or pass in as a variables.  Either way, it doesn't need to be fixed at compile time and will just lead to code bloat.

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,427 @@
...
+        vital::vector_2d pt_img = cameras[i]->project(pt3d);
+        points_2d[i] = pt_img;
+        // border check from the camera i
+        if (pt_img(0) < 0 || pt_img(0) >= images[i].width() || pt_img(1) < 0 || pt_img(1) >= images[i].height())
+        {
+          point_scores[i] = 0;
+          continue;
+        }
+        // visibility test from the camera i
+        double interpolated_depth = bary_coords(0) * depths_pt1[i] + bary_coords(1) * depths_pt2[i] + bary_coords(2) * depths_pt3[i];
```
mleotta: You probably don't need to pre-compute all the vertex depths for all the images and then interpolate them here.  Instead you could just project the interpolated, `pt3d`, with the camera depth function.

MatthieuZins:
I made all the requested changes except  the change about pre-computing all the vertex depths because I am not completely sure that it's better.
For now, I can take some time on an internal project to finish the integration.

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,441 @@
...
+ * \param images [in] list of images
+ * \param depths_pt1 [in] depths of pt1 w.r.t cameras
+ * \param depths_pt2 [in] depths of pt2 w.r.t cameras
+ * \param depths_pt3 [in] depths of pt3 w.r.t cameras
+ * \param depth_maps [in] list of depthmaps
+ * \param texture [out] texture where the triangle is rendered
+ * \param depth_threshold [in] threshold used for depth test
+ */
+template <class T>
+void render_triangle_from_images(vital::vector_2d const& v1, vital::vector_2d const& v2, vital::vector_2d const& v3,
```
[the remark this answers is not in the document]
mleotta: A second way to do this (involves more refactoring) would be to compute these scores over all all triangles first and put the scores into a texture coordinate image of type `double` for each image.  You may also want to cache the mapping from texture coordinates to image coordinates in a 2-channel image, or they could be recomputed later.  Anyway, the advantage here is that you could compute these score/coordinate images independently for each input depth map, then have a function that takes just takes all score images and adjusts contributions, then have a final loop that interpolates and weights the source images and accumulated into the final texture.  Only the last of these functions need to be templated.

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,441 @@
...
+    kwiver::arrows::core::render_triangle<char>(tcoords[f * 3], tcoords[f * 3 + 1],
+                                                tcoords[f * 3 + 2], 1, texture_mask);
+  }
+
+  kwiver::arrows::core::dilate_atlas<T>(texture, texture_mask, 4);
+
+  // Update texture coordinates
+  for (auto& tc : tcoords)
+  {
+    tc[0] += 0.5;   // half-pixel shift
```
mleotta: should we be accounting for this offset in the creation of the texture image instead?  Specifically, in the bilinear interpolation?

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,441 @@
...
+
+  kwiver::vital::mesh_vertex_array<3>& vertices = dynamic_cast< kwiver::vital::mesh_vertex_array<3>& >(mesh->vertices());
+  auto const& triangles = static_cast< kwiver::vital::mesh_regular_face_array<3> const& >(mesh->faces());
+
+  // Unwrap the mesh
+  if (mesh->has_tex_coords() == 0)
+  {
+    uv_unwrap_mesh unwrap;
+    vital::config_block_sptr config = unwrap.get_configuration();
+    unwrap.set_configuration(config);
```
mleotta: this should not be needed unless you are changing the config from the default.

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,441 @@
...
+
+    if (area_2d > 0 && area_3d > 0 && !std::isinf(area_2d) && !std::isinf(area_3d))
+    {
+      scale = static_cast<size_t>(std::ceil(sqrt(area_3d / area_2d) / resolution));
+      break;
+    }
+  }
+  for (auto& tc : tcoords)
+  {
+    tc.y() = 1.0 - tc.y();
```
mleotta: why do we have to flip Y here?

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,441 @@
...
+    tc *= scale;
+  }
+
+  // Render the depth maps of the mesh seen by the different cameras
+  std::vector<vital::image> depth_maps(images.size());
```
mleotta: make this `image_of<double>`

arrows/core/generate_texture.h
```diff
@@ -0,0 +1,441 @@
...
+ * \param depths_pt2 [in] depths of pt2 w.r.t cameras
+ * \param depths_pt3 [in] depths of pt3 w.r.t cameras
+ * \param depth_maps [in] list of depthmaps
+ * \param texture [out] texture where the triangle is rendered
+ * \param depth_threshold [in] threshold used for depth test
+ */
+template <class T>
+void render_triangle_from_images(vital::vector_2d const& v1, vital::vector_2d const& v2, vital::vector_2d const& v3,
+                                 vital::vector_3d const& pt1, vital::vector_3d const& pt2, vital::vector_3d const& pt3,
+                                 std::vector<vital::camera_sptr> const& cameras, std::vector<vital::image> const& images,
```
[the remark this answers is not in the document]
MatthieuZins: From my point of view, we can force the input and output images to have the same type.
I allowed different types mainly for convenience because the satellite images that we used in core3D were `uint16` and the output texture needed to be `uchar` to be directly compatible with any viewer.

`image_of` also forces the different input images to have the same pixel type, which is important.

arrows/core/generate_texture.h
MatthieuZins: We need to flip Y because the texture coordinate (0, 0) is the bottom-left corner whereas the image origin is the top-left corner. Y is flipped back at the end of the function.

arrows/core/generate_texture.h
MatthieuZins: We could add this offset in the creation of the texture, but I think it is not directly related to the bilinear interpolation of the source image.

It is more related to the texture and how opengl interpolates it. For opengl (0, 0) is the bottom-left corner of the bottom-left pixel and (1, 1) is the top-right corner of the top-right pixel.


If we don't want to do this here, I guess we would have to modify the triangle_scan_iterator (i.e which pixels of the texture are we filling when we render a certain triangle in texture coordinates)

MatthieuZins:
I refactored a bit the code based on your propositions. 
Now, I first put the scores in a texture coordinates image and the corresponding image points in another image. Then, a function adjusts these scores and a final loop fills the texture using the source images weighted with the scores.

However, I kept the computation of the scores of all input images in the same loop because I need to check the visibility at the same time. Doing it independently for each input image would mean recomputing several times the mesh points corresponding to a texture pixels (with barycentric coordinates).

MatthieuZins:
I added a new triangle iterator which takes into account the pixels that are partially inside the triangle (even if their center is outside). This enhances the texture and is more in line with what was done in Core3D.

MatthieuZins:
I changed the way the scores and images points are cached because it required too much memory for large meshes with many images. Now, the scores and image points are only cached for one face at a time and the face is directly rendered in the texture.

PR #686, opened 2018-12-05
Update use of Darknet

neal-siekierski:
This PR makes changes to `darknet_detector.cxx` for compatibility with the latest Darknet. (I have a MR for the KWIVER fork here: https://gitlab.kitware.com/kwiver/darknet/merge_requests/6.)

Darknet now makes use of `network *` in more places, has a new `load_network` function wrapping `parse_network_cfg` and `load_weights`, and makes use of a new `detection` type, including removing `get_region_boxes` with `get_region_detections` as the replacement. This PR compensates for all that, though that means that older versions of Darknet are no longer supported.


mattdawkins:
Deprecated in favor of https://github.com/Kitware/kwiver/pull/687

PR #687, opened 2018-12-05
Update YOLO wrappers from v2 to v3 and add new features

mattdawkins:


mattdawkins:
From Neal: This PR makes changes to darknet_detector.cxx for compatibility with the latest Darknet. (I have a MR for the KWIVER fork here: https://gitlab.kitware.com/kwiver/darknet/merge_requests/6.)

Darknet now makes use of network * in more places, has a new load_network function wrapping parse_network_cfg and load_weights, and makes use of a new detection type, including removing get_region_boxes with get_region_detections as the replacement. This PR compensates for all that, though that means that older versions of Darknet are no longer supported.

neal-siekierski:
Regarding the new commits (taken from https://github.com/Kitware/VIAME/pull/71):

> This PR [...] changes the deep learning model training to create a YOLOv3 model by default, replacing YOLOv2.
> 
> [...]
> 
> Using YOLOv2 is still possible, but now requires a line like `detector_trainer:darknet:model_type = yolov2` in the configuration files. I've opted to switch the default from YOLOv2 to YOLOv3 as part of the KWIVER change, but it may be cleaner to split that out instead.

linus-sherrill:
Jenkins test this please

linus-sherrill:
Jenkins test this please

linus-sherrill:
This still has some build errors.

PR #688, opened 2018-12-05
Created new GDAL image container

chetnieter:
Reworked the GDAL arrow so there is a GDAL specific image container that holds and manages the GDAL data set rather copying the data to a vital image right away. This will make it more memory efficient when the sub image accessor is put in place.

### DIFF
--- a/arrows/gdal/image_container.h
+++ b/arrows/gdal/image_container.h
@@ -30,54 +30,61 @@
 
 /**
  * \file
- * \brief GDAL image_io implementation
+ * \brief GDAL image_container inteface
  */
 
-#include "image_io.h"
+#ifndef KWIVER_ARROWS_GDAL_IMAGE_CONTAINER_H_
+#define KWIVER_ARROWS_GDAL_IMAGE_CONTAINER_H_
 
-#include <arrows/gdal/image_container.h>
 
-#include <vital/exceptions/algorithm.h>
-// #include <vital/exceptions/io.h>
-// #include <vital/types/metadata.h>
-// #include <vital/types/metadata_traits.h>
+#include <vital/vital_config.h>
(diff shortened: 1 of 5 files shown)


dstoup:
Jenkins test this please

chetnieter:
@mleotta I figured we should go ahead and merge this since it is a simple change and it will allow for development of sub-image access via GDAL.

arrows/gdal/image_io.cxx
```diff
@@ -35,92 +35,22 @@
...
+// #include <vital/types/metadata.h>
+// #include <vital/types/metadata_traits.h>
 
-#include <gdal_priv.h>
-#include <ogr_spatialref.h>
+// #include <gdal_priv.h>
+// #include <ogr_spatialref.h>
 
-#include <sstream>
+//#include <sstream>
```
linus-sherrill: If all these includes are not used, could they be removed?

arrows/gdal/image_container.h
```diff
@@ -0,0 +1,90 @@
...
+  /// The width of the image in pixels
+  virtual size_t width() const { return gdal_dataset_->GetRasterXSize(); }
+
+  /// The height of the image in pixels
+  virtual size_t height() const { return gdal_dataset_->GetRasterYSize(); }
+
+  /// The depth (or number of channels) of the image
+  virtual size_t depth() const { return gdal_dataset_->GetRasterCount(); }
+
+  /// Get image. Unlike other image container must allocate memory
```
mleotta: I would remove "Unlike other image container must allocate memory".

Some other image containers (maybe all for now) already have the image data in shareable memory, but it is not the intent of the base class for this to be the primary use case.

If you want you can add a comment that this function allocates memory and loads the pixels from disk, but that should not be in the brief (i.e. `///`) line and doesn't need to contrast with other containers.

PR #689, opened 2018-12-11
Fix memory problem in image_container

m-chaturvedi:
When we pass `memory` to [the `image` constructor](https://github.com/Kitware/kwiver/blob/master/vital/types/image.cxx#L242), and redefine `memory` in the `else` branch, should we modify the pointer to the first pixel as well?

### DIFF
--- a/arrows/ocv/image_container.cxx
+++ b/arrows/ocv/image_container.cxx
@@ -116,6 +116,7 @@ image_container
   // counted reference too it.  If it doesn't own its memory, then the
   // vital image won't take ownership either
   image_memory_sptr memory;
+void* first_pixel = NULL;
 #ifndef KWIVER_HAS_OPENCV_VER_3
   if ( !img.refcount )
 #else
@@ -123,15 +124,16 @@ image_container
 #endif
   {
     memory = std::make_shared<mat_image_memory>(img);
+    first_pixel = img.data;
   }
   else
   {
     cv::Mat img_clone = img.clone();
     memory = std::make_shared<mat_image_memory>(img_clone);
+    first_pixel = img_clone.data;
   }
 
-  return image(memory, img.data,
-               img.cols, img.rows, img.channels(),
+  return image(memory, first_pixel, img.cols, img.rows, img.channels(),
                img.channels(), img.step1(), 1,
                ocv_to_vital(img.type()));
 }


m-chaturvedi:
@mattdawkins for feature review, please?

m-chaturvedi:
To clarify further, it seems to me that:

The shared pointer, `memory` and the [`.data`](https://docs.opencv.org/2.4/modules/core/doc/basic_structures.html#mat) which returns the pointer to the memory should correspond to each other, which they didn't before.

m-chaturvedi:
Please take a look, thank you.

mleotta:
This looks right to me, but @as6520 should comment on what he was trying to achieve by making this change.

m-chaturvedi:
FYI, This started showing up as the root cause to a 'double free' problem

as6520:
I had made the change to fix segmentation fault that was arising when we were trying to serialize the image memory associated with the image.

mleotta:
@mattdawkins can you describe more specifically issues you are trying to work around?  There might be a better fix.  The design is supposed be such that if you wrap an opencv image in a vital image and then convert it back, you get the same image back, not a copy.

Keep in mind that OpenCV, VXL, and Vital all support both images in which the memory is owned (and reference counted) as well as images that views into someone else's memory.  It is perfectly acceptable for an image to have a first pixel pointer, but a null image memory block.  If some code segfaults because the memory block was null, the solution is **not** to deep copy the image so that they memory block is never null.

mattdawkins:
I'll need to check the exact details/use case when I have the time (won't be for a week or two) but I saw a similar segfault using kwiver images when the image was created in opencv, wrapped with kwiver, then used elsewhere, but only with certain types of opencv cv::Mat. It's likely my issue was solved by: bf68e65e6fd74dc3e36efc2f74e1712406ee6bf5 , which is not on master, which looks a little similar to Ameya's change, but exactly the same. Ameya does reverting your change but keeping bf68e65e6fd74dc3e36efc2f74e1712406ee6bf5 solve your issue?

mattdawkins:
but not exactly the same*

as6520:
Yeah it would.

mleotta:
@mattdawkins bf68e65 is even worse in that it always deep copies the memory.  I would prefer the fix from @as6520 with the patch from @m-chaturvedi.  However, both of these changes break the original design of KWIVER image passing, which is to convert between image formats without deep copies whenever possible.  I'm not saying that it's always possible to avoid a deep copy, but this doesn't seem like the right place for it.

I suggest someone should write a minimal test case that currently fails to illustrate the problem, then we can decide the appropriate fix.

mattdawkins:
(that's not on master though so whatever happens here is independent of that)

I see, you're doing std::make_shared<mat_image_memory> not std::make_shared<cv::Mat>

mattdawkins:
http://www.kwiver.org/maptk/docs/nightly/master/d0/d3c/mat__image__memory_8cxx_source.html

just a ref counter incr?

mattdawkins:
https://github.com/Kitware/kwiver/commit/bf68e65e6fd74dc3e36efc2f74e1712406ee6bf5 has been the only thing that's solved all problems for me, even this branch still ended up with garbled opencv images sometimes

mleotta:
@mattdawkins you probably have a use case where you need to make a deep copy of your image, such as a video reader that reloads each new frame into the same block of memory.  It's okay to force a deep copy in your algorithm/application when needed.  You can do that on either the OpenCV or Vital side, but in the conversion is not the place for that.  It forces everyone to get a deep copy anytime you convert OpenCV to Vital.  There are certainly many other use cases where you want the shallow copy and it works fine.

mattdawkins:
@mleotta I was just commenting on what worked for me, not what is necessarily the best solution. Did you very that that commit causes a deep copy? I haven't yet.

These issues all arise when the user constructs a KWIVER image from certain opencv cv::Mat. Not every cv::Mat is the same. The user shouldn't have to know when they need to deep copy an image first, or at the very least, it should throw an exception when the conversion can't be completed. Even then I believe this function is still broken because most of the places having issues is taking the cv::Mat output of opencv functions - haven't been able to delve into it in too much detail yet.

mleotta:
@mattdawkins you are correct, your commit (bf68e65) does not introduce a deep copy, but a deep copy has already been merged into master.  I dug into this a little deeper.  It suspect the segfaults you are seeing are likely due to bugs that have been introduced into the `vital::image` class.  The `image` class is supposed to support the case where it has `data_ == NULL`.  That NULL indicates that the image does not own the memory.  It is a view into someone else memory.  In this case that happens when the OpenCV does not own its memory (i.e. has no ref count), which might happen if the `cv::Mat` was a wrapper around memory allocated by libjpeg or ffmpeg or something else.

What has happened is the `vital::image` class has now been modified to assume `data_` is not NULL.  In commit fff5c0f by @as6520 a few months ago, the added `operator==` assumes that data_ is not NULL and compares the data for image equality.  I would argue that the current `operator==` is broken.  We need to define what `==` means on an image.  VXL has the concept of shallow and deep equality.  The `==` only does shallow equality, meaning the image has the same dimensions and points to the exact same memory.  The deep equality test is a separate function that is much more expensive and needs to check that all the pixels have the same values even if the memory is different.  What is implemented now is not quite correct as a deep equality test because it only works if the images have contiguous memory and the same pixel strides.

In short, I suggest we fix `vital::image` and revert the changes to `ocv::image_container` and see if that solves the problem.  I would suggest making `==` a shallow equality operator.  Are there any objections?  How is `==` being used now for images?

By the way, your commit (bf68e65) works because it acts like OpenCV own the memory but when it goes to delete the memory it can't find a refcount so it doesn't delete anything.  That behavior is actually fine, but it hides the fact that the the memory is not owned.  Some algorithms may want to check if the memory is owned to decide if they need to make a deep copy of the data to retain ownership.

linus-sherrill:
This is a good discussion of the semantics and design intent of an image container. I would be helpful if some of this discussion could be added (copied) to the documentation of ``image_container``, possibly in the header so it will be easily noticed. This would help establish the design intent for that class and define how derived classes need to behave.

mleotta:
Also see #718

linus-sherrill:
Jenkins test this please

linus-sherrill:
I'm seeing a lot of OCV tests fail on this branch.

        226 - ocv:image_io/image_type<unsigned char, 3>.type (Failed)
        227 - ocv:image_io/image_type<unsigned char, 4>.type (Failed)
        229 - ocv:image_io/image_type<unsigned short, 3>.type (Failed)
        230 - ocv:image_io/image_type<unsigned short, 4>.type (Failed)
        231 - ocv:image_conversion/unsigned char.ocv_to_vital_single_channel (Failed)
        232 - ocv:image_conversion/unsigned char.ocv_to_vital_multi_channel (Failed)
        233 - ocv:image_conversion/unsigned char.ocv_to_vital_cropped (Failed)
        235 - ocv:image_conversion/unsigned char.vital_to_ocv_multi_channel (Failed)
        237 - ocv:image_conversion/signed char.ocv_to_vital_single_channel (Failed)
        238 - ocv:image_conversion/signed char.ocv_to_vital_multi_channel (Failed)
        239 - ocv:image_conversion/signed char.ocv_to_vital_cropped (Failed)
        241 - ocv:image_conversion/signed char.vital_to_ocv_multi_channel (Failed)
        243 - ocv:image_conversion/unsigned short.ocv_to_vital_single_channel (Failed)
        244 - ocv:image_conversion/unsigned short.ocv_to_vital_multi_channel (Failed)
        245 - ocv:image_conversion/unsigned short.ocv_to_vital_cropped (Failed)
        247 - ocv:image_conversion/unsigned short.vital_to_ocv_multi_channel (Failed)
        249 - ocv:image_conversion/short.ocv_to_vital_single_channel (Failed)
        250 - ocv:image_conversion/short.ocv_to_vital_multi_channel (Failed)
        251 - ocv:image_conversion/short.ocv_to_vital_cropped (Failed)
        253 - ocv:image_conversion/short.vital_to_ocv_multi_channel (Failed)
        255 - ocv:image_conversion/int.ocv_to_vital_single_channel (Failed)
        256 - ocv:image_conversion/int.ocv_to_vital_multi_channel (Failed)
        257 - ocv:image_conversion/int.ocv_to_vital_cropped (Failed)
        259 - ocv:image_conversion/int.vital_to_ocv_multi_channel (Failed)
        261 - ocv:image_conversion/float.ocv_to_vital_single_channel (Failed)
        262 - ocv:image_conversion/float.ocv_to_vital_multi_channel (Failed)
        263 - ocv:image_conversion/float.ocv_to_vital_cropped (Failed)
        265 - ocv:image_conversion/float.vital_to_ocv_multi_channel (Failed)
        267 - ocv:image_conversion/double.ocv_to_vital_single_channel (Failed)
        268 - ocv:image_conversion/double.ocv_to_vital_multi_channel (Failed)
        269 - ocv:image_conversion/double.ocv_to_vital_cropped (Failed)
        271 - ocv:image_conversion/double.vital_to_ocv_multi_channel (Failed)
        273 - ocv:image_bgr_conversion/unsigned char.bgr_to_rgb (Failed)
        274 - ocv:image_bgr_conversion/unsigned char.bgra_to_rgba (Failed)
        275 - ocv:image_bgr_conversion/unsigned short.bgr_to_rgb (Failed)
        276 - ocv:image_bgr_conversion/unsigned short.bgra_to_rgba (Failed)
        277 - ocv:image_bgr_conversion/float.bgr_to_rgb (Failed)
        278 - ocv:image_bgr_conversion/float.bgra_to_rgba (Failed)
        481 - ffmpeg:ffmpeg_video_input.seek_frame (SEGFAULT)
        488 - ffmpeg:ffmpeg_video_input.seek_nth_frame_output (SEGFAULT)

dstoup:
Is this branch still relevant? If so, can someone investigate the failing tests? It might be helpful to bring it up to date with current master first. @m-chaturvedi can you do that first since this branch is in your private fork?

mleotta:
I'm trying to remember the context here, but I don't this this branch is still needed.  I think the discussion generated some alternative solutions that have since landed.  I suggest we close it and let someone re-open it if there are still issues.

linus-sherrill:
Don't think this branch is relevant any more.

PR #690, opened 2018-12-14
Add downsampler process

mattdawkins:


mattdawkins:
It would be nice if this process excepted arbitrary port names as input instead of input_1, etc...   but not necesarry for this to land

linus-sherrill:
I can look at making the port naming dynamic after this one lands.

mattbrown11:
I think "downsample_process" is too vague for what this process does. I have similar process I made from Squad-X that we are looking to merge soon that I call " throttle_image_stream_process."

sprokit/processes/core/downsample_process.cxx
```diff
@@ -0,0 +1,263 @@
...
+
+namespace kwiver
+{
+
+using sprokit::process;
+
+create_config_trait( target_frame_rate, double, "1.0", "Target frame rate" );
+create_config_trait( burst_frame_count, unsigned, "0", "Burst frame count" );
+create_config_trait( burst_frame_break, unsigned, "0", "Burst frame break" );
+create_config_trait( renumber_frames, bool, "false", "Renumber output frames" );
```
mattdawkins: should clarify, renames output frame IDs

mattdawkins:
slightly against throttle_image_stream  because it doesn't have to just work on images; I like frame_rate_limiter slightly more though that still has the same problem just to a lesser degree. I understand your issue with downsampler and conflicts with image resizing however

mattbrown11:
> slightly against throttle_image_stream because it doesn't have to just work on images; I like frame_rate_limiter slightly more though that still has the same problem just to a lesser degree. I understand your issue with downsampler and conflicts with image resizing however

What is the scope of things it can limit/throttle? What do we call the stuff that flows between processes in sprokit, a stream? Maybe **throttle_stream** or **stream_throttle**. I think "frame" is just as bad as "image" in that context. I throttle is exactly the word/connotation here, it implies limiting the rate. Or **rate_limiter** works too.

linus-sherrill:
Sprokit has the notion of processes generating more or less outputs than inputs, but it is at a fixed rate on a per port basis. I am not sure a variable rate will work in all topologies.

I am not thoroughly familiar with how sprokit handles datum numbering to assure that processes wait until the correct input has arrived, but dropping data in a non-constant manner may cause processes to hang.

mattdawkins:
This process is designed to drop data in a non-constant manner but it's fine, in situations where input frame rates are inconsistent (e.g. network stream or certain video types). I think Linus your command has to do more with built-in sprokit features, though they don't cover all cases that this downsampler process does.

I still prefer downsampler_process to any of the proposed alternatives and want to land this branch. Any objections once I deal with the conflicts?

mattdawkins:
worst case is temporal_downsampler_process though it's long

linus-sherrill:
I can merge it this week   
  

  
  

  
  
>

PR #695, opened 2018-12-18
Make Darknet trainer batch size configurable

neal-siekierski:
This PR allows specification of the batch size and number of subdivisions when using `darknet_trainer`, which it does via two new configuration options, `batch_size` and `batch_subdivisions`, and two new placeholders for configuration files, `[-BATCH_SIZE_INSERT-]` and `[-BATCH_SUBDIVISIONS_INSERT-]`. These correspond to Darknet's `[net]` section's `batch` and `subdivisions` options, respectively.

This PR also reduces the amount of `#ifdef`'d code in `darknet_trainer`.

I haven't tested this code, but with the project I'm on winding down, I'm making it available in the hope that it might be useful.

### DIFF
--- a/arrows/darknet/darknet_trainer.cxx
+++ b/arrows/darknet/darknet_trainer.cxx
@@ -306,36 +306,28 @@ train_from_disk(
     //
     // (This code should be re-written at some point, converted to C++)
 #ifdef WIN32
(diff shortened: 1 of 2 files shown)


neal-siekierski:
@mattdawkins: In the event we do want to use this in VIAME, it will require editing the seed model configuration files.

PR #696, opened 2018-12-20
Don't load all plugins during main.

dstoup:
The main is 'run' when the tests are building. When main tries to load_all_plugins, it sometimes fails and always in this test. This branch moves the load_all_plugins call to the location that requires it, which is only executed when the test is run, not at build time.


linus-sherrill:
Looks good to me

dstoup:
I am going to redo this MR and put it into release first. I think it's an issue there as well.

PR #700, opened 2018-12-21
Don't require doxygen unless we've enabled docs

dstoup:
We are currently including doxygen regardless of whether DOCS was enabled. This configuration fails on systems that don't have doxygen installed.


linus-sherrill:
LGTM

Issue #708, opened 2019-01-04
Nested includes in pipes don't work


mattdawkins:
if I include a .pipe, then include another .pipe in that .pipe it results in an error similar to:

2019-01-04 17:15:46.667 DEBUG plugin_loader.cxx(373): Loading plugins from: /home/matt/Dev/viame/build/build/src/kwiver-build/lib/kwiver/modules/applets/kwiver_tools.so
Exception caught: Expecting assignment operator but found "common_default_input" at /home/matt/Dev/viame/build/install/configs/pipelines/common_default_input_with_downsampler.pipe:8, thrown from /home/matt/Dev/viame/src/packages/kwiver/sprokit/src/sprokit/pipeline_util/pipe_parser.cxx:563

PR #709, opened 2019-01-04
Fix include directive bug

linus-sherrill:
When an include directive was handled, the first line in the included file was not processed. The "include" keyword was still in the current token and as passed to the parser. The first line in the include file needs to be processed when an include directive is encountered.

This addresses issue #708


mleotta:
I would prefer that this was fixed without a `goto` statement.  The logic here is already complex enough and adding a `goto` makes it harder to follow and more likely to result in bugs in the future.  The function is already too long and would be easier to follow if broken up using some helper functions.  For example, you might use a "get_next_line" helper function to handle the code between your `get_new_line` label and the `goto` statement.  Then you can call this function recursively instead of a `goto`.

PR #712, opened 2019-01-10
Fix kwiversys INTERFACE_INCLUDE_DIRECTORIES

KyleFromKitware:
vital_vpm's dependency on kwiversys has also been made PUBLIC due to
the inclusion of <kwiversys/DynamicLoader.hxx> in a public header file.


Cookt2:
Jenkins test this please

PR #713, opened 2019-01-10
Add ability to read images from a directory

KyleFromKitware:
With this change, arrows/core/video_input_image_list automatically
detects if the input file is a directory, and if it is, reads all
files in that directory that match a whitelist of file extensions.
(Note that this search is not recursive.)

### DIFF
--- a/arrows/core/video_input_image_list.cxx
+++ b/arrows/core/video_input_image_list.cxx
@@ -40,8 +40,10 @@
 #include <vital/util/data_stream_reader.h>
 #include <vital/util/tokenize.h>
 
+#include <kwiversys/Directory.hxx>
 #include <kwiversys/SystemTools.hxx>
 
+#include <algorithm>
 #include <string>
 #include <vector>
 #include <stdint.h>
@@ -54,15 +56,19 @@ namespace core {
 class video_input_image_list::priv
 {
 public:
-  priv()
-  : m_current_file( m_files.end() )
+  priv( video_input_image_list* parent )
+  : m_parent( parent )
+  , m_current_file( m_files.end() )
   , m_frame_number( 0 )
   , m_image( nullptr )
   , m_have_metadata_map( false )
   {}
 
+  video_input_image_list* m_parent;
+
   // Configuration values
   std::vector< std::string > c_search_path;
+  std::vector< std::string > c_allowed_extensions;
 
   // local state
   std::vector < kwiver::vital::path_t > m_files;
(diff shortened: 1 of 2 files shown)


Cookt2:
Jenkins test this please

arrows/core/video_input_image_list.cxx
```diff
@@ -426,4 +389,133 @@ ::metadata_map()
...
+  std::ifstream ifs( filename.c_str() );
+  if ( ! ifs )
+  {
+    throw kwiver::vital::invalid_file( filename, "Could not open file" );
+  }
+  if ( ! this->m_image_reader )
+  {
+    throw kwiver::vital::algorithm_configuration_exception( m_parent->type_name(),
+          m_parent->impl_name(), "invalid image_reader." );
+  }
```
linus-sherrill: Since this check of the image reader is done twice, once in each reader type, couldn't it be done once in ```set_configuration()``` ?

arrows/core/video_input_image_list.cxx
```diff
@@ -426,4 +389,133 @@ ::metadata_map()
...
+video_input_image_list::priv
+::read_from_file( std::string const& filename )
+{
+  typedef kwiversys::SystemTools ST;
+
+  // open file and read lines
+  std::ifstream ifs( filename.c_str() );
+  if ( ! ifs )
+  {
+    throw kwiver::vital::invalid_file( filename, "Could not open file" );
```
linus-sherrill: The ``VITAL_THROW()`` macro is available to provide the actual source location where the exception is thrown from if you think it us useful. Some times knowing what file threw the exception can aid in debugging. This comment applies to all places where a kwiver::vital exception is thrown.

KyleFromKitware: This section of code was lifted directly above - I didn't write it. That said, do you think it's worth converting all of these to `VITAL_THROW()` as long as we're here?

arrows/core/video_input_image_list.cxx
KyleFromKitware: Done

arrows/core/video_input_image_list.cxx
linus-sherrill: If you are editing the file and have the time, it would make the file more consistent.

linus-sherrill:
Jenkins test this please

KyleFromKitware:
What happened on Windows?

KyleFromKitware:
Is this set to merge then?

PR #716, opened 2019-01-14
Fix image equality operators

mleotta:
The image equality operator was assuming that `data_` is not NULL causing segfaults in some cases.  Furthermore it was implementing an incomplete notion of image equality.  The changes here make the `==` and `!=` operators only check shallow equality (images share the same memory).  The deep equality test function is called `equal_content` and already exists.


linus-sherrill:
Is this ready to merge?

mleotta:
It's been ready for a while.  It may make sense for @as6520 to comment, at least to acknowledge the change, since it undoes some changes he made.  However, I'm confident that we want the change in this PR.

as6520:
Sorry about the delay. LGTM.

PR #720, opened 2019-01-15
AFRL approved MAP-Tk work

mleotta:
This branch is the last 9 months of work on the MAP-Tk TeleSculptor that was recently approved for public release.  All of this work has been code reviewed internally at Kitware and is documented with release notes.  This doesn't need to be reviewed at a detailed line-by-line level, but it might be good to spot check changes in files that are being used more broadly (especially code in vital).  Much of the work is specific to 3D vision applications and will not impact other KWIVER users, but there are some more general changes as well.

PR #721, opened 2019-01-15
more ranges

mleotta:
Here is chunk of the enormous #720 that was somewhat isolated and I rebased/cherry-picked it out the best I could to allow for additional review.


linus-sherrill:
Jenkins test this please

PR #722, opened 2019-01-15
Add image timestamp capability

KyleFromKitware:


mleotta:
I'm not sure why we need a separate method to access the timestamp.  We already have a metadata object that gets attached to the image container when returned, and metadata already has a field for timestamp.  The idea behind the metadata class is that you put all the metadata there and then you don't need to have separate accessor functions for each metadata field.  You can add new metadata fields without needing to change all the algorithm APIs.

KyleFromKitware:
The problem with attaching the metadata to the image is that, unless we change the API<sup>[1]</sup>, the entire image has to be loaded just to get the timestamp. This means the image would be loaded just by calling `seek_frame()`, which I don't think is desirable.

<sup>[1]</sup> That said, we could make a case for replacing `image_io::timestamp()` with `image_io::metadata()`. Does this seem more reasonable?

mleotta:
Yes, I think a function to return the metadata is more reasonable than a function to return the timestamp.  This also mirrors what we have in the video input API.  

You could also do this with the current APIs if you only need this metadata when working with the video interface.  We have an example that does exactly this sort of thing.  See 

https://github.com/Kitware/kwiver/blob/master/arrows/core/video_input_pos.h

which reads metadata out of a separate text file for each image file.  Then we use

https://github.com/Kitware/kwiver/blob/master/arrows/core/video_input_split.h

to combine metadata reading and image reading from two different video sources.  The video_input_split is then configured to pull metadata from video_input_pos and images from video_input_image_list.  You could do the same but write a replacement for video_input_pos.  However, this does not allow you to use the image_io API directly and still get metadata.

KyleFromKitware:
Writing a replacement for `video_input_pos` would work, but it would mean duplicating the image list logic, which already lives in `video_input_image_list`, and which I would like to reuse if possible. Putting it at the `image_io` layer wouldn't be too difficult, and others might find it useful. I'm going to work on that.

PR #724, opened 2019-01-16
Add track merger process

mattdawkins:


mattdawkins:
I would have liked for this to work on both feature and object tracks, by having the input and output ports both be track_sets, as opposed to object_track_sets which I reverted in a commit, but that was messing up the sprokit pipeline logic. ie converting the object track to it's base track_set class over a port connection was failing. Is there a way around this?

sprokit/processes/core/merge_track_sets_process.cxx
```diff
@@ -0,0 +1,236 @@
...
+  if( !output )
+  {
+    output = std::make_shared< vital::object_track_set >();
+  }
+
+  std::map< vital::track_id_t, vital::track_id_t >& mappings = id_remapping[index];
+
+  for( auto track_ptr : input->tracks() )
+  {
+    if( !track_ptr )
```
linus-sherrill: Does this really happen? I would think having a null track pointer in a ``track_set`` is a failure in ``track_set``. Unless there is a reason for ``nullptr``s in the ``track_set``, they should not be allowed. Then users of the set would not have to do this test.

sprokit/processes/core/merge_track_sets_process.cxx
```diff
@@ -0,0 +1,236 @@
...
+::make_config()
+{
+
+}
+
+
+// ============================================================================
+sprokit::process::port_info_t
+merge_track_sets_process
+::_input_port_info(port_t const& port_name)
```
linus-sherrill: Overriding this method will work, but there is a newer API (something like input_port_undefined() ) that does the same thing but is not called when data passes through a port to this process. It just of an efficiency thing.

PR #726, opened 2019-01-16
Add metadata capability to vital::algo::image_io

KyleFromKitware:


KyleFromKitware:
Both `load()` and `load_metadata()` return metadata, but `video_input_image_list` is only using it from `load_metadata()`. It seems like we have duplicated functionality, and I'm not sure how to resolve it.

KyleFromKitware:
What's the purpose of `frame_metadata()` returning a `metadata_vector` rather than just a `metadata_sptr`? And what's the distinction between that and `metadata_map()`?

mleotta:
`frame_metadata()` returns a vector because, for video, there is no guarantee that the metadata is time synchronized with the video frames, but our step sizes are in frames.  So if your metadata (coming from an IMU, for example) is running at 100Hz and your video is 30Hz you might get 3 or 4 metadata measurements between each frame.  The video_input API is not responsible for figuring out how to combine those, so it gives you a vector of all metadata since the last frame.  In many applications multiple metadata packets will have very similar values and it's okay to just pick one (e.g. the first one).

In most (but not all) cases that we've seen, you only get zero or one metadata packet, and we often enforce that there is always at least some metadata by putting the file name into metadata packet if there is nothing else.

The `metadata_map()` function is designed to scan through the whole video and produce a mapping between frame_number and the metadata vector returned for that frame.   Often you want to scan through the video once to collect all metadata and then do something with that metadata in batch without needing to keep seeking around the video.

KyleFromKitware:
The only problem with this current implementation is that `frame_seek()` does not return a timestamp if it's not included with the image, but you can then get the timestamp from `frame_timestamp()` if you call `frame_image()` first. This doesn't seem ideal... `frame_seek()` should return a timestamp regardless of whether `load_metadata()` is present. What I'm thinking we could do is have `frame_seek()` preemptively load the image to get the timestamp *only if* `image_io::HAS_TIME` is set *and* `load_metadata()` returns null. Does this seem reasonable?

mleotta:
I don't see where `frame_seek()` is not returning a timestamp.  It looks like it does...

KyleFromKitware:
What I mean is that it doesn't get the timestamp from the image. So it returns a timestamp *without a time* unless `load_metadata()` is present. I'm thinking that maybe it should also try `load()` to get it from the image.

mleotta:
I think that is okay.  If anything it should log a warning if `HAS_TIME` is set to true but load_metadata returns NULL.  Image readers that advertise `HAS_TIME` should ensure that they implement a way to get it.  You should not expect to get different metadata by loading the image and accessing the metadata than if you just load the metadata directly.

KyleFromKitware:
Alright, I've fixed up the code per our offline discussion.

KyleFromKitware:
> This looks good to me. However, there are some long lines you might consider breaking.

Done.

mleotta:
Jenkins test this please

PR #727, opened 2019-01-17
Dev/fix image serializer

linus-sherrill:
Fix issue where serializer did not handle image containers that did not own the image memory.

### DIFF
--- a/arrows/serialize/json/load_save.cxx
+++ b/arrows/serialize/json/load_save.cxx
@@ -229,16 +229,33 @@ void load( ::cereal::JSONInputArchive& archive, kwiver::vital::detected_object_t
 void save( ::cereal::JSONOutputArchive& archive, const kwiver::vital::image_container_sptr ctr )
 {
   kwiver::vital::image vital_image = ctr->get_image();
+  kwiver::vital::image local_image;
+
+  if ( vital_image.memory() == nullptr || ! vital_image.is_contiguous() )
+  {
+    // Either we do not own the memory or it is not contiguous.  We
+    // need to consolidate the input image into a contiguous memory
+    // block before it can be serialized.
+    local_image.copy_from( vital_image );
+  }
+  else
+  {
+    local_image = vital_image;
+  }
 
   // Compress raw pixel data
   const uLongf size = compressBound( vital_image.size() );
   uLongf out_size(size);
   std::vector<uint8_t> image_data( size );
   Bytef* out_buf = reinterpret_cast< Bytef* >( &image_data[0] );
-  Bytef const* in_buf = reinterpret_cast< Bytef * >(vital_image.memory()->data());
+  Bytef const* in_buf = reinterpret_cast< Bytef * >( local_image.first_pixel() );
+
+  // Since the image is contiguous, we can calculate the size
+  size_t local_size = local_image.width() * local_image.height()
+    * local_image.depth() * local_image.pixel_traits().num_bytes;
 
   int z_rc = compress( out_buf, &out_size, // outputs
-                       in_buf, vital_image.size() ); // inputs
+                       in_buf, local_size ); // inputs
   if (Z_OK != z_rc )
   {
     switch (z_rc)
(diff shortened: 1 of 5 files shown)


arrows/serialize/json/tests/test_serialize.cxx
```diff
@@ -266,18 +266,35 @@ TEST( serialize, image)
   {
     *cp++ = i;
   }
+  {
```
mleotta: Consider adding duplicates of this test for some more complex image memory layouts.  Those are more likely to fail.  Create a view that is a crop of the original image and test that.  At the moment you need to manually use one of the more complex constructors to do this.

```
// 'offset' is some positive amount
// some of {w, h, d} should be smaller than the original size 
image( other.memory(), other.first_pixel() + offset,
             w, h, d, other.w_step(), other.h_step(), other.d_step(), other.pixel_format());
```

(As a side note, we should add helper functions to make it easier to do these crops.  I've only not added them because I don't have a good place to put them.  After restructuring KWIVER a bit I think we should have an image manipulation library in arrows)

arrows/serialize/protobuf/convert_protobuf.cxx
```diff
@@ -283,16 +283,40 @@ void convert_protobuf( const kwiver::protobuf::image&      proto_img,
...
+    in_buf = reinterpret_cast< Bytef * >( local_image.first_pixel() );
+
+    proto_img.set_w_step( static_cast< int64_t > ( local_image.w_step() ) );
+    proto_img.set_h_step( static_cast< int64_t > ( local_image.h_step() ) );
+    proto_img.set_d_step( static_cast< int64_t > ( local_image.d_step() ) );
+
+  }
+  else
+  {
+    in_buf = reinterpret_cast< Bytef * >( vital_image.first_pixel() );
```
mleotta: same comments for cereal case.

arrows/serialize/json/tests/test_serialize.cxx
linus-sherrill: done

arrows/serialize/json/load_save.cxx
```diff
@@ -230,12 +230,37 @@ void save( ::cereal::JSONOutputArchive& archive, const kwiver::vital::image_cont
...
+    in_buf = reinterpret_cast< Bytef * >( local_image.first_pixel() );
+
+    archive( ::cereal::make_nvp( "w_step", local_image.w_step() ),
+             ::cereal::make_nvp( "h_step", local_image.h_step() ),
+             ::cereal::make_nvp( "d_step", local_image.d_step() )
+      );
+  }
+  else
+  {
+    in_buf = reinterpret_cast< Bytef * >( vital_image.first_pixel() );
```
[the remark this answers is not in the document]
linus-sherrill: done

arrows/serialize/json/load_save.cxx
```diff
@@ -230,12 +230,37 @@ void save( ::cereal::JSONOutputArchive& archive, const kwiver::vital::image_cont
...
+
+    archive( ::cereal::make_nvp( "w_step", vital_image.w_step() ),
+             ::cereal::make_nvp( "h_step", vital_image.h_step() ),
+             ::cereal::make_nvp( "d_step", vital_image.d_step() )
+      );
+}
 
   int z_rc = compress( out_buf, &out_size, // outputs
                        in_buf, vital_image.size() ); // inputs
```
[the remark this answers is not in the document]
linus-sherrill: done

arrows/serialize/protobuf/convert_protobuf.cxx
linus-sherrill: done

arrows/serialize/json/load_save.cxx
```diff
@@ -229,16 +229,29 @@ void load( ::cereal::JSONInputArchive& archive, kwiver::vital::detected_object_t
...
   const uLongf size = compressBound( vital_image.size() );
   uLongf out_size(size);
   std::vector<uint8_t> image_data( size );
   Bytef* out_buf = reinterpret_cast< Bytef* >( &image_data[0] );
-  Bytef const* in_buf = reinterpret_cast< Bytef * >(vital_image.memory()->data());
+  Bytef const* in_buf = reinterpret_cast< Bytef * >( local_image.first_pixel() );
 
   int z_rc = compress( out_buf, &out_size, // outputs
-                       in_buf, vital_image.size() ); // inputs
+                       in_buf, local_image.size() ); // inputs
```
mleotta: There is still one case that is not handled correctly by using `size()` here.  `size()` is currently defined as `memory()->size()` if `memory() != nullptr` else `0`.  There is a case where the image memory is still contiguous but does not use the full memory (e.g. I adjust `first_pixel` and `height` to crop off a few rows of pixels from the start).  In this case you will be copying more data than needed and it will extend beyond the allocated memory.  The correct solution is to use `width * height * depth * pixel_format.num_bytes` instead of `size()`.

There is some argument that says this is what `size()` should be returning.  But that would give the wrong impression if someone copied `size()` bytes for a non-contiguous image.  Another option is that `size()` could return the number of bytes spanning the first pixel to the last pixel, even if not all pixels in between are seen.  What do you think?

PR #733, opened 2019-01-22
Fixed bug with the depth step for ffmpeg reader

chetnieter:
Fixed bug where the depth step for the output images for the ffmpeg
implementation of video_input was not set correctly.


mleotta:
This change looks good, but I think it probably applies to the `release` branch since this bug should also be in the last release.  Please rebase on `release` and add a release note in the bug fixes section of `release.txt`

chetnieter:
Closing since this should goto release branch

chetnieter:
@mleotta Do copyrights need to be updated here as well?

mleotta:
It couldn't hurt.  It's probably less important for a minimal change like this, but I see no reason not to.

PR #738, opened 2019-02-04
Added check for missing metadata tag

mleotta:
metadata_map has_item checked that the map contained a given
frame's data but not that the tag that we want is there too.
Added a check for the particular tag.


mleotta:
This fix from @bclipp-kitware looks good to me

PR #739, opened 2019-02-04
WIP: Added more values to the object_track_state class to properly store all…

anthony-barricelli:
… of a kw18 file's values. (tracking_plane_loc, velocity, world_loc_xyz)

### DIFF
--- a/arrows/core/read_object_track_set_kw18.cxx
+++ b/arrows/core/read_object_track_set_kw18.cxx
@@ -229,16 +229,15 @@ read_object_track_set_kw18::priv
     vital::time_usec_t frame_time = atof( col[COL_TIME].c_str() );
     int track_index = atoi( col[COL_ID].c_str() );
 
-
     vital::vector_2d tracking_plane_loc = {atof(col[COL_LOC_X].c_str()),
-    				           										 atof(col[COL_LOC_Y].c_str())};
+    	atof(col[COL_LOC_Y].c_str())};
 
     vital::vector_2d velocity = {atof(col[COL_VEL_X].c_str()),
-    			         							 atof(col[COL_VEL_Y].c_str())};
+    	atof(col[COL_VEL_Y].c_str())};
 
     vital::vector_3d world_loc_xyz = {atof(col[COL_WORLD_X].c_str()),
-    			              							atof(col[COL_WORLD_Y].c_str()),
-    			             	 							atof(col[COL_WORLD_Z].c_str())};
+    	atof(col[COL_WORLD_Y].c_str()),
+    	atof(col[COL_WORLD_Z].c_str())};
 
     vital::bounding_box_d bbox(
       atof( col[COL_MIN_X].c_str() ),
--- a/arrows/core/read_object_track_set_kw18.cxx
+++ b/arrows/core/read_object_track_set_kw18.cxx
@@ -35,6 +35,8 @@
 
 #include "read_object_track_set_kw18.h"
 
+#include <vital/types/vector.h>
+
 #include <vital/util/tokenize.h>
 #include <vital/util/data_stream_reader.h>
 
@@ -175,7 +177,6 @@ read_object_track_set_kw18
 
     set = new_set;
   }
-
   ++d->m_current_idx;
 
   // Return if we are done parsing
@@ -193,6 +194,7 @@ read_object_track_set_kw18::priv
 
   m_tracks_by_frame_id.clear();
   m_all_tracks.clear();
+
 
   while( stream_reader.getline( line ) )
   {
@@ -227,6 +229,17 @@ read_object_track_set_kw18::priv
     vital::time_usec_t frame_time = atof( col[COL_TIME].c_str() );
     int track_index = atoi( col[COL_ID].c_str() );
 
+
+    vital::vector_2d tracking_plane_loc = {atof(col[COL_LOC_X].c_str()),
+    				           										 atof(col[COL_LOC_Y].c_str())};
+
+    vital::vector_2d velocity = {atof(col[COL_VEL_X].c_str()),
+    			         							 atof(col[COL_VEL_Y].c_str())};
+
+    vital::vector_3d world_loc_xyz = {atof(col[COL_WORLD_X].c_str()),
+    			              							atof(col[COL_WORLD_Y].c_str()),
+    			             	 							atof(col[COL_WORLD_Z].c_str())};
+
     vital::bounding_box_d bbox(
       atof( col[COL_MIN_X].c_str() ),
       atof( col[COL_MIN_Y].c_str() ),
@@ -245,9 +258,12 @@ read_object_track_set_kw18::priv
       std::make_shared< vital::detected_object >( bbox, conf );
 
     // Create new object track state
-    vital::track_state_sptr ots =
+    vital::object_track_state_sptr ots =
       std::make_shared< vital::object_track_state >( frame_index, frame_time, det );
 
+    ots->set_tracking_plane_loc( tracking_plane_loc );
+    ots->set_velocity( velocity );
+    ots->set_world_loc_xyz( world_loc_xyz );
     // Assign object track state to track
     vital::track_sptr trk;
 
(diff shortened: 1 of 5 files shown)


vital/types/track.h
```diff
@@ -61,6 +64,7 @@ typedef std::shared_ptr< track > track_sptr;
 typedef std::weak_ptr< track > track_wptr;
 typedef std::shared_ptr< track_state > track_state_sptr;
 
+//typedef Eigen::Matrix< double, 2, 1 > vector_type;
```
linus-sherrill: If this isn't needed, it would be good to remove it

vital/types/track.h
```diff
@@ -41,6 +41,9 @@
 #include <vital/vital_config.h>
 #include <vital/vital_types.h>
 
+#include <vital/types/vector.h>
+//#include <Eigen/Geometry>
```
linus-sherrill: If this isn't needed, it would be good to remove it

arrows/core/read_object_track_set_kw18.cxx
```diff
@@ -227,6 +229,17 @@ ::read_all()
...
 
+
+    vital::vector_2d tracking_plane_loc = {atof(col[COL_LOC_X].c_str()),
+    				           										 atof(col[COL_LOC_Y].c_str())};
+
+    vital::vector_2d velocity = {atof(col[COL_VEL_X].c_str()),
+    			         							 atof(col[COL_VEL_Y].c_str())};
+
+    vital::vector_3d world_loc_xyz = {atof(col[COL_WORLD_X].c_str()),
+    			              							atof(col[COL_WORLD_Y].c_str()),
```
mattdawkins: Looks like weird spacing / line endings here.

anthony-barricelli: Yeah that was interesting, should be fixed now.

arrows/core/write_object_track_set_kw18.cxx
```diff
@@ -37,6 +37,12 @@
 
 #include <time.h>
 
+#include <stdio.h>
+#include <execinfo.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <unistd.h>
+
```
mattdawkins: Why were all of these headers added?

vital/types/object_track_set.h
```diff
@@ -105,6 +112,16 @@ class VITAL_EXPORT object_track_state : public track_state
...
+  
+  ///Vector getters/setters for missing kw18 fields
+  vector_2d tracking_plane_loc() { return tracking_plane_loc_; }
+  vector_2d velocity() { return velocity_; }
+  vector_3d world_loc_xyz() { return world_loc_xyz_; }
+
+  void set_tracking_plane_loc( vector_2d tracking_plane_loc ) { tracking_plane_loc_ = tracking_plane_loc;}
+  void set_velocity( vector_2d velocity ) { velocity_ = velocity;}
+  void set_world_loc_xyz( vector_3d world_loc_xyz ) { world_loc_xyz_ = world_loc_xyz;}
+  ///////////////////////////////////////////////////////
```
mattdawkins: I don't think we use this style of ////////////////////// comments elsewhere in the code, I would remove them

mattdawkins:
I guess another comment about this branch is that not all tracker variants use velocities, world plane locations, and I would possibly consider adding a new object_track_state derived class, but I would need to think through the ramifications of this a little more (it might just create more code headaches by making that change). Thoughts?

vital/types/object_track_set.h
```diff
@@ -113,15 +113,14 @@ class VITAL_EXPORT object_track_state : public track_state
     return time_;
   }
   
-  ///Vector getters/setters for missing kw18 fields
+  ///Vector getters/setters for kw18 fields
```
mattdawkins: I wouldn't call these kw18 fields either, vital types should be abstracted away from file formats, algorithms, etc...

anthony-barricelli:
Would the new track state class simply avoid the possibility of those fields going unused? We could make a new class that extends object_track_state with these fields I suppose. Granted, I am still not entirely familiar with the coding practices employed here, so if there are more reasons as to why we should make a new class, I'd love to learn why.

collinsr:
Do we have an overall track state design concept / goal / document / plan? The current vidtk track state is sort of a "throw everything in there" approach, with e.g. magic numbers (444 for lat/lon not valid)... we certainly don't want to tie the track state to any particular format (least of all kw18).

(We've tried this a couple of different ways in the past:

- VIRAT phase 1's tracks on the API side had an enumeration of all possible attributes with get / set / has_a members. I don't think this survived VIRAT phase 1.

- In parallel, the nascent vidtk tracker came along and (I think) used a base class with the "obvious" attributes with everything else being thrown into a big boost_any property map. (This is the "data_" member of the current vidtk track_state, dunno if anybody still uses it.)

These were both awkward for different reasons.)

If we're starting fresh and clean with kwiver, I think we should consider listing out the track state variants and deciding if we want to have a tightly-coupled class hierarchy or something more loosely coupled. (This may have been a past / present / future goal for a kwiver summit.)

For example, one alternative might be to have abstract classes for position, extent, appearance, etc. I think there's a finite set of these categories of track state information. 

FMV instances of position might be (img_x, img_y; world_x, world_y) while WAMI instances might be (lat, lon) or MGRS.

FMV extent might be a bounding box (image coords) while WAMI extent might just be a radius in meters.

FMV appearance might be an image chip, an AHMI model, etc., while the WAMI appearance might be null.

...Then the overall track state could be loose aggregations of these classes.

I guess my concern is that we don't incrementally code ourselves into a position where either (a) the track state is a monolithic aggregation of all possibilities or (b) we have a twisty little hierarchy of class states with lots of if-then-else conditions in the code to check if we have velocities or homographies or whatever.

kfieldho:
I'll drag in @mattbrown11 who's got a group looking at this to address the notion of an overriding plan.

In the short term, however, a few considerations:

* We currently have KW18 readers and writers which presently which presently do not reflect.  A pipeline consisting only of a reader and a writer looses information.
* We have over a decades's worth of KW18 files that we'll want to process in various ways, while retaining compatibility with heritage tooling

Given that, is there a way to cast this MR in such a way that:
* KW18 support is complete and correct
* Better, more flexible, modern etc. mechanisms for maintaining tracks in KWIVER are not precluded
* Current work can proceed while we consider what "better, more flexible, modern" means

mattbrown11:
I agree with Roddy that we do not want to emulate the kw18 wholesale. One frustration I have with kw18 is that it allows you to save multiple sets of parameters which are often mutually derivable. For example, in the case of WAMI, we usually stabilize the raw imagery to align with some canonical image view (e.g., the first frame). Then, the tracker runs on the set of already-stabilized images with a fixed mapping between image coordinates and latitude/longitude defined by a homography. So, the entire two columns of image coordinates in the kw18 can be derived from the lat/lon coordinates, and vice versa, if you know that homography. Often, one set of values is actually the dominant one and the other is an estimated conversion. E.g., we are measuring the object's position precisely in image coordinates, and then we are estimating the conversion to lat/lon based on some assumptions and hoping we did the stabilization well (e.g., the tops of tall buildings will oscillate over a WAMI orbit in image coordinates but they are actually at fixed lat/lon). Likewise, if you have the GPS receiver output from a vehicle, you can estimate where that vehicle should be in an image, but the lat/lon coordinates are really the more-accurate measurements. Throw in the tracking plane x,y, and you have a third derivable set of quantities. 

I see why we chose to store both in the kw18, it prevents needing a separate file encoding the homography, which can get confusing. Also, the conversion between image coordinates may be a very complicated process that is difficult to reproduce outside of the running software. Consider the case where we have some video feed from UAV and we are tracking people within an environment that we have a full 3-D model for. We are tracking the people in image coordinates, but the particular tracking processing that is running is aware of the 3-D model and ray trace a particular pixel to get the (x,y,z) coordinates in meters and the associated (lat, lon, alt). We may choose to save all three sets of coordinates to the kw18 because it would be very difficult to derive one from the other after the fact. But while still in memory, it probably makes more sense to provide a class function to derive the later two sets of world coordinates on demand.

Likewise, velocity can generally be derived from the other information itself. Generally, we estimate velocity by the looking at the distance and time between track states. This again might be something that is calculated on demand by the track object.

I think resolving these issues and defining what these Vital types should be is going to involve some extended conversations. We might we to convene for a meeting to discuss.

PR #742, opened 2019-02-08
Added default plugin search paths relative to the kwiver executable

mleotta:
When installed, you should be able to run the kwiver tool without
needing to set any environment variables.

Ultimately we should only need to set the path to `lib/kwiver/modules` but the plugin loading does not yet automatically look in `applets` and `processes`.

### DIFF
--- a/tools/kwiver_tool_runner.cxx
+++ b/tools/kwiver_tool_runner.cxx
@@ -1,5 +1,5 @@
 /*ckwg +29
- * Copyright 2018 by Kitware, Inc.
+ * Copyright 2018-2019 by Kitware, Inc.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -35,6 +35,7 @@
 #include <vital/plugin_loader/plugin_manager.h>
 #include <vital/plugin_loader/plugin_factory.h>
 #include <vital/exceptions/base.h>
+#include <vital/util/get_paths.h>
 
 #include <kwiversys/SystemTools.hxx>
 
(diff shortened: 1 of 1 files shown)


mleotta:
The reason to check for this path relative to the executable at run time is to make the tool relocatable.  For example, on macOS this allows you to build an application bundle that can be installed anywhere on the system.  The bundle contains both the executable in bin and the plugins under lib but the absolute path to the bundle root can change.  Likewise for building a windows installer.  You can choose where you want to put the files when you extract the installer.   In my particular case I am packaging the `kwiver` command line tool as a side car tool along side the TeleSculptor application which has both a Windows installer and a Mac bundle.  I'd like users to be able to use `kwiver` without setting the environment.

If the path to `processes` is wrong I can fix that.  As we've discussed before, I would be nice to just set a single path and have the plugin loader know to look in a fixed set of subdirectories for specific plugin types.  However, that is outside the scope of this change.

linus-sherrill:
Then it makes sense to have the plugin loader look relative to the executable. You can look at the other plugin paths as specified in the CMake file or roll that forward to another branch.

mleotta:
@linus-sherrill I'm not quite sure what you are asking me to do.

linus-sherrill:
The patch looks good as it is. My other comment was that if you had the time, you could change the install directory for processes to .../lib/kwiver/modules/processes in the root CMakeLists.txt or that can be done later if you can't get to it.

mleotta:
I think that change should be done as a larger separate branch that reorganizes how modules are loaded.  In the long term I'm not sure we need the `modules` either.  It might be better to just have
```
lib/kwiver/applets
lib/kwiver/algorithms
lib/kwiver/processes
lib/kwiver/loggers
...
```
and then you only set the search path to `lib/kwiver` and the plugin loader knows to which sub directories to look in for specific types of plugins.  Again, that should be a separate branch.

PR #748, opened 2019-02-19
use plugin registrar cuda

linus-sherrill:
Convert plugin registration to use new registrar approach.


mleotta:
looks okay, but this also pulls in the OpenCV change that is already done on another branch.

PR #749, opened 2019-02-19
Convert OCV Arrow to use new algorithm registrar

linus-sherrill:
Convert plugin registration to use new registrar approach.


linus-sherrill:
Jenkins test this please

PR #756, opened 2019-02-20
Fixed a bug in warp_image introduce by addition of OpenMP

mleotta:
When switching to OpenMP, the column pointer was computed in the loop
for each row rather than incremented, but the computed pointer did
not account for the starting row, which is not always zero.


mleotta:
@bclipp-kitware This seems to fix some crashes we've been seeing lately

Issue #759, opened 2019-02-26
Replace KWIVER_ALGO with KWIVER_ARROWS in include guards


mleotta:
Many include guards in arrows use the pattern `KWIVER_ALGO_*` instead of `KWIVER_ARROWS_*`.  Some correctly use `ARROWS`.  The `ALGO` is left over from KWIVER's origins in MAP-Tk.  We should replace all `KWIVER_ALGO_` with `KWIVER_ARROWS_` in include guards.

PR #761, opened 2019-02-28
Updated include guards to be KWIVER_ARROWS_* instead of KWIVER_ALGO_*

anthony-barricelli:
Updated include guards to be KWIVER_ARROWS_* instead of KWIVER_ALGO_* as per Issue #759

### DIFF
--- a/arrows/cuda/integrate_depth_maps.h
+++ b/arrows/cuda/integrate_depth_maps.h
@@ -95,4 +95,4 @@ private:
 }  // end namespace arrows
 }  // end namespace kwiver
 
-#endif // KWIVER_ALGO_ARROWS_INTEGRATE_DEPTH_MAPS_H_
+#endif // KWIVER_ARROWS_CUDA_INTEGRATE_DEPTH_MAPS_H_
--- a/arrows/cuda/integrate_depth_maps.h
+++ b/arrows/cuda/integrate_depth_maps.h
@@ -33,8 +33,8 @@
 * \brief Header file for compute depth
 */
 
-#ifndef KWIVER_ALGO_CUDA_INTEGRATE_DEPTH_MAPS_H_
-#define KWIVER_ALGO_CUDA_INTEGRATE_DEPTH_MAPS_H_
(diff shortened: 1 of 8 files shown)


arrows/cuda/integrate_depth_maps.h
```diff
@@ -95,4 +95,4 @@ class KWIVER_ALGO_CUDA_EXPORT integrate_depth_maps
 }  // end namespace arrows
 }  // end namespace kwiver
 
-#endif // KWIVER_ALGO_CUDA_INTEGRATE_DEPTH_MAPS_H_
+#endif // KWIVER_ALGO_ARROWS_INTEGRATE_DEPTH_MAPS_H_
```
mleotta: This one replaced "CUDA" instead of "ALGO"

mleotta:
These changes look good, but it only covers a subset of the files that have this problem.  Almost all of the arrows have classes that need to be updated in this way.  We could merge this branch as is, but additional changes are need to fully resolve #761

anthony-barricelli:
I see, I wasn't sure if we just wanted to update the include guards or the class names as well.

mleotta:
I think the class names are okay, the included guards the primary focus of this issues, but feel free to point out class names that don't look consistent.

anthony-barricelli:
Hopefully the last commit addresses what you meant.

mleotta:
Ah, I see what you mean now by class names.  I didn't realize that most of those "KWIVER_ALGO_*" instances were in the "EXPORT" macros.  I think we do want to change the naming as you did here, expect I think we also need to change where those macros are created to match.  These are defined in the "*_export.h" header files that are generated by CMake and are based on the name of the libraries, which also needs to be updated (I was going to make a separate issue for that).

So I think the best course of action is to remove your last commit and put it on a new branch with a new PR.  We can merge you previous two commits as is.  On the new branch we also need to change the library names in CMake from "kwiver_algo_*" to "kwiver_arrows_*"

mleotta:
Jenkins test this please

PR #766, opened 2019-03-07
Add utility to compute the MD5 of a file.

collinsr:
This commit adds a utility routine which takes a filename and computes
the md5 of the file.

### DIFF
--- a/vital/util/file_md5.h
+++ b/vital/util/file_md5.h
@@ -1,0 +1,57 @@
+/*ckwg +29
+ * Copyright 2019 by Kitware, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  * Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ *  * Neither name of Kitware, Inc. nor the names of any contributors may be used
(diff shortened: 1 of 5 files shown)


vital/util/file_md5.h
```diff
@@ -0,0 +1,57 @@
...
+ * Compute the md5 sum of the named file
+ *
+ * @param fn Path to the file
+ *
+ * @return MD5 sum as a string; empty if an error ocurred.
+ */
+
+VITAL_UTIL_EXPORT std::string
+file_md5( const std::string& fn );
+
```
linus-sherrill: Is there any benefit for a function that operates on a istream?

collinsr: I'm not sure about either the API or the semantics-- you'd need some way to communicate which part of the stream the hash applies to. I guess the API could be a manipulator or something that doesn't consume the data, e.g.

```
string s;
while (is) {
   is >> md5ticker >> s;
   cout << "String is '" << s << "\n"
        << "md5 so far: " << md5ticker.hash()
        << " after " << md5ticker.bytes_seen() << " bytes\n";
}
```

...but that would be a very different implementation.

linus-sherrill: Maybe a string is more appropriate. It seems useful to be able to get the MD5 of a block of data, although I cannot come up with a time when I needed to do that.

collinsr: Maybe a char[] array (slightly more general than a string, since it could contain multiple nul bytes)... I think that would come up in e.g. crypto applications (although they've probably moved past md5, it's so 1990s...)

My use case for this is basically logging the state of the input for some DIVA tools; I'd like to have something a little more robust than the filename in the logs or metadata when I'm backtracking where homographies got misapplied or something like that. (The scoring code has this, and it's been pretty helpful a few times.)

Issue #760, opened 2019-02-28
Add image_container::get_image(x_offset, y_offset, width, height)


mleotta:
To support efficiently loading small crops of very large images we should add the method:
```c++
kwiver::vital::image_container::get_image(unsigned x_offset, unsigned y_offset,
                                          unsigned width, unsigned height);
```
For basic image containers that store images in memory, this method should just return a cropped view into the full image.  However, for containers that load from disk on demand (e.g. the GDAL container) this would allow them to only load the pixel that they need.  This is important for processing very large images, like satellite images, where you only need to load a small area of interest.

chetnieter:
Address in #767

PR #767, opened 2019-03-07
Add cropped get_image method to image containers

chetnieter:
Added a new get_image method the returns an image showing a cropped view of the full image from the container. In most cases this just uses the allocated memory of the full image. For the GDAL image container, only the memory needed to represent the cropped view is allocated.

This PR addressed #760.


chetnieter:
@mleotta PTAL

arrows/gdal/image_container.cxx
```diff
@@ -240,6 +240,30 @@ ::get_image() const
   return img;
 }
 
+// ----------------------------------------------------------------------------
+/// Return a cropped vital image. Unlike other image container must allocate memory.
```
mleotta: minor nitpick: This summary comment does not match the one in the header, so I think Doxygen will generate a description that concatenates both comments together, which is probably no what we want.

vital/types/image.h
```diff
@@ -537,6 +537,15 @@ class VITAL_EXPORT image
    */
   void set_size( size_t width, size_t height, size_t depth );
 
+  /// Get a cropped view of the image
+  /**
+   * Get a cropped view of the image.
```
mleotta: It may be worth being extra explicit here in the description that the cropped image returned shares memory with the source image and is not a deep copy.

chetnieter:
@mleotta Can I go ahead and merge this?

PR #770, opened 2019-03-14
Probes for missing KPF packets no longer invalidate the stream.

collinsr:
When the user calls reader.transfer_packet_from_buffer(), if it fails,
the reader should still be good. The catch is that the reader
also internally calls TPFB(); if called indirectly from the reader,
it should set the reader state to bad.

This commit adds a test case and adds an extra bool parameter to
TPFB (and the sibling transfer_kv_packet_from_buffer) which
defaults to false (don't change reader state) but is true
when called internally.


kfieldho:
Tested.  Looks Good To Me

PR #772, opened 2019-03-15
Be more forgiving when reading detected_object kpf

kfieldho:
We'll only demand an ID, a Frame Number and a Bounding box
in each record (suffficent to build a detected_object).  We'll
more gently look for an Index (track id), detector_name and
confidence and use them if they're there.  Also update the
writer to write out detected_object->index.  This has the side
effect of allowing "GEOM" KPF annotation files from DIVA to
be read in as detected objects and provides enough information
(by adding index) to add annotation labels in a separate
process.

### DIFF
--- a/arrows/kpf/detected_object_set_output_kpf.cxx
+++ b/arrows/kpf/detected_object_set_output_kpf.cxx
@@ -140,6 +140,8 @@ write_set( const kwiver::vital::detected_object_set_sptr set, std::string const&
     static std::atomic<unsigned> id_counter( 0 );
     const unsigned id = id_counter++;
 
+    auto det_index = (*det)->index();
+
     std::ostringstream oss;
     oss << "Record " << line_count++;
     w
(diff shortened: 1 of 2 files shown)


arrows/kpf/detected_object_set_output_kpf.cxx
```diff
@@ -148,6 +150,7 @@ write_set( const kwiver::vital::detected_object_set_sptr set, std::string const&
     w.set_schema( KPF::schema_style::GEOM )
       << KPF::writer< KPFC::kv_t >("detector_name", (*det)->detector_name())
       << KPF::writer< KPFC::id_t >(id, KPFC::id_t::DETECTION_ID)
+      << KPF::writer< KPFC::id_t >(id, KPFC::id_t::TRACK_ID)
```
collinsr: Just making sure I understand-- this is setting the track ID to be the same as the detection ID. A possible alternative would be to omit the track ID here and the reader would say "If the track ID is not otherwise set, use the detection ID."

kfieldho: That wasn't the intent.  The goal is to set both DETECTION_ID and TRACK_ID to two different values?

collinsr: Determined that det_index, not id, should be written here; fixed in subsequent commit.

PR #774, opened 2019-03-19
Visual Studio requires include algorithm for min max

dstoup:
BUG: The omission of include algorithm causes a build failure on Windows.


dstoup:
The Windows dashboard is broken. This branch helps some but the remainder of the dashboard needs to be ignored until we fix it.

PR #783, opened 2019-04-10
Perform Imports Alphabetically

mattdawkins:
This makes the log prettier.

But the real reason is this fixes an obscure tensorflow->opencv->pytorch import error when dealing with builds on all of the above, but only on some OS.

### DIFF
--- a/vital/bindings/python/vital/modules/loaders.py
+++ b/vital/bindings/python/vital/modules/loaders.py
@@ -143,7 +143,13 @@ class ModuleLoader(Loader):
                             break
 
     def _findPluginModules(self, namespace):
-        for filepath in self._findPluginFilePaths(namespace):
+
+        filepaths = [ p for p in self._findPluginFilePaths(namespace) ]
+
+        if len( filepaths ) > 0:
+            filepaths.sort( reverse=True )
+
+        for filepath in filepaths:
             path_segments = list(filepath.split(os.path.sep))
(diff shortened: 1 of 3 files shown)


linus-sherrill:
Why does sorting the paths solve the problem?

mattdawkins:
It changes the order of 'imports' as called by the python code. Fairly confident it's a conflict between dependencies tensorflow's build with and other libraries but won't have much time to look into for another month

mattdawkins:
Ameya found this:

https://github.com/tensorflow/tensorflow/issues/19010#issuecomment-386846086

so technically this is a bug in my old Ubuntu OS and this is the only fix; considering it doesn't hurt anything else I'd say it's still worth putting in this fix

as6520:
https://github.com/tensorflow/tensorflow/issues/19010#issuecomment-386846086 sheds some light on the problem but the import order is a mystery.

mattdawkins:
Ameya found this:

https://github.com/tensorflow/tensorflow/issues/19010#issuecomment-386846086

So it is an actual issue - don't think there's a better fix then this

linus-sherrill:
This is not a very robust solution to a dependency ordering problem. What if a file gets renamed? Generating a dependency graph would be a provably correct solution.

mattdawkins:
A dependency graph would require more complicated code for an issue which isn't going to come up often (or if all) again, not to mention having to hard-code or soft-code values in a config file?

PR #786, opened 2019-04-11
Fix GDAL warnings

mleotta:


arrows/gdal/image_container.cxx
```diff
@@ -227,14 +227,17 @@ ::get_image() const
 
   // Loop over bands and copy data
   CPLErr err;
-  for (int i = 1; i <= depth(); ++i)
+  for (size_t i = 1; i <= depth(); ++i)
   {
     GDALRasterBand* band = gdal_dataset_->GetRasterBand(i);
     auto bandType = band->GetRasterDataType();
     err = band->RasterIO(GF_Read, 0, 0, width(), height(),
       static_cast<void*>(reinterpret_cast<GByte*>(
```
dstoup: I realize this is existing code, but I just noticed the
    reinterpret_cast<GByte*>
Will this still work on 16-bit images?

mleotta: @chetnieter can confirm, but I think this is just using byte pointer arithmetic to find the first pixel of the `i-1` channel.  the `num_bytes` should account for number of bytes per pixel.

chetnieter: That is correct, the sixth argument to [RasterIO](https://www.gdal.org/classGDALRasterBand.html#a30786c81246455321e96d73047b8edf1) is just a generic pointer to the location of the first pixel in the image memory block. I think I arrived at the various type-casts by trial and error.  `bandType` tells GDAL the bit-depth used. One of the test cases reads in a 16-bit image from a GeoTIFF file.

PR #788, opened 2019-04-11
FFMPEG logging

mleotta:


dstoup:
The CentOS failure is random and not related to this PR. It can be ignored in this case.

arrows/ffmpeg/ffmpeg_init.cxx
```diff
@@ -35,22 +35,70 @@
...
+
 std::mutex ffmpeg_init_mutex;
 
-//--------------------------------------------------------------------------------
+static auto ffmpeg_logger = kwiver::vital::get_logger( "arrows.ffmpeg" );
+
+//-----------------------------------------------------------------------------
+// trim whitespace from end of string (in place)
+static inline void rtrim(std::string &s) {
+    s.erase(std::find_if(s.rbegin(), s.rend(),
```
linus-sherrill: There are string trimming functions in vital/util/string.h

arrows/ffmpeg/ffmpeg_init.cxx
```diff
@@ -35,22 +35,70 @@
...
+  av_log_format_line(ptr, level, fmt, vl, line, sizeof(line), &print_prefix);
+  std::string msg(line);
+  rtrim(msg);
+  switch(level)
+  {
+    case AV_LOG_PANIC:
+    case AV_LOG_FATAL:
+      LOG_ERROR(ffmpeg_logger, msg);
+      break;
+    case AV_LOG_ERROR:
```
linus-sherrill: Is it intended that AV_ERRORs get logged as warnings?

arrows/ffmpeg/ffmpeg_init.cxx
```diff
@@ -35,22 +35,70 @@
...
+}
+
+//-----------------------------------------------------------------------------
 void ffmpeg_init()
 {
   std::lock_guard< std::mutex > lock(ffmpeg_init_mutex);
   static bool initialized = false;
   if ( ! initialized ) {
     av_register_all();
     av_log_set_level(AV_LOG_ERROR);
```
linus-sherrill: I assume that this limits logging to ERRORS and above. This is also appears to be done by the logging properties file. It may be useful, at some point, to see more than error logs generated by FFMPEG.

arrows/ffmpeg/ffmpeg_init.cxx
mleotta: Yes,  FFMPEG has many more logging levels, as you can see.  I think "FATAL" is more inline with what we call an "ERROR".  I get hundreds of "ERROR" messages from FFMPEG on most videos because there is always some bad data somewhere, but there are recoverable, so I think KWIVER should treat them as warnings.

arrows/ffmpeg/ffmpeg_init.cxx
mleotta: That was left over from before.  I'm actually not sure if it does anything when not using the ffmpeg default logger.  During testing I was seeing message from FFMPEG at various different levels (warnings, info, etc.)  maybe I should just remove that line?

arrows/ffmpeg/ffmpeg_init.cxx
linus-sherrill: Makes sense

arrows/ffmpeg/ffmpeg_init.cxx
linus-sherrill: If it doesn't seem to have any effect, then you could remove it.

arrows/ffmpeg/ffmpeg_init.cxx
mleotta: okay, I'll use that

arrows/ffmpeg/ffmpeg_init.cxx
mleotta: I'll remove it and test again

mleotta:
All CI builds look clean.  This works for me and was approved by reviewers.  I'm going to merge.

PR #793, opened 2019-04-16
Dev/disable vla

kfieldho:
MS Visual Studio does not support variable length arrays (essentially array's declared with a non-constant size).  GNU C and Clang do.  This PR turns off that support to make them consistent with MSVC (and fixes the single instance in KWIVER where it mattered).  This way, code that builds on Linux/Mac OS have a greater likelihood of getting through MSVC testing.


arrows/serialize/protobuf/convert_protobuf.cxx
```diff
@@ -304,7 +304,7 @@ void convert_protobuf( const kwiver::vital::image_container_sptr img,
   const uLongf size = compressBound( local_image.size() );
   uLongf out_size(size);
   std::vector<uint8_t> image_data( size );
-  Bytef out_buf[size];
+  Bytef *out_buf = new Bytef[size];
```
linus-sherrill: An alternate implementation would be to use a `std::unique<Bytef>` to manage the storage.

PR #795, opened 2019-04-17
set FFMPEG metadata capability

mleotta:
The HAS_METADATA capability was not properly set for FFMPEG.  Also, improved logging messages about metadata stream detection.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_input.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_input.cxx
@@ -186,7 +186,7 @@ public:
 
     if (this->f_data_index < 0)
     {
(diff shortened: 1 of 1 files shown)


arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -497,6 +498,8 @@ ::open( std::string video_name )
     {
       throw kwiver::vital::video_runtime_exception("Video stream open failed for unknown reasons");
```
linus-sherrill: You can use VITAL_THROW here if the source location of the exception will be useful.

mleotta: That's good to know, but I think outside the scope of this branch.  I could address that on another branch.

mleotta:
Jenkins test this please

PR #796, opened 2019-04-30
Dev/remove Findpybind11

as6520:
* Removed Findpybind11 as the variables are provided by fletch
* Corrected includes to use lower case pybind11


dstoup:
Jenkins test this please

PR #799, opened 2019-05-15
AFRL MAP-Tk public released changes Apr 2019

mleotta:
This is the collection of changes on the MAP-Tk and TeleSculptor contracts over the last 4 months that were recently approved for public release by AFRL.  These changes have already been reviews internally at Kitware.


mleotta:
@linus-sherrill @kfieldho @collinsr @mattdawkins

All, I'm hoping to land this branch ASAP so I can make the v1.0 release of TeleSculptor before GEOINT and CVPR.  This is the latest round of AFRL public release changes for MAP-Tk/TeleSculptor since the last one in December.  All changes have been reviewed and tested internally already.  Most changes are specific to the 3D work with the exception of video input changes.

@chetnieter has improved the FFMPEG arrow with direct video input (bypasses VXL wrappers).  It now supports KLV metadata and much faster seeking than the VXL version.  The only limitation is that it does not have the logic used in the VXL arrow to set the timestamp from various configurable sources.  There are also extensive new and improved unit tests for video input (Thanks @chetnieter !).  

One issue is that the old FFMPEG 2.6 used in Fletch is buggy when it comes to seeking.  Some of the new seeking tests fail with FFMPEG 2.6 (this accounts for the CI failures on this branch).  Updating to FFMPEG 3.3 resolves this issue.  This is currently an option for 3.3 in Fletch, and there is a branch (Kitware/fletch#525) to make 3.3 the default.  Does that need to land first so we can get clean CI builds on this branch before merging?

The only problem with updating to FFMPEG 3.3 is that there may be some issues with the old VXL video reader.  @chetnieter may be able to comment more on this.  I'd like prefer to update to FFMPEG 3.3 by default and let folks using the old VXL video reader either
1. configure Fletch to build FFMPEG 2.6 as before,
2. update the VXL implementation to work with FFMPEG 3.3, or
3. switch to using the new direct FFMPEG reader (and add timestamp support if needed).

I'm not looking for a detailed review of this branch, but I'm looking to see if anyone has any objections to me merging it or thoughts about how we move forward with FFMPEG in Fletch/KWIVER.

chetnieter:
As @mleotta mentioned, some of the VXL ffmpeg reader tests fail on Windows for FFMPEG 3.3.3. In particular an FFMPEG library call `sws_getCachedContext` is returning a null pointer on Windows for FFMPEG 3.3.3. This does not happen for the VXL ffmpeg reader on MacOS or Linux.

mleotta:
Jenkins test this please

mleotta:
All CI builds have passed on this branch.  I plan to merge it on Monday unless there are any objections.

PR #800, opened 2019-05-16
Remove warnings from super3D code

linus-sherrill:


dstoup:
> 
> Do any of the Jenkins build machines turn OpenMP on?

I don't believe so. OpenMP is a recent addition and hasn't been added. I can start adding it to the nightly dashboards and do a run to make sure it's building everywhere, then add it to the PRs. If there is a doubt about this PR building correctly, we should wait until that's done and tested before merging.

PR #808, opened 2019-05-30
Prefer unix-style dependencies over Frameworks on macOS

mleotta:
Sometimes a build on macOS will find incompatible system Frameworks
versions of dependencies before those provided by Fletch.  This setting
gives unix-style packages, like those in Fletch, priority.


Cookt2:
LGTM

PR #814, opened 2019-06-24
Make Windowing Arrow around Arbitrary Detectors and Trainers

mattdawkins:


arrows/ocv/windowed_detector_resize.cxx
```diff
@@ -0,0 +1,111 @@
...
+  double scale = 1.0;
+
+  if( src.rows == height && src.cols == width )
+  {
+    dst = src;
+    return scale;
+  }
+
+  double original_height = static_cast< double >( src.rows );
+  double original_width = static_cast< double >( src.cols );
```
linus-sherrill: These could be const

arrows/ocv/windowed_detector_resize.h
```diff
@@ -0,0 +1,55 @@
...
+
+#ifndef KWIVER_ARROWS_OCV_WINDOWED_RESIZE
+#define KWIVER_ARROWS_OCV_WINDOWED_RESIZE
+
+
+#include <arrows/ocv/kwiver_algo_ocv_export.h>
+
+#include <opencv2/core/core.hpp>
+
+#include <vital/algo/image_object_detector.h>
```
linus-sherrill: Is this needed here? It doesn't look like it is needed for the interface.

as6520:
I think we should refactor this into three algorithms rather than handling all these operations into a single algorithm. 
1. ImageChipper: An algorithm that creates image chips. It would take an ImageContainer and returns list of ImageContainer and list of Metadata for each of the ImageContainer.
2. DetectedObjectSetScaler: Companion algorithm for Image Chipper that would take detected object set for a chip along with the metadata of the chip and scales the detected object set based on the chip
2. WindowedDetector: Has an instance of ImageChipper, ImageObjectDetector, DetectedObjectSetScaler. It gets the chips from the ImageChipper, schedules the chips on one or more instances  of ImageObjectDetector and scales up the detection using DetectedObjectSetScaler. 

We can also provide reasonably generic support for multiple gpus with this as individual instances of ImageObjectDetector can run on different gpus without relying on an implementation to support multi gpu operation.

PR #815, opened 2019-06-24
Fix kw18 track reader for active mode

mattdawkins:


linus-sherrill:
What was the problem and how did this fix it?

linus-sherrill:
Is there a test for this problem/fix?

linus-sherrill:
Nudge

PR #817, opened 2019-06-26
Dev/refactor image python bindings

as6520:
This branch is intended to propose a refactored version of the vital types python bindings. The primary motivation is to be able to reuse code across bindings and is applicable to functions like asarray.

The following changes are proposed
1. Split the binding into header and source
2. Move the custom functions used for the binding in a namespace kwiver::vital::python::vital_type_name
3. Add an additional file type_module.cxx for all the type modules.


linus-sherrill:
Looks reasonable to me but I did not do a thorough examination.

PR #818, opened 2019-06-27
Dev/reuse image binding code

as6520:
Depends on #817 
1. Reuse asarray logic from image bindings inside image container bindings.
2. The vital image in fromarray is created using new_image_from_numpy in image binding.
3. numpy array in asarray is created by checking the attributes of pixel trait. Creating a py::object from the image sometimes returns a numpy array with data type int8 rather than uint8. This occurs when asarray is called on an image in the image container obtained from another process in the pipeline. In this case, the asarray function of the ImageContainer returns a numpy array with the correct data type.

### DIFF
--- a/vital/bindings/python/vital/tests/helpers.py
+++ b/vital/bindings/python/vital/tests/helpers.py
@@ -46,7 +46,6 @@ from vital.types import (
     EigenArray,
     Feature,
     Landmark,
-    LandmarkMap,
     Rotation,
     Track,
     TrackSet,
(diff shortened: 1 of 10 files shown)


vital/bindings/python/vital/tests/test_image_container.py
```diff
@@ -66,3 +68,79 @@ def test_height(self):
...
+                       'float64']
+
+        def _test_numpy(dtype_name, nchannels, order='c'):
+            np_img = create_numpy_image(dtype_name, nchannels, order)
+            img_container = ImageContainer.fromarray(np_img)
+            recast = img_container.asarray()
+
+            if nchannels is None:
+                # asarray always returns 3 channels
+                np_img = np_img[..., None]
```
Erotemic: Can you save a line using `np.atleast_3d`?

vital/bindings/python/vital/types/image_container.cxx
```diff
@@ -151,33 +132,23 @@ PYBIND11_MODULE(image_container, m)
...
-          pil_img = PILImage.fromarray(arr)
-          vital_img = VitalPIL.from_pil(pil_img)
-          self = ImageContainer(vital_img)
-          )", py::globals(), locals);
-      return locals["self"].cast<s_image_cont>();
-    }, py::doc(R"(
-    Create an ImageContainer from a numpy array
-
-    Example:
-        >>> from vital.types import ImageContainer
```
Erotemic: Are these tests replicated somewhere else?

vital/bindings/python/vital/tests/helpers.py
```diff
@@ -251,41 +175,39 @@ def reprojection_error_sqr(cam, lm, feat):
     return (reprojection_error_vec(cam, lm, feat) ** 2).sum()
 
 
-def reprojection_rmse(cmap, lmap, tset):
```
Erotemic: Why was this removed? It seems unrelated to the goal of this PR. 

If its truly cruft code, I don't mind removing it in this PR, I just want to be sure it is not needed.

vital/bindings/python/vital/types/image_container.cxx
as6520: Not as a doc test. However, [fromarray test in image container](https://github.com/Kitware/kwiver/pull/818/files/8d3daa35c8d298c057a4c239e191cf6585a43e90#diff-3a76da4c16d20f43318ff80409d1380dR80) checks for it with all types we support.

vital/bindings/python/vital/tests/helpers.py
as6520: It requires landmarkMap as the second parameter. I believe landmarkMap bindings were not present in the transition to pybind11 as I can see the c bindings for it but not the python bindings for it. 

I agree that it is not related to this PR but to make helpers.py work LandmarkMap had to be removed from the imports. Since we had multiple helpers that used LandmarkMap as one of the parameters or returned LandmarkMap, I removed them too.

vital/bindings/python/vital/types/image_container.cxx
```diff
@@ -151,33 +132,23 @@ PYBIND11_MODULE(image_container, m)
...
-  ;
+  .def(py::init(&kwiver::vital::python::image_container::new_cont), py::arg("image"))
+
+  // Create initialzer based on numpy array type
+  #define fromarray( T ) \
+  .def_static("fromarray", \
+              &kwiver::vital::python::image_container::new_image_container_from_numpy<T>, \
+              py::arg("array"),\
+      py::doc("Create an ImageContainer from a numpy array"))
+  fromarray( uint8_t )
```
Erotemic: Unless this is following some C++ convention, it may be more clear to rename this macro from `fromarray` to `def_fromarray` to indicate that the macro is creating a definition.

linus-sherrill:
If there are no open issues with this PR, I will merge it on the 9th

PR #725, opened 2019-01-16
Add VXL image filters (color frequency, pixel-level averaging, image conversion)

mattdawkins:

PR #820, opened 2019-07-01
Include imgproc header in convert_color_space

neal-siekierski:
Also include .hpp files from the opencv2 directory, not .h files from
the opencv directory.

This allows the call to cv::cvtColor to work, at least under OpenCV
3.4.0.

### DIFF
--- a/arrows/ocv/convert_color_space.cxx
+++ b/arrows/ocv/convert_color_space.cxx
@@ -11,8 +11,8 @@
 
 #include <arrows/ocv/image_container.h>
 
-#include <opencv/cxcore.h>
-#include <opencv/cv.h>
+#include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
 
 namespace kwiver {
 namespace arrows {


neal-siekierski:
This still uses the legacy headers; I don't know whether using `opencv2/core.hpp` and `opencv2/imgproc.hpp` is compatible with OpenCV 2, though.

Let me know if the base branch should be something other than viame/release.

mattdawkins:
There's an exisiting PR for this into master:

https://github.com/Kitware/kwiver/pull/725

You might want to add it to (then we can merge master in viame/master once it lands, that branch could use a review too)

neal-siekierski:
What do you mean? `arrows/ocv/convert_color_space.cxx`, which this PR changes, isn't part of that one, and adding all of `convert_color_space` seems out of scope for that PR, since it's unrelated to VXL.

mattdawkins:
Hmm, my bad I guess I didn't cherry-pick it onto that one, I'll cherry-pick these commits and others onto a branch after the VXL one lands

PR #829, opened 2019-07-26
update travis configuration

Cookt2:
- Updates Travis target OS from Ubuntu Trusty to Bionic
- Updates gcc version from 4.8 to 7.0 (default)
- Updates clang version from 3.4 to 3.7 (also default)
- Uses the CMake that travis CI has by default (3.12.X) rather than installing from source
- Changes to use a Fletch binary that is generated on Ubuntu Bionic 18.04 rather than Ubuntu Trusty 14.04


dstoup:
Looking good!

PR #830, opened 2019-08-07
Dev/cuda docker file

as6520:


dstoup:
I think there is some work done here that might conflict with a change @Cookt2 made and merged recently though the PR doesn't show any. Otherwise, it looks good to me. I will wait for Taylor to chime in.

dockerfile
```diff
@@ -1,20 +1,46 @@
 # Install KWIVER to /opt/kitware/kwiver
 # Use latest Fletch as base image (Ubuntu 18.04)
+ARG BASE_IMAGE=kitware/fletch:latest
```
Cookt2: This should be changed to: "kitware/fletch:latest-ubuntu18.04-py3"

With the latest changes to the fletch dockerfile, we are now producing different labeling syntax.

The default base fletch image is now "kitware/fletch:latest-ubuntu18.04-py3", I plan on removing "kitware/fletch:latest" altogether.

dockerfile
```diff
@@ -1,20 +1,46 @@
...
+ARG PY_MAJOR_VERSION=3
+ARG ENABLE_CUDA=OFF
+ARG KWIVER_INSTALL=/opt/kitware/kwiver
 
+# Conditonal installtion of six
+RUN if [ "$PY_MAJOR_VERSION" = "2" ]; then \
+      pip install six; \
+    else \
+      pip3 install six; \
+    fi
```
Cookt2: Why are we installing six? I haven't had it in the previous dockerfiles.

linus-sherrill:
Is this ready to go or is there still a problem?

as6520:
The original intent of this PR was to provide a dockerfile that would build base images for DIVA. Thus the dockerfile using the components that diva requires to build. I discussed this PR with @Cookt2 and that is not the intent of the dockerfile. Thus I am closing the PR.

PR #836, opened 2019-08-21
Camera distortion limits

mleotta:
This branch adds functions to the camera intrinsics to test if a point lies in the valid input domain of the mapping function.  In particular, this is implemented in the `simple_camera_intrinics` as a maximum radius of distortion.  For radial distortion we can set a maximum radius bounding the domain of distortion.  For pincushion distortion maximum radius is infinity.  However, for barrel distortion there is often a radius beyond which the distortion wraps back around on itself.  3D points that are far outside the field of view, but still in front of the camera plane may erroneously project into the image bounds.  

This branch automatically finds the maximum radius for radial distortion and caches it whenever the distortion parameters are set or changed.  The maximum radius is the first (smallest positive) inflection point of the radial distortion function.  This involves solving a cubic polynomial in the general case.  The math for that cubic case is a bit hairy.  I would not expect anyone to validate the equations. The test cases validate that the solution works.


mattbrown11:
Another consideration is short-circuiting projection calculations for points that are valid (i.e., they exceed the valid range of the radial distortion function) but will necessarily be off the focal plane. For example, if during initialization of the camera model, you unprojected a bunch of points from the perimeter of the image-plane boundary and collected the max_distort_radius_sq_, you would know that any world point with a distort_radius_sq larger than this will necessarily not be in the image, skipping additional projection calculations. I could still see use cases where you want to model projections that land off of the focal plane during bundle adjustment, though.

bclipp-kitware:
@mattbrown11 @mleotta 

In all the previous code I've worked on we did this.

1.  Project the point with the linear model (no distortion applied)
2.  Check if it's in the image boundary  if not stop
3.  Now distort the point with the distortion model. 
4.  Check that the point is still in the image.  (Points that linearly project to the edge of the image may map outside the image with the radial distortion applied.  This kind of pairing would not happen when fitting the radial distortion model. So these points are invalid.)

The model is only valid for points that project linearly to the image plane.  So this check works (2) to prevent wrap around issues etc.

mattbrown11:
Brian and I talked through this whole issue quite in depth a while ago. I re-read our previous chats.

Brain's algorithm starts during initialization (at least conceptually) by taking points along the perimeter of the focal plane and sending them back out of the distortion function. This effectively produces a boundary in normalized image coordinates (either barrel or pincushion) bounding the rays that can make it onto the focal plane. You can circumscribe this shape with a circle or rectangular, and then at runtime, if a point doesn't project inside that circle/rectangle, then it cannot make it onto the focal plane. This becomes a more-conservative bound on max_distort_radius_sq_.

A counter argument is that there may be cases where you want to know where points project even if they are off of the focal plane (e.g., during bundle adjustment).

mleotta:
I think I'm addressing a slightly different problem than you guys have in
mind.  My problem is visualization of the 3D scene in the context of the
source image.  I'm not trying to limit points to the bounds of the image.
I'd like to show points that project into the image as well as points that
project outside of the image to provide more context.  Initially, without
distortion, I was showing everything in front of the image plane.  I
limited this to within a radius of about 10x the image width, but only for
numerical precision reasons.  I'd like to show everything in front of the
camera, but when the geometry distorts back on itself the visualization
gets confusing.

In short, I want the radius to be as large as possible such that the
distortion function is still uniquely invertable in that domain.

A test for whether a point projects into the image bounds is a tighter
constraint that has value in other applications.

--Matt

linus-sherrill:
Is this branch still active?

mleotta:
I hope so.  I'd still like to get this merged.  It was useful on DIVA.  I actually have a half-written paper documenting this that may try to publish some day.

linus-sherrill:
This has a minor conflict that needs to be resolved before merge

PR #838, opened 2019-09-02
Dev/camera to camera detection transform

dmjoy:
This pull request adds an arrow for transforming detected object sets from one camera view to another (as an implementation of detected_object_set_filter).  The arrow name should probably be a bit more descriptive.

### DIFF
--- a/arrows/core/transform_detected_object_set.cxx
+++ b/arrows/core/transform_detected_object_set.cxx
@@ -44,15 +44,17 @@ namespace arrows {
 namespace core {
 
 // ------------------------------------------------------------------
-transform_detected_object_set::transform_detected_object_set()
+transform_detected_object_set::
+transform_detected_object_set()
   : src_camera_krtd_file_name( "" )
   , dest_camera_krtd_file_name( "" )
 {
 }
 
 // ------------------------------------------------------------------
-transform_detected_object_set::transform_detected_object_set(kwiver::vital::camera_perspective_sptr src_cam,
-							     kwiver::vital::camera_perspective_sptr dest_cam)
+transform_detected_object_set::
+transform_detected_object_set(kwiver::vital::camera_perspective_sptr src_cam,
+			      kwiver::vital::camera_perspective_sptr dest_cam)
   : src_camera( src_cam )
   , dest_camera( dest_cam )
 {
@@ -83,11 +85,15 @@ set_configuration( vital::config_block_sptr config_in )
   vital::config_block_sptr config = this->get_configuration();
 
   config->merge_config( config_in );
-  this->src_camera_krtd_file_name = config->get_value< std::string > ( "src_camera_krtd_file_name" );
-  this->dest_camera_krtd_file_name = config->get_value< std::string > ( "dest_camera_krtd_file_name" );
+  this->src_camera_krtd_file_name =\
+    config->get_value< std::string > ( "src_camera_krtd_file_name" );
+  this->dest_camera_krtd_file_name =\
+    config->get_value< std::string > ( "dest_camera_krtd_file_name" );
 
-  this->src_camera = kwiver::vital::read_krtd_file( this->src_camera_krtd_file_name );
-  this->dest_camera = kwiver::vital::read_krtd_file( this->dest_camera_krtd_file_name );
+  this->src_camera =\
+    kwiver::vital::read_krtd_file( this->src_camera_krtd_file_name );
+  this->dest_camera =\
+    kwiver::vital::read_krtd_file( this->dest_camera_krtd_file_name );
 }
 
 
@@ -101,7 +107,8 @@ check_configuration( vital::config_block_sptr config ) const
 
   if ( ! key_list.empty() )
   {
-    LOG_WARN( logger(), "Additional parameters found in config block that are not required or desired: "
+    LOG_WARN( logger(), "Additional parameters found in config block that are "
+	      "not required or desired: "
               << kwiver::vital::join( key_list, ", " ) );
   }
 
@@ -150,12 +157,14 @@ backproject_bbox( const kwiver::vital::camera_perspective_sptr camera,
   // project base of the box to the ground
   auto base_pt = vector_2d((box(0) + box(2)) / 2, box(3));
   vector_3d pc = this->backproject_to_ground( camera, base_pt );
-  vector_3d ray = pc - (-camera->rotation().matrix().transpose() * camera->translation());
+  vector_3d ray = pc - camera->center();
   ray(2) = 0.0;
   ray.normalize();
 
-  vector_3d p1 = this->backproject_to_ground( camera, vector_2d(box(0), box(3)) );
-  vector_3d p2 = this->backproject_to_ground( camera, vector_2d(box(2), box(3)) );
+  vector_3d p1 =\
+    this->backproject_to_ground( camera, vector_2d(box(0), box(3)) );
+  vector_3d p2 =\
+    this->backproject_to_ground( camera, vector_2d(box(2), box(3)) );
 
   vector_3d vh = p2 - p1;
   vector_3d vd = ray * vh.norm();
@@ -172,7 +181,8 @@ backproject_bbox( const kwiver::vital::camera_perspective_sptr camera,
 
   vector_4d back_plane;
   back_plane << n, d;
-  vector_3d p5 = this->backproject_to_plane( camera, vector_2d(box(0), box(1)), back_plane );
+  vector_3d p5 =\
+    this->backproject_to_plane( camera, vector_2d(box(0), box(1)), back_plane );
   double height = p5(2);
 
   Eigen::Matrix<double, 8, 3> box3d;
@@ -197,7 +207,8 @@ box_around_box3d( const kwiver::vital::camera_perspective_sptr camera,
   Eigen::Matrix<double, 8, 2> projected_points;
 
   // project the points
-  for (int i=0; i<8; i++) {
+  for (int i=0; i<8; i++)
+  {
     projected_points.row(i) = camera->project(box3d.row(i)).transpose();
   }
 
@@ -215,7 +226,8 @@ view_to_view( const kwiver::vital::camera_perspective_sptr src_camera,
 	      const kwiver::vital::camera_perspective_sptr dest_camera,
 	      const vector_4d bounds ) const
 {
-  Eigen::Matrix<double, 8, 3> box3d = this->backproject_bbox( src_camera, bounds );
+  Eigen::Matrix<double, 8, 3> box3d =\
+    this->backproject_bbox( src_camera, bounds );
   vector_4d tgt_box = this->box_around_box3d( dest_camera, box3d );
 
   return tgt_box;
@@ -231,8 +243,8 @@ transform_bounding_box( vital::bounding_box<double>& bbox ) const
     bbox.max_x(), bbox.max_y();
 
   vector_4d out_bounds = this->view_to_view( this->src_camera,
-						   this->dest_camera,
-						   bbox_bounds );
+					     this->dest_camera,
+					     bbox_bounds );
 
   return vital::bounding_box<double>(out_bounds(0),
 				     out_bounds(1),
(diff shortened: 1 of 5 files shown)


doc/release-notes/master.txt
```diff
@@ -29,6 +29,10 @@ Arrows
 
 Arrows: Core
 
+ * Added an arrow (transform_detected_object_set) for transforming detected
```
mleotta: Minor terminology nitpick here:  The "arrow" is usually what we call the plugin library.  So in this case you are adding an algorithm implementation to the existing "core" arrow.

arrows/core/transform_detected_object_set.h
```diff
@@ -0,0 +1,110 @@
...
+
+using namespace kwiver::vital;
+
+namespace kwiver {
+namespace arrows {
+namespace core {
+
+// ----------------------------------------------------------------
+/**
+ * @brief Transforms detections based on source and destination cameras.
```
mleotta: Instead `@brief` in this large comment block, we usually add a single line starting with `///` (note 3 slashes, not 2) right before the first line of the class or function.  See other files for examples.

arrows/core/transform_detected_object_set.h
```diff
@@ -0,0 +1,110 @@
...
+class KWIVER_ALGO_CORE_EXPORT transform_detected_object_set
+  : public vital::algorithm_impl<transform_detected_object_set,
+				 vital::algo::detected_object_filter>
+{
+public:
+  PLUGIN_INFO( "transform_detected_object_set",
+               "Transforms a detected object set based on source and "
+	       "destination cameras.\n\n" )
+
+  transform_detected_object_set();
```
mleotta: In general, we should have at least a brief comment (starting with `///`, see above comment) for each member function and variable in the header.  You can also add more detail comments below the brief line when needed.  We only copy the brief comment line into the CXX.

arrows/core/transform_detected_object_set.cxx
```diff
@@ -0,0 +1,275 @@
...
+#include <vital/types/bounding_box.h>
+#include <Eigen/Core>
+
+using namespace kwiver::vital;
+
+namespace kwiver {
+namespace arrows {
+namespace core {
+
+// ------------------------------------------------------------------
```
mleotta: This line serves two visual purposes.  First as a vertical separator between functions, but second as a horizontal ruler to show the target maximum line length of 80 characters.  Your lines are a bit short.  Use this:
```c++
// ---------------------------------------------------------------------------
``` 
Also, if you see any lines that are longer than this bar it is a good idea to try to wrap them neatly.  This line length limit is not a hard requirement in KWIVER, but we try to enforce it when reasonable.

arrows/core/transform_detected_object_set.cxx
```diff
@@ -0,0 +1,275 @@
...
+void
+transform_detected_object_set::
+set_configuration( vital::config_block_sptr config_in )
+{
+  vital::config_block_sptr config = this->get_configuration();
+
+  config->merge_config( config_in );
+  this->src_camera_krtd_file_name =\
```
mleotta: Are these trailing slashes intended here?  These look like a Pythonic line continuation, but that is not needed in C++.  I'm actually not sure what effect that slash has, but I would remove it here and elsewhere in this function.

arrows/core/transform_detected_object_set.cxx
```diff
@@ -0,0 +1,275 @@
...
+
+  return true;
+}
+
+
+// ------------------------------------------------------------------
+vector_3d
+transform_detected_object_set::
+backproject_to_ground( const kwiver::vital::camera_perspective_sptr camera,
+		       const vector_2d img_pt ) const
```
mleotta: The indentation here is does not line up with the previous line.  You also might have some tabs here instead of spaces.  I would convert your tabs to spaces and then line up lines like this throughout the file using spaces.  It may help to tell your editor to use spaces instead of tabs for C++ files.  Most editors support that.

arrows/core/transform_detected_object_set.cxx
```diff
@@ -0,0 +1,275 @@
...
+
+  vector_4d back_plane;
+  back_plane << n, d;
+  vector_3d p5 =\
+    this->backproject_to_plane( camera, vector_2d(box(0), box(1)), back_plane );
+  double height = p5(2);
+
+  Eigen::Matrix<double, 8, 3> box3d;
+  box3d << p1.transpose(),
+    p2.transpose(),
```
mleotta: I would indent so that `p2` is below `p1` and use the same indentation for following lines.

dmjoy:
Thanks Matt, I believe I've addressed each of your requested changes, but please let me know if that's not the case.

arrows/core/transform_detected_object_set.h
mleotta: You actually don't need the `@brief` here when using `///`.  Basically
```c++
/// This is my brief comment
```
means the same thing as
```c++
/** 
 * @brief This is my brief comment
 */
```

arrows/core/transform_detected_object_set.cxx
```diff
@@ -0,0 +1,276 @@
...
+  vector_3d Mp = M * npt;
+
+  return Mp * (n.dot( Mt ) - d) / n.dot( Mp ) - Mt;
+}
+
+// ---------------------------------------------------------------------------
+Eigen::Matrix<double, 8, 3>
+transform_detected_object_set::
+backproject_bbox( const kwiver::vital::camera_perspective_sptr camera,
+                  const vector_4d box ) const
```
mleotta: It might make the code easier to understand if you just pass `box` as a `vital::bounding_box<double>` all the way into this function.  Then you can access `box.min_x()` instead of `box(0)` and it's more clear what that is.  There is nothing technically wrong with what you have here, but using the more specific type will help self-document the function a little better.

This change is not a requirement, just a suggestion.

arrows/core/transform_detected_object_set.h
dmjoy: Ah, I see.  Should be fixed now.

arrows/core/transform_detected_object_set.cxx
dmjoy: Agreed that that's a more transparent approach, should be fixed now.

dmjoy:
Thanks again for the feedback, I've pushed up the requested changes.

arrows/core/transform_detected_object_set.cxx
```diff
@@ -182,8 +182,8 @@ backproject_bbox( const kwiver::vital::camera_perspective_sptr camera,
 
   vector_4d back_plane;
   back_plane << n, d;
-  vector_3d p5 =
-    this->backproject_to_plane( camera, vector_2d(box(0), box(1)), back_plane );
+  vector_3d p5 = this->backproject_to_plane(
+    camera, vector_2d(bbox.min_x(), bbox.min_y()), back_plane );
```
mleotta: not critical, but `bbox.upper_left()` is the same as `vector_2d(bbox.min_x(), bbox.min_y())`

arrows/core/transform_detected_object_set.cxx
```diff
@@ -213,23 +213,24 @@ box_around_box3d( const kwiver::vital::camera_perspective_sptr camera,
     projected_points.row(i) = camera->project(box3d.row(i)).transpose();
   }
 
-  vector_4d out_box;
-  out_box << (vector_2d)projected_points.colwise().minCoeff(),
-             (vector_2d)projected_points.colwise().maxCoeff();
+  auto out_box = vital::bounding_box<double>(
```
mleotta: you could bypass the temporary variable `out_box` and just directly `return vital::bounding_box(`...

arrows/core/transform_detected_object_set.cxx
```diff
@@ -213,23 +213,24 @@ box_around_box3d( const kwiver::vital::camera_perspective_sptr camera,
...
 }
 
 // ---------------------------------------------------------------------------
-vector_4d
+vital::bounding_box<double>
 transform_detected_object_set::
 view_to_view( const kwiver::vital::camera_perspective_sptr src_camera,
               const kwiver::vital::camera_perspective_sptr dest_camera,
-              const vector_4d bounds ) const
+              const vital::bounding_box<double> bbox ) const
```
mleotta: It's probably better for efficiency to pass `bbox` as  `vital::bounding_box<double> const& bbox`.  The const reference avoids making a deep copy when pass it in.  This is a small enough structure that it probably doesn't make much difference, but it's good practice for anything bigger than the built in basic types.  The other exception here is that we usually pass shared pointer by value (though there is some debate about whether we should) since they are intended to make a light-weight shallow copy.

Also, the `const` can go before or after the type name.  It is now KWIVER style to put the const after the type name.  See here https://github.com/Kitware/kwiver/blob/master/doc/code-style.rst#type-names-types

arrows/core/transform_detected_object_set.cxx
```diff
@@ -213,23 +213,24 @@ box_around_box3d( const kwiver::vital::camera_perspective_sptr camera,
...
 {
   Eigen::Matrix<double, 8, 3> box3d =
-    this->backproject_bbox( src_camera, bounds );
-  vector_4d tgt_box = this->box_around_box3d( dest_camera, box3d );
+    this->backproject_bbox( src_camera, bbox );
+  vital::bounding_box<double> tgt_box =
```
mleotta: Here is another case where you can eliminate the temporary variable a directly return the result of the function call.

arrows/core/transform_detected_object_set.cxx
```diff
@@ -239,18 +240,11 @@ vital::bounding_box<double>
...
-
-  vector_4d out_bounds = this->view_to_view( this->src_camera,
-                                             this->dest_camera,
-                                             bbox_bounds );
-
-  return vital::bounding_box<double>(out_bounds(0),
-                                     out_bounds(1),
-                                     out_bounds(2),
-                                     out_bounds(3));
+  vital::bounding_box<double> out_bbox = this->view_to_view( this->src_camera,
```
mleotta: Here is another case where you can eliminate the temporary variable a directly return the result of the function call.

arrows/core/transform_detected_object_set.h
```diff
@@ -96,16 +96,16 @@ class KWIVER_ALGO_CORE_EXPORT transform_detected_object_set
 
   virtual Eigen::Matrix<double, 8, 3>
     backproject_bbox(const kwiver::vital::camera_perspective_sptr camera,
-                     const vector_4d box) const;
+                     const vital::bounding_box<double> bbox) const;
```
mleotta: see previous comments about passing by `const&`

arrows/core/transform_detected_object_set.h
```diff
@@ -96,16 +96,16 @@ class KWIVER_ALGO_CORE_EXPORT transform_detected_object_set
...
+  virtual vital::bounding_box<double>
     box_around_box3d(const kwiver::vital::camera_perspective_sptr camera,
                      const Eigen::Matrix<double, 8, 3> box3d) const;
 
-  virtual vector_4d
+  virtual vital::bounding_box<double>
     view_to_view(const kwiver::vital::camera_perspective_sptr src_camera,
                  const kwiver::vital::camera_perspective_sptr dest_camera,
-                 const vector_4d bounds) const;
+                 const vital::bounding_box<double> bbox) const;
```
mleotta: see previous comments about passing by `const&`

arrows/core/transform_detected_object_set.h
```diff
@@ -0,0 +1,114 @@
...
+
+private:
+  std::string src_camera_krtd_file_name;
+  std::string dest_camera_krtd_file_name;
+
+  kwiver::vital::camera_perspective_sptr src_camera;
+  kwiver::vital::camera_perspective_sptr dest_camera;
+
+  virtual vital::bounding_box<double>
+    transform_bounding_box(vital::bounding_box<double>& bbox) const;
```
mleotta: The input here should be `vital::bounding_box<double>const& bbox`, note the `const`.  This is because you are not changing the memory of `bbox` that is passed in.  Instead you are returning a new object with the transformed values.  An alternative approach is to modify the memory in place by passing `bbox` as you are, modifying that object in place, and then returning `void` or maybe `bool` (if the function could fail).  I think the best option here is to pass by `const&`.

arrows/core/transform_detected_object_set.cxx
dmjoy: Ah, I see.  I've made the suggested pass by reference change here and elsewhere where it seemed appropriate (including your other suggested locations).

Gotcha, I've made that change as well.

arrows/core/transform_detected_object_set.cxx
dmjoy: Aye, that's certainly cleaner.  Fixed here and the other suggested locations as well.

arrows/core/transform_detected_object_set.h
dmjoy: Understood, fixed!

PR #840, opened 2019-09-09
Add altitude to geo_point; add geo_covariance type

aaron-bray:


aaron-bray:
@linus-sherrill or @mwoehlke-kitware 
I think this looks pretty straight forward, but I am not sure if my changes to metadata_io or the vital/klv stuff could have any negative impacts, or if I am missing anything else I may need to change. (There is no python geo_point binding, so I think I'm good there)

PR #842, opened 2019-09-11
Wrap f2f_homography

neal-siekierski:
The current changes for this PR form a (relatively minimal) Python wrapping of the `f2f_homography` type that I wrote in order to get the `homography_writer` process working, together with some changes for the latter. It suffices for that purpose, but may fall short on other grounds:
- To get the wrapped `f2f_homography` instance out of the wrapped port in `homography_writer`, I added a method to wrapped `datum` type. Is there a better way to do this? (Traits? I'm not very familiar with them.)
- One of the constructors for the wrapped `f2f_homography` class, along with its `homography` attribute, don't work as is (failing with a `TypeError` at run-time) because the `homography` type is wrapped as a separate class. I've started looking at changing the wrapping for `homography`, but I don't know if that's the right way forward.
- There are no tests (related to the unusable constructor).

### DIFF
--- a/sprokit/processes/python/homography_writer.py
+++ b/sprokit/processes/python/homography_writer.py
@@ -50,7 +50,7 @@ class HomographyWriterProcess(sprokit.pipeline.process.PythonProcess):
     # ----------------------------------------------------------------
     def _step(self):
         dat = self.grab_datum_from_port('homography')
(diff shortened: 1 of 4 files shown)


mattdawkins:
LGTM

PR #844, opened 2019-09-13
Local cartesian conversion utility

aaron-bray:
Based on the NGA GeoTrans Library

### DIFF
--- a/vital/tests/test_local_cartesian.cxx
+++ b/vital/tests/test_local_cartesian.cxx
@@ -90,40 +90,57 @@ TEST(geo_point, api)
   EXPECT_EQ(lc1.get_orientation(), 22);
 }
 
+void CompareLLA(kwiver::vital::vector_3d gp1, kwiver::vital::vector_3d gp2)
+{
+  EXPECT_NEAR(gp1[0], gp2[0], 1e-8);
+  EXPECT_NEAR(gp1[1], gp2[1], 1e-8);
+  EXPECT_NEAR(gp1[2], gp2[2], 1e-3);
+}
+
 // ----------------------------------------------------------------------------
 TEST(geo_point, conversion)
 {
   kwiver::vital::plugin_manager::instance().load_all_plugins();
 
-  kwiver::vital::geo_point geo_out;
-  kwiver::vital::vector_3d cart_out;
+  kwiver::vital::geo_point geo_outA;
+  kwiver::vital::vector_3d cart_outA;
   kwiver::vital::local_cartesian lc_lla(origA);
 
   // Get the geopoint from an offset
-  lc_lla.convert_from_cartesian(offset1, geo_out);
-  auto const epsilon_geo_out_to_geo1 = (geo_out.location() - geo1.location()).norm();
-  EXPECT_MATRIX_NEAR(geo_out.location(), geo1.location(), 1e-3);
-  EXPECT_LT(epsilon_geo_out_to_geo1, 1e-3);
+  lc_lla.convert_from_cartesian(offset1, geo_outA);
+  CompareLLA(geo_outA.location(), geo1.location());
 
   // Now get the cartesian value from that geo_point
-  lc_lla.convert_to_cartesian(geo_out, cart_out);
-  auto const epsilon_cart_out_to_offset1 = (cart_out - offset1).norm();
-  EXPECT_MATRIX_NEAR(cart_out, offset1, 1e-8);
-  EXPECT_LT(epsilon_cart_out_to_offset1, 1e-8);
+  lc_lla.convert_to_cartesian(geo_outA, cart_outA);
+  CompareLLA(cart_outA, offset1);
 
   // Get the geopoint from another offset
-  lc_lla.convert_from_cartesian(offset2, geo_out);
-  auto const epsilon_geo_out_to_geo2 = (geo_out.location() - geo2.location()).norm();
-  EXPECT_MATRIX_NEAR(geo_out.location(), geo2.location(), 1e-3);
-  EXPECT_LT(epsilon_geo_out_to_geo2, 1e-3);
+  lc_lla.convert_from_cartesian(offset2, geo_outA);
+  CompareLLA(geo_outA.location(), geo2.location());
 
   // Now get the cartesian value from that geo_point
-  lc_lla.convert_to_cartesian(geo_out, cart_out);
-  auto const epsilon_cart_out_to_offset2 = (cart_out - offset2).norm();
-  EXPECT_MATRIX_NEAR(cart_out, offset2, 1e-8);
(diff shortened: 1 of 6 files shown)


vital/types/local_cartesian.h
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * Set local origin parameters as inputs and sets the corresponding state variables.
+   * NOTE : If the origin changes, this method needs to be called again to recompute
+   *        variables needed in the conversion math.
+   *
+   * @param [in] geo_point       : Geograpical origin of the cartesian system
+   * @param [in] orientation     : Orientation angle of the local cartesian coordinate system,
+   *                               in radians along the Z axis, normal to the earth surface
+   */
+  void set_origin( geo_point const& origin, double orientation=0 );
```
mleotta: Should orientation be a full 3D rotation?  If we are going to add orientation, why limit it to the Z axis?  I'm not saying this has to change, I'm just curious about the motivation for this.

vital/types/local_cartesian.cxx
```diff
@@ -0,0 +1,524 @@
...
+
+namespace kwiver {
+namespace vital {
+
+const double PI = 3.14159265358979323e0;
+const double PI_OVER_2 = (PI / 2.0e0);
+const double TWO_PI = (2.0 * PI);
+const double DEG_TO_RAD = (PI / 180.0);
+const double RAD_TO_DEG = (180.0 / PI);
+const double AD_C = 1.0026000;            /* Toms region 1 constant */
```
mleotta: These last two, `AD_C` and `COS_67P5` are custom for this code and probably don't belong in `math_constants.h`.  However, they are only used in one function, so I would just make them local to that function.

vital/types/local_cartesian.cxx
```diff
@@ -0,0 +1,524 @@
...
+  {
+    vector_3d geocentric_coordinate;
+    convert_to_geocentric(cartesian_coordinate, geocentric_coordinate);
+
+    double X = geocentric_coordinate.x();
+    double Y = geocentric_coordinate.y();
+    double Z = geocentric_coordinate.z();
+    double latitude, longitude, height;
+
+    if (Iterative_Geocent_algorithm)
```
mleotta: What does this compile-time switch do?  Should we just remove the code that is not being used?  Or maybe make it a run time option?  It looks like there are two algorithm alternatives, do we know the pros and cons of each?

vital/tests/test_local_cartesian.cxx
```diff
@@ -0,0 +1,129 @@
...
+{
+  kwiver::vital::plugin_manager::instance().load_all_plugins();
+
+  kwiver::vital::geo_point geo_out;
+  kwiver::vital::vector_3d cart_out;
+  kwiver::vital::local_cartesian lc_lla(origA);
+
+  // Get the geopoint from an offset
+  lc_lla.convert_from_cartesian(offset1, geo_out);
+  auto const epsilon_geo_out_to_geo1 = (geo_out.location() - geo1.location()).norm();
```
mleotta: consider `lpNorm<Infinity>()` instead of `norm()`.  That gives you the maximum absolution value of the elements.  The L2 distance doesn't make as much sense here.  Also, you really just want to make sure the error is low in all elements, so checking the max is probably the right thing to do.

vital/tests/test_local_cartesian.cxx
```diff
@@ -0,0 +1,129 @@
...
+  EXPECT_MATRIX_NEAR(geo_out.location(), geo1.location(), 1e-3);
```
mleotta: This error tolerance seems quite high.  I think 1e-3 in lat/long is about 100 meters.  Is this conversion not more accurate than that?

vital/types/local_cartesian.h
aaron-bray: I am not sure the motivation for this.
This is pretty old NGA code that I am just porting into kwiver.
I never used the orientation when I used this, but the code does support it if needed.
Not sure what use case they were coding for

vital/types/local_cartesian.cxx
linus-sherrill: These could be `constexpr` values.

vital/tests/test_local_cartesian.cxx
aaron-bray: Note the matrix compare eigen is doing is not based on the decimal places (not sure how it does this), but you can see the accuracy of the conversions are pretty good down to at least the 6th decimal place

geo_out.location() evaluates to   -73.758985147971799
   42.850126094497369
0.0002865474671125412
geo1.location() evaluates to -73.758985150000001
 42.850126090000003
                  0
1e-4 evaluates to 0.0001

vital/types/local_cartesian.cxx
aaron-bray: Well, it looks like they released geotrans 3.8 this year, but I need certs to download it...
It looks like they added spherical coordinate support and fixed a but we may want to get (if I can get the source code)

https://earth-info.nga.mil/GandG/update/wgs84/apps/geotrans/current-version/docs/MSP_Geotrans_Release_Notes%203.8.pdf

vital/tests/test_local_cartesian.cxx
mleotta: here are the docs:
https://eigen.tuxfamily.org/dox/group__TutorialReductionsVisitorsBroadcasting.html

I would not use `norm()` here because the units on altitude are very different from those on lat and long

vital/tests/test_local_cartesian.cxx
mleotta: The latitude and longitude values need to be a lot more precise than altitude.  I would not score these against the same epsilon.  1e-4 is sub-millimeter in height but 10s of meters in horizontal position.

vital/types/local_cartesian.h
mleotta: that's fine, we can leave it as is.

vital/tests/test_local_cartesian.cxx
aaron-bray: I just switched the test to test each component individually, seems more flexible and straight forward

vital/tests/test_local_cartesian.cxx
aaron-bray: I just switched the test to test each component individually, seems more flexible and straight forward

aaron-bray:
@mleotta  I Think this is ready to go. I updated the test to compare individual components, and removed the legacy geocentric-to-geodetic algorithm. I did not update the code styling just so the code base is easily compatible to the original geotrans code just in case. I originally did a few changes to get things into a class with better functions, but for the most part, I tried to preserve styling to the original.

PR #857, opened 2019-09-25
Fix memory leaks in ffmpeg arrow

as6520:


arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -339,7 +339,7 @@ class ffmpeg_video_input::priv
       avformat_close_input(&this->f_format_context);
       this->f_format_context = nullptr;
     }
-
+    avcodec_close(this->f_video_encoding);
```
mleotta: This looks like the right thing to do, and the primary culprit of the memory leak.  However, for added safety and consistency for the other similar statements in this function, can you change it to
```c++
if (this->f_video_encoding)
{
  avcodec_close(this->f_video_encoding);
  this->f_video_encoding = nullptr;
}
```

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -300,7 +299,8 @@ class ffmpeg_video_input::priv
     }
     this->frame_advanced = 0;
     this->f_frame->data[0] = NULL;
-
+    avcodec_close(codec_context_origin);
```
mleotta: I'm not as convinced that this need. `codec_context_origin` was not allocated by us.  It is a pointer to codec memory that was already allocated elsewhere.  I would think this would be closed/deleted automatically when we clean up `this->f_format_context`.  Can you confirm that this line makes a difference in valgrind?

as6520: Just checked it does not. I think my main motivation there was to find all codec contexts and close them explicitly. Will remove the close

dstoup:
Jenkins test this please

dstoup:
That CentOS error is the same random one we've seen before.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -285,7 +285,6 @@ class ffmpeg_video_input::priv
         this->f_start_time = -1;
         return false;
     }
-
```
mleotta: Not critical, but I would undo the whitespace changes in this function.  They aren't relevant to this branch and don't make any more compliant with style guidelines

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -331,16 +329,19 @@ class ffmpeg_video_input::priv
 
     if (this->f_video_stream)
     {
-      avcodec_close(this->f_video_stream ->codec);
```
mleotta: This whitespace change is good.  I'm not sure where that space came from.

linus-sherrill:
There is a conflict in the release notes that needs to be addressed before merge

PR #862, opened 2019-10-01
Dev/vital wheel [WIP]

as6520:
This is the first attempt to create a wheel for vital. It supports the following
1. Registering python and c++ algorithm from an external python package
2. Vital python bindings under kwiver module.

### Aim
The aim of this PR is to discuss the build and registration process for the wheel, document caveats encountered and get feedback about improving either processes. 

### Build Process
We are using [scikit-build](https://scikit-build.readthedocs.io/en/latest/#) to drive the build. The wheel is built using the cmake workflow. Static builds of Kwiver and Fletch were used for the most part to limit the linking of libraries in the package. 

#### Fletch Dependencies
1. Boost
2. Eigen
3. Log4CPlus
4. pybind11

#### Generating Kwiver Wheel
``
python setup.py bdist_wheel -- -Dfletch_DIR=<path to static fletch build> -- -j8
``
#### Major Changes
1. Hardcoded [lib](https://github.com/Kitware/kwiver/blob/master/CMake/utils/kwiver-utils-targets.cmake#L195) has been replaced with library_dir variable with default value of lib. kwiver_add_python_library uses this change to install python bindings in `package_root`
2. Headers, Exported targets and setup scripts are not installed when skbuild is driving the build.
3. The vital bindings have been moved into kwiver directory.
4. The tests have been taken out of vital and moved into the kwiver directory.
 
#### Build Process Caveats
libvital_vpm, libvital_algo and libvital_config, kwiver plugins and the python bindings are build as shared libraries. Among the aforementioned components,  libvital_vpm, libvital_algo and libvital_config can create issues as they are not leaf modules in the build. These libraries can't be build statically as the plugin_manager in libvital_vpm implements singleton pattern using static variable. In the static build, a static variable can have multiple copies as discussed in this [issue](https://stackoverflow.com/questions/13869809/multiple-copies-of-static-members-when-mixing-shared-static-libs). Thus the singleton pattern does not hold and the plugins are not properly registered. 
Any shared library in kwiver can use rpaths with respect to $ORIGIN to link against these libraries. However a library outside the package can't specify the rpath at the moment. Until this issue is rectified, a library that requires libvital_vpm, libvital_config and libvital_algo must be present in the Kwiver wheel. 

#### Python Algorithm Registration
Kwiver exposes `kwiver.python_plugin_registration` entrypoint for external python packages to register python algorithms. An external package can pass the module path with the registration function through the entrypoint to register the algorithm. An example demonstrating the functionality is available [here](https://github.com/as6520/external_arrow/blob/master/setup.py#L29).

#### C++ algorithm registration
Kwiver exposes `kwiver.cpp_search_paths` entrypoint for external python packages to register c++ algorithms. We add additional search paths to plugin manager to register c++ algorithm. However,  due to entrypoint specification, the path must be returned by a function rather than being added directly to the entrypoint. The example is available [here](https://github.com/as6520/external_arrow/blob/master/setup.py#L33).

 ### Wheels
The kwiver wheel for python-3.5 is available [here](https://data.kitware.com/#item/5d938fe2d35580e6dc5881f5) along with the README for installation and testing


as6520:
Superseded by #924

PR #868, opened 2019-10-02
Dev/remove ctypes python bindings

as6520:
Remove the older ctypes bindings from python bindings since we are using pybind11 for python bindings


vital/bindings/python/Kwiver_Python_Demo.ipynb
```diff
@@ -1,606 +0,0 @@
-{
```
Purg: I don't remember exactly what was in this notebook, but it might be worth while adapting what may have previously been in here to the current bindings as a quick-start or otherwise example usage of something.

as6520: It was an introduction to c based python bindings and algorithms

as6520: Noted in #878 and would be addressed once kwiver wheel is complete

Erotemic:
Pending ghostflow this LGTM

linus-sherrill:
Jenkins, test this please

linus-sherrill:
Will this cause any problems with external projects?

as6520:
I don't think so, ctypes based bindings were the predecessor of pybind11 bindings.

PR #875, opened 2019-10-07
Refactor transform functions

mleotta:
This branch refactors the `core` transform functions with the following changes
- Make the order of transform arguments consistent (xform always last)
- Add a few more helper transform functions for more specific data types
- Fix the namespace of these functions. Moved from `arrows` into `arrows::core`
- Move the Necker reverse function into a new `necker_reverse.h` header.
- Refactored the Necker reverse function into its components so that each can be called directly

There are no changes to any algorithms on this branch, just moving code around and exposing more APIs.

### DIFF
--- a/vital/types/camera_perspective_map.h
+++ b/vital/types/camera_perspective_map.h
@@ -1,0 +1,57 @@
+/*ckwg +29
+ * Copyright 2019 by Kitware, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  * Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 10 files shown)


vital/types/camera_perspective_map.h
```diff
@@ -0,0 +1,57 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * \file
+ * \brief Typedefs for camera_perspective camera_maps
+ */
+
+#ifndef VITAL_CAMERA_PERSPECTIVE_MAP_H_
```
mleotta: Note: this file was also added in #874, but it was also need here.  If we merge #874 first then I can rebase to remove this change.

mleotta:
@linus-sherrill, I'm trying to get this refactoring branch landed.  Any objections?

PR #876, opened 2019-10-07
fix config name

mleotta:
The wrong string name was used to access configuration of this nested algorithm.


mleotta:
@Cookt2 @dstoup @linus-sherrill 

This is a very trival one-line bug fix that has been sitting here for weeks. Can someone merge this to release and merge release into master?

PR #848, opened 2019-09-19
Dev/move config bindings

as6520:
Move the config bindings to vital since that is where the cpp code is

PR #877, opened 2019-10-09
Dev/create py kwiver

as6520:
1. Refactor python bindings, python related utilities and python tests out of vital into a directory at project root called python
2. Refactor python bindings, python related utilities and python tests out of sprokit into a directory at project root called python
3. Create root directory kwiver for all python bindings associated with kwiver
4. Modify python cmake utils to include project name as module root
5. Move sprokit python schedulers to kwiver.sprokit.schedulers
6. Move sprokit python processes to kwiver.sprokit.processes

This PR depends on #848, #864 and #868

This PR is the first step towards creating a wheel for kwiver with python/kwiver acting as the root package

Note: This is a breaking change for kwiver python bindings used in any project since the import statements used by those projects would have to change.


linus-sherrill:
What is the benefit of this change?

as6520:
The intent of this branch is source code organization for the kwiver wheel.  The code reorganization would allow python directory in the project root to be the package directory for kwiver wheel. Additionally the reorganization ensures that we can leverage the existing build process without significant changes in it to create the wheel. 

The kwiver wheel would introduce a top level module name kwiver for python bindings since we have a single python package. The changes in this branch ensure that the current build process is following the same pattern and we can see the effect of those change independent of the wheel building process and wheel related changes. 

These changes also consolidate the python bindings and python tests for sprokit under a single directory.

as6520:
Closing since the changes have been merged in dev/overhaul branch

PR #879, opened 2019-10-10
Add IOU-based tracker with homography support

neal-siekierski:
This PR adds a (Python) process `simple_homog_tracker` that performs IOU-based tracking using homographies to place bounding boxes in the same coordinate system.

Follow-up concerns:
- This conflicts with #877
- ~Related to reconciling this with #877, where should the source file be and where should it be installed? I currently have the source file in `arrows/python` but it gets installed in `kwiver.processes`. (This somewhat parallels `srnn_tracker.py` on the `viame/master` branch, but doesn't mean it's a great solution.)~
  - The source file is now in `sprokit/processes/python`, as suggested by https://github.com/Kitware/kwiver/pull/879#issuecomment-540716179
- I've mostly used Numpy and custom wrapper types internally, only converting to/from KWIVER types at the end. Perhaps this could be better written with existing types and methods? (This would be easiest if direct equivalents, vectorization and all, are available)
- ~Grabbing input using the `homography_src_to_ref` trait returns a `Datum`, not a `F2FHomography`, perhaps due to something I missed in #842.~
  - The underlying issue is fixed in #880 and I've removed the work-around here.
- There are no tests. What is the recommended way of testing a process? (Alternatively, doing something like testing the `track` function would also be possible.)

@mattdawkins: Anyone else to tag for review?

### DIFF
--- a/sprokit/processes/python/CMakeLists.txt
+++ b/sprokit/processes/python/CMakeLists.txt
@@ -14,6 +14,10 @@ kwiver_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/homography_writer.py
   kwiver/processes
   homography_writer)
 
+kwiver_add_python_module( ${CMAKE_CURRENT_SOURCE_DIR}/simple_homog_tracker.py
+  kwiver/processes
+  simple_homog_tracker)
+
 kwiver_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/ProcessImage.py
   kwiver/processes
   ProcessImage)
--- a/sprokit/processes/python/simple_homog_tracker.py
+++ b/sprokit/processes/python/simple_homog_tracker.py
@@ -1,0 +1,352 @@
+# ckwg +29
+# Copyright 2019 by Kitware, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice,
+#    this list of conditions and the following disclaimer.
+#
+#  * Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+#  * Neither name of Kitware, Inc. nor the names of any contributors may be used
+#    to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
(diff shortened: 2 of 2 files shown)


as6520:
On the current master, sprokit/processes/python should be the place to where python processes reside.   They are built and installed in kwiver/processes. With #877 the processes would move to python/kwiver/sprokit/processes and would be installed in kwiver/sprokit/processes. 

A new bindings needs to be added as a port trait in [KwiverProcess](https://github.com/Kitware/kwiver/blob/master/sprokit/processes/bindings/python/kwiver/kwiver_process.py) and you would need  to convert datum to and from a particular type by adding couple of methods in the datum bindings as shown [here](https://github.com/Kitware/kwiver/blob/master/sprokit/src/bindings/python/sprokit/pipeline/datum.cxx#L108) 

As for testing are we talking about testing this process or testing homography bindings?

neal-siekierski:
> On the current master, sprokit/processes/python should be the place to where python processes reside. They are built and installed in kwiver/processes. With #877 the processes would move to python/kwiver/sprokit/processes and would be installed in kwiver/sprokit/processes.

Sounds good; I'll move the process to `sprokit/processes/python` for the time being.

> A new bindings needs to be added as a port trait in [KwiverProcess](https://github.com/Kitware/kwiver/blob/master/sprokit/processes/bindings/python/kwiver/kwiver_process.py) and you would need to convert datum to and from a particular type by adding couple of methods in the datum bindings as shown [here](https://github.com/Kitware/kwiver/blob/master/sprokit/src/bindings/python/sprokit/pipeline/datum.cxx#L108)

Ah, I see. I added the `get_*` method but didn't do the other things. I think it makes sense to do that as a separate PR.

> As for testing are we talking about testing this process or testing homography bindings?

Testing the process. I think the remaining homography issues would be better addressed separately.

How far do you think #877 is from landing? It might make sense to hold off on the additional homography PR (and maybe even this one) until then.

as6520:
> How far do you think #877 is from landing? It might make sense to hold off on the additional homography PR (and maybe even this one) until then.

Not sure, although most of the changes in #877 are related to reorganization they are still breaking changes. Additionally the PR's that it has been conditioned on #848 and #868 have to landed yet. 

I would recommend continuing the development and I will periodically rebase on master to fix conflicts from python dev stuff.

mattdawkins:
I've tested this code and delivered to a customer, will merge in 2 days unless I hear any feedback

PR #884, opened 2019-10-11
Add occluded space constant in TSDF

mleotta:


bastienjacquet:
Conceptually, this hard to me to get proper ground or modeling framework for doing so. You really don't know anything about what happen beyond your depth measurement (in contrast to in front of it, where it has to be somehow empty, up to voxelisation-vs-ray artefacts).
The prior on "object thickness" is usually taken car of by Delta.
Usually, you handle a emptiness or fullness prior by pre-assigning a constant value to all voxels.

The double shell effect is actually as plausible as anything else (some walls are actually empty inside, right ;) ).
You might want to extract the occupied/empty values not as a raw extraction, but with an optimisation enforcing some smoothness, and have a heavier smoothness prior.

mleotta:
I considered a pre-assigning a constant to all voxels as a fullness prior.  That is sort of what I want, but not quite because in part of the scene the prior should be "empty" and in other part of the scene the prior should be "full".  I don't know which parts are which a priori.  The parts that are always occluded should have a "full" prior but the parts that are never observed (no rays pass through) should have an empty prior.

What I am trying to avoid is creating boundaries between unobserved voxels and observed voxels with value either positive or negative function values.  There is no reason to believe that the boundary between freespace I have observed and freespace just outside my camera frustums should be a surface.  Likewise, there is no reason to believe that there should be a boundary between areas that I've marked as occupied (close to a surface) and areas that are occluded in all views.  Adding a very small positive value beyond delta allows me to set a fullness prior in the occluded areas, but allows that prior to be easily overridden by a single observation that sees free space.

My particular application would prefer solid objects over thin shells.  Is there any reason not to make this an option?  It can be easily disabled by setting epsilon to zero to get the old behavior.

bastienjacquet:
My intuition is that pre-assigning to all voxels a constant equal to the same `epsilon * \rho` should have the same effect, as it is a very small positive value that allows that prior to be easily overridden by a single observation that sees free space.
But I fully hear the "observed-free" space in the camera frustum vs unobserved space just outside of the camera frustum.
The volume filled-in with those ray potential is used a unary potential of a TV-L1 regularization that penalize the surface area. See e.g C. Häne & C. Zach line of work, that does it by primal/dual optimization. This gets rid of weakly supported boundaries.
Extracting the 0-isosurface is just a no-binary/no-smoothness shortcut we did.

In practice, since it keeps the old behavior accessible, lets do it. But maybe comment that it's not suggested by the paper authors, and is a recipee that "just works" as a tradeoff between:
a slight emptyness prior that create shelling effect of objects bigger than Delta.
a slight fullness prior that creates occupied-voxels/"objects" also in unobserved area outside of the frustum, in particular near from the camera center.

mleotta:
Yes, the pre-assigned constant should have the same effect except in the boundaries between observed and unobserved freespace, which is a critical distinction.  I get a surface between zero and positive values, but not between zero and negative values.  So as soon as I add the small positive constant I get surfaces along the camera frustums boundaries.

Is there a particular paper that we should be citing that implements the TSDF with these specific parameters?  I can add a comment about deviating from what is described in a paper, but I'm not sure which paper this is based on.

bastienjacquet:
Sure, here is the most related paper related to cuda depthmap integration followed by optimization:
Fast and High Quality Fusion of Depth Maps. C. Zach. Proceedings of 3DPVT'08
https://www.cc.gatech.edu/conferences/3DPVT08/Program/Papers/paper196.pdf

Notice that even as early as in "A volumetric method for building complex models from range images". B Curless and M Levoy 96. they did handle 3 voxel states (empty, unseen, full), and threated empty-vs-full different than empty-vs-unseen.
To me the "good" behavior is the sligh emptiness prior that create shell effect, adjusted with a choice of Delta. It reflects the prior you have on object scale, and on the actual scale you chose to model the object. You might want a house to be full or an empty shell, depending on your expected modeling scale.

arrows/cuda/integrate_depth_maps.cu
```diff
@@ -126,7 +128,8 @@ __device__ void rayPotential(double realDistance, double depthMapDistance, doubl
   int sign = diff != 0 ? diff / absoluteDiff : 0;
 
   if (absoluteDiff > c_rayPotentialDelta)
-    res = diff > 0 ? 0 : - c_rayPotentialEta * c_rayPotentialRho;
+    res = diff > 0 ? c_rayPotentialEpsilon * c_rayPotentialRho
+                   : - c_rayPotentialEta * c_rayPotentialRho;
```
bastienjacquet: The two lines could read
```suggestion
    res = sign * c_rayPotentialEpsilon * c_rayPotentialRho;
```

mleotta: I don't think this is quite right because in addition to the sign of the value we are also switching between using Epsilon for positive values and Eta for negative values.

bastienjacquet: My bad.

PR #885, opened 2019-10-15
Change frame_list_input to frame_list_process

as6520:


dstoup:
Seems like a reasonable change since it is a process. At the risk of being too wordy, it seems like frame_list_input_process might be more descriptive. but I'm not that concerned either way. Will wait for anyone else to chime in.

linus-sherrill:
The name of a process plugin should not contain "process", however, the file name can contain "process". Having "process" in the process plugin name is redundant". The only process that has "process"  as part of the plugin name is frame_list_process. Maybe that should be changed.

PR #900, opened 2019-11-05
Add filter function to DOS

mattdawkins:


vital/types/detected_object_set.h
```diff
@@ -220,6 +220,26 @@ class VITAL_EXPORT detected_object_set
...
 
+  /**
+   * @brief Filter detections via an arbitrary function
+   *
+   * Filter the detected object set based on some predicate function
+   *
+   * @param p Predicate function which returns true if an element should
+   *          be removed.
+   *
+   * @return List of detections.
```
linus-sherrill: It looks like this method returns VOID not a list of detections.
It may be helpful to indicate that the filter exits the current set of detections.

linus-sherrill:
Applying a filter to a DOS is not really a property of a DOS, but an operation on a DOS; therefore, this should be implemented as a function and could go in the `detected_object_set_util.*` files. Other operators from the DOS are migrating out to these files.

PR #904, opened 2019-11-26
Callbacks in loggers

mleotta:
I'd like to get initial impression on this attempt to add callbacks to the logger framework.  So far this is only implemented for the default and log4cplus loggers.  Each logger instance can register a callback, but there is also a global callback that will apply to all loggers.  The global callback is the one I am most interested in.  My use case is capturing log messages to display in a GUI window.

There is currently a difference in approach between the default and log4cplus loggers treatment of global callbacks.  The default logger overrides the local callback for each logger when a global callback is set.  So there is always only one callback.  The log4cplus logger supports both local and global callbacks at the same time.  I'm not sure which approach is better.

log4cplus has it's own callback mechanism which could be used, but does not appear to be supported until a later version of log4cplus than we have in Fletch.

### DIFF
--- a/doc/release-notes/master.txt
+++ b/doc/release-notes/master.txt
@@ -20,6 +20,9 @@ Vital
  * Added a variant of config_block::get_enum_value that accepts a default
    value to fall back on.
 
+ * Added support for registering callbacks to the KWIVER logger.  Callbacks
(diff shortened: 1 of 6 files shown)


mleotta:
@linus-sherrill I've addressed the issues we've discussed.  This branch should be complete now.  Please review.

PR #905, opened 2019-12-02
Change test for boolean options.

linus-sherrill:
When BOOST regex is used on CentOS, the test for boolean options
were failing due to possible undefined behaviour.

For boost::regex_match() - If the function regex_match() returns false, then the effect on parameter match_results m is undefined.

For std::regex_match() - If the recex_match() returns false, then the effect on parameter match_results m is as follows:
m.ready() == true; m.empty() == true; m.size() == 0


dstoup:
Looks good to me. I suspect that Mac dashboard error is not related to this issue. Rerunning.

dstoup:
Jenkins test this please

PR #906, opened 2019-12-04
Update KPF reader to provide empty detected object sets from frame 1

dmjoy:
Assume the first frame of the source of possible KPF detections is 1,
rather than whenever the first detection is reported.

When syncing the KPF detections with video for example, we want to know what (if any) were the detections starting from frame 1.  Currently, if the first detection in the KPF input file occurs on frame 10, it will be associated with the first frame of the video.

It's very possible that I've missed an option somewhere or another processes that's intended to correct for this.


linus-sherrill:
Nudge - any comments on this branch?

linus-sherrill:
Nudge

PR #912, opened 2019-12-24
Refactor detected_object_set-move operations out of class

linus-sherrill:
The scale and translate methods are not inherently operate on the collection, but on each individual element and are not properties of the set. These methods were converted into unbound functions to reduce method proliferation in the detected_object_set class.

The existing methods are marked as DEPRECATED and will be removed at a later time.

- moved scale and translate from class to stand-alone operations
- Updated. improved and corrected documentation
- Added const iterators (cbegin, cend) to some collections
- Fixed pointer dereference bug

### DIFF
--- a/doc/release-notes/master.txt
+++ b/doc/release-notes/master.txt
@@ -45,6 +45,9 @@ Vital Types
  * Added iterator, const_iterator and iterable classes to support container-
    independent iteration.
 
+ * Refactored detected_object_set to convert some methods to external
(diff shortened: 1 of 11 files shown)


linus-sherrill:
Jenkins, test this please

linus-sherrill:
Jenkins, test this please

linus-sherrill:
Jenkins, test this please

PR #921, opened 2020-01-24
Add support for finalize stage of process life cycle.

linus-sherrill:
The finalize method is called when the process has determined that
there are no more inputs and it should terminate. This event is useful
so a process can gracefully terminate while the pipeline is still active
without having to wait for the DTOR (which may be too late).


mattdawkins:
Can we not have camel case in the folder name RightTrack?

one of my biggest issues, on a related note, in kwiver style is KWSys using it because it doesn't mesh with the snake case everywhere else

linus-sherrill:
Ready for review

linus-sherrill:
Jenkins, test this please

Issue #547, opened 2018-06-21
Detection writers writing to subdirectory doesn't complain when subdir doesn't exist


mattdawkins:
If the detection writers are writing to a directory called 'output' and 'output' doesn't exist, no tracks/detections get written out but the software doesn't error, leaving the user to not understand why tracks don't exist. The software should either:

(1) Print an error when the directory doesn't exist
or
(2) Make the directory

PR #922, opened 2020-01-27
Improve file management in various I/O algos

neal-siekierski:
These changes improve the file management in several readers and writers in `vital/algo` so that:
- An exception is thrown when the file fails to properly open
- Memory is not leaked when that exception is thrown
- *(Edit 2020-01-28):* All exceptions are thrown using `VITAL_THROW`

#547 was likely caused by the algos' previous state, so this PR should fix it.

Besides the changes made, I'll note that all the code in the `open` and `close` methods here is redundant between the various algos, so it would be good to factor it out, in particular to create a class to manage the ownership and get it right once using RAII, since errors remain in the ownership code. (`close` should be called in destructors but isn't consistently, and calling `open` twice without calling `close` in the middle results in a leaked stream in some of these classes. `close` should be the destructor of a maybe-owned smart pointer.) I don't know where such code would best be placed, though.

### DIFF
--- a/vital/algo/detected_object_set_input.cxx
+++ b/vital/algo/detected_object_set_input.cxx
@@ -93,7 +93,7 @@ detected_object_set_input
   std::unique_ptr< std::istream > file( new std::ifstream( filename ) );
   if ( ! *file )
   {
-    throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
+    VITAL_THROW( file_not_found_exception, filename, "open failed" );
   }
 
   m_stream = file.release();
--- a/vital/algo/detected_object_set_input.cxx
+++ b/vital/algo/detected_object_set_input.cxx
@@ -35,6 +35,8 @@
 
 #include "detected_object_set_input.h"
 
+#include <memory>
+
 #include <vital/algo/algorithm.txx>
 #include <vital/exceptions/io.h>
 #include <vital/vital_types.h>
@@ -88,13 +90,13 @@ detected_object_set_input
   }
 
   // try to open the file
-  std::istream* file( new std::ifstream( filename ) );
+  std::unique_ptr< std::istream > file( new std::ifstream( filename ) );
   if ( ! *file )
   {
     throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
   }
 
-  m_stream = file;
+  m_stream = file.release();
   m_stream_owned = true;
 
   new_stream();
--- a/vital/algo/detected_object_set_input.cxx
+++ b/vital/algo/detected_object_set_input.cxx
@@ -89,7 +89,7 @@ detected_object_set_input
 
   // try to open the file
   std::istream* file( new std::ifstream( filename ) );
-  if ( ! file )
+  if ( ! *file )
   {
     throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
   }
--- a/vital/algo/detected_object_set_output.cxx
+++ b/vital/algo/detected_object_set_output.cxx
@@ -72,7 +72,7 @@ detected_object_set_output
 {
   // try to open the file
   std::ostream* file( new std::ofstream( filename ) );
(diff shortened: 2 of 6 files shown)


vital/algo/detected_object_set_input.cxx
```diff
@@ -88,13 +90,13 @@ ::open( std::string const& filename )
   }
 
   // try to open the file
-  std::istream* file( new std::ifstream( filename ) );
-  if ( ! file )
+  std::unique_ptr< std::istream > file( new std::ifstream( filename ) );
+  if ( ! *file )
   {
     throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
```
linus-sherrill: Just a minor comment - This could be `VITAL_THROW` .

vital/algo/detected_object_set_output.cxx
```diff
@@ -71,13 +73,13 @@ detected_object_set_output
 ::open( std::string const& filename )
 {
   // try to open the file
-  std::ostream* file( new std::ofstream( filename ) );
-  if ( ! file )
+  std::unique_ptr< std::ostream > file( new std::ofstream( filename ) );
+  if ( ! *file )
   {
     throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
```
linus-sherrill: Could be` VITAL_THROW` for better diagnostics

vital/algo/read_object_track_set.cxx
```diff
@@ -95,14 +97,14 @@ ::open( std::string const& filename )
...
 
   // try to open the file
-  std::istream* file( new std::ifstream( filename ) );
+  std::unique_ptr< std::istream > file( new std::ifstream( filename ) );
 
-  if( ! file )
+  if( ! *file )
   {
-    kwiver::vital::file_not_found_exception( filename, "open failed"  );
+    throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
```
linus-sherrill: Also could be `VITAL_THROW`

vital/algo/write_object_track_set.cxx
```diff
@@ -71,14 +73,14 @@ write_object_track_set
...
 {
   // try to open the file
-  std::ostream* file( new std::ofstream( filename ) );
+  std::unique_ptr< std::ostream > file( new std::ofstream( filename ) );
 
-  if( ! file )
+  if( ! *file )
   {
-    kwiver::vital::file_not_found_exception( filename, "open failed"  );
+    throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
```
linus-sherrill: `VITAL_THROW` as above

vital/algo/write_track_descriptor_set.cxx
```diff
@@ -71,14 +73,14 @@ write_track_descriptor_set
 ::open( std::string const& filename )
 {
   // try to open the file
-  std::ostream* file( new std::ofstream( filename ) );
+  std::unique_ptr< std::ostream > file( new std::ofstream( filename ) );
 
-  if( ! file )
+  if( ! *file )
   {
     throw kwiver::vital::file_not_found_exception( filename, "open failed"  );
```
linus-sherrill: As above

neal-siekierski:
One of the builds failed in testing with:
```
The following tests FAILED:
	597 - test-python-run-python_to_python-pythread_per_process (SEGFAULT)
```
I'm restarting it as it's likely intermittent and unrelated to the current changes.

vital/algo/detected_object_set_input.cxx
neal-siekierski: These should all be converted to use `VITAL_THROW` now.

linus-sherrill:
Jenkins, test this please

mattdawkins:
works for me after testing, though one might question whether or not file streams should be handled in the base impl as opposed to just the derived classes, but that's out of scope I suppose

PR #923, opened 2020-01-29
Add track::clear method

mwoehlke-kitware:
Add a method to vital::track to remove all history.


mwoehlke-kitware:
Eek! Good catch! Should be fixed now.

PR #924, opened 2020-02-03
Python Overhaul + Activity Detector [WIP]

as6520:

PR #925, opened 2020-02-03
Python Overhaul + Activity Detector [WIP] - V2

Erotemic:
This is a rebased version of #924 

I'm leaving  #924  as it is until I ensure I'm not messing anything up.


arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -845,116 +811,7 @@ kwiver::vital::image_container_sptr
 ffmpeg_video_input
 ::frame_image( )
 {
-  // Quick return if the stream isn't valid
```
mleotta: I'm not sure what all these major FFMPEG changes are doing, but I would prefer these be reviewed as a separate branch.  FFMPEG is a key functionality that is tricky to get right.  Changes could introduce errors across many projects.  It would be better to isolate these changes for separate review and explain the motivation behind them.

as6520: #927 has been added with the isolated changes

dstoup:
Jenkins test this please

arrows/ffmpeg/ffmpeg_video_input.cxx
as6520: @mleotta  As were are addressing the concerns FFMPEG and now ocv image container changes in #927 are there an other concerns about this branch?

as6520:
Jenkins test this please

dstoup:
Jenkins test this please

as6520:
Jenkins test this please

as6520:
Jenkins test this please

as6520:
jenkins test this please

as6520:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

python/kwiver/vital/modules/module_helpers.cxx
```diff
@@ -114,27 +76,31 @@ register_factories(kwiver::vital::plugin_loader& vpm)
...
 
-  _load_python_library_symbols();
-
+/*
+ * check if the python shared library symbols have been loaded
+ * From: https://stackoverflow.com/questions/50570223/dlopenrtld-noload-still-returns-not-null-after-dlclose
+ */
+bool is_python_library_loaded(const std::string& python_library_path)
+{
+  void *handle = dlopen(python_library_path.c_str(), RTLD_NOW | RTLD_NOLOAD);
```
dstoup, line 87: I mentioned it over email but the call to dlopen will fail when Linux header dlfcn.h isn't there, e.g. Windows. You should use kwiversys' DynamicLoader instead.

as6520:
Jenkins test this please

Cookt2:
Jenkins test this please

Cookt2:
Jenkins test this please

as6520:
Jenkins test this please

as6520:
Jenkins test this please

CMake/kwiver-flags-gnu.cmake
```diff
@@ -116,3 +116,9 @@ if (KWIVER_CPP_COVERAGE   AND   CMAKE_BUILD_TYPE EQUAL "DEBUG")
   # It seems as though the flag isn't detected alone.
   kwiver_check_compiler_flag(-fprofile-arcs)
 endif ()
+
+# GCC Flag used for stripping binaries of any unused symbols
+# Used for reducing the size of kwiver wheel since pypi has 60Mb constraint on wheel size
```
Erotemic, line 121: This comment is very helpful. So is the one for zlib. It helps me know what its doing and why I want to be doing it.

as6520: Yes, I agree that it does help with the explanation. Originally I was using commit messages or more detailed descriptions to document the caveats. But I did drop the ball on these two changes.

Erotemic: While descriptive git messages are nice, my opinion is that in-code comments are better because they are more tightly coupled with the code. The git history can be unintentionally overwritten and its not always intuitive that an explanation for a cryptic line of code might be in the git history. However, in-code comments are much harder to miss and unintentionally remove. 

That being said, I'm certainly no paragon of [good git commit messages](https://github.com/Erotemic/ibeis/commits/master).

dstoup:
Jenkins test this please

tao558:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
This has been rebased on top of the latest master and is a complete fast-forward merge. I added 1 commit in order for this branch to work with #931. In case this rebase is not correct, I have a copy of the previous state of the branch on my local machine as well as on my fork of kwiver.

There are still some commits that could be squashed here to give a cleaner git history, but I wanted to push this version now to test the CI.

tao558:
Jenkins test this please

Erotemic:
Can we remove track_oracle/tests/data/2018-03-07.16-55-06.17-00-06.school.G336.geom.yml from commit cbe67bb? If we can't remove it can we change this to use CMake's [ExternalData](https://cmake.org/cmake/help/v3.3/module/ExternalData.html#id1) feature to download it on demand instead if keeping it in the repo?

collinsr:
If the issue is the size of the data, another option would be to shrink these data files to be exemplars rather than copies of the original problematic file. I don't want to remove them outright, since the problem they check against is real.

Erotemic:
@collinsr I think it would make sense to shrink these files. I saw the email you sent containing the smaller versions  (a reduction from 7.2MB to 33KB should make ghostflow happy). 

I also wonder if it would be also be a good change such that this test data was hosted externally and downloaded on demand (which smaller sizes still help with). I generally like the practice of hosting test data external to the repo itself (e.g. https://github.com/Kitware/kwimage/blob/master/kwimage/im_demodata.py#L8). IMO It keeps the repo cleaner. However, it does mean that some tests will now require a network connection to pass (which I usually work around by setting a flag that explicitly enables network tests). These files are now small enough where I wouldn't complain either way, in some sense it is easier to just keep them in the repo, but does anyone else have opinions on this?

collinsr:
I have not given much thought to moving test data out of the repo. On the
one hand, I would guess test data tends to be more static and can certainly
get bulky. If we revive the idea of end-to-end regression tests (aka the
old midas-based "tracker tracker"), that sort of data would certainly not
live in the repo (especially for the ITAR tests.)

On the other hand, you end up with a parallel infrastructure to maintain;
if I have a branch which changes the test code, where does the matching
branch for the test data live? How does that get managed? Plus this opens
an ownership question: in the link you gave, we've ceded our ability to
test our code to the whims of imgur and wikimedia. That makes me
uncomfortable.

I would also be uncomfortable dividing our tests into those we can run in a
classified envrionment, and those we can't.

One solution might be to essentially set up a resolver for test data.
Rather than have a explicit URI for test data, have a symbolic name backed
up by a source URI and some lightweight infrastructure which can either
fetch it from a local cache or straight from the source if you're in an
environment where bandwidth is cheap. The same infrastructure would allow
building the local cache and a verification capability to keep an eye out
for when the data changes / becomes unavailable / other weirdnesses.

tao558:
Moved to #1013

Erotemic:
It should be noted that the "grabdata" pattern that I use does do the lightweight local caching to the "correct" OS-specific application cache directory powered by [ubelt.grabdata](https://ubelt.readthedocs.io/en/latest/ubelt.html#ubelt.grabdata), it also checks that the data matches the expected hash, so you are all-but-guaranteed that the data returned is the expected data. It does use stamp files to make subsequent runs very fast, so there is a chance of failure of a adversarial actor messes with those stamp files. 

For more robust hosting we can use something like data.kitware.com. Obviously any URI based solution will have a chance of failure, but the same is true for even building fletch. Girder actually provides a nice way to host data and reference it. Any change to the underlying test data is handled in the source repo by changing the appropriate hash. We just need to be sure to continue hosting all data that might need to be accessed. Also note in my kwimage example if wikimedia or imgur ever did change those links, the hashes would not match, so there is not much of a security issue; and because the data is cached locally (and I believe everything is creative commons), I have backups on all my machines so I can always reupload it to a new robust source. 

In any case I do think we should consider moving all test data to be external from the repo (not in this PR we just need this Python stuff merged). When we do make that change CMake's ExternalData mechanism seems like an appealing implementation. From what I understand, its effectively the same concept as "grabdata". There would need to be some sort of staging step that could be performed on the low-side and then once things are staged the repo can be transferred to the high-side. 

I don't think KWIVER should ever have tests that can't be performed in an open environment. For regression tests on sensitive data it should be the responsibility of the data maintainers to also maintain regression tests (and I'd even argue that regression tests on open datasets are sufficient). Its not like we can just simply do a pull in a closed environment anyway, if they're are tests that really need to be done in the closed environment there can be a protocol to perform them when a new version is used in that setting. But again, the open dataset tests should be good enough to grant confidence about the integrity of the system.

Erotemic:
Closing in favor of #1013

PR #927, opened 2020-02-05
Dev/refactor ffmpeg

as6520:
This branch refactors out the implementation of the ffmpeg arrow from ffmpeg. The change enables the creation of a static library that provides the ffmpeg routines. The original implementation did not support a static build since ffmpeg arrow depends on `plugin_manager`. The dependency originates from vital algorithms which uses `plugin_manager` in `get_nested_algo_configuration` and `check_nested_algo_configuration`. 

`plugin_manager` cannot be built statically since the static build violates the singleton pattern that is used by `plugin_manager::instance`. This has been discussed in some detail in [here](https://stackoverflow.com/questions/13869809/multiple-copies-of-static-members-when-mixing-shared-static-libs).


mleotta:
I don't think this is the right strategy to fix this design problem with vital algorithm.  This works, but it makes the code a lot more confusing and we would have to apply a similar hack for every other algorithm that we want to build static.  That will lead to confusion and a very messy code base.  I think the right thing to do is to fix this in the base classes.  That may require some fundamental and breaking changes to the vital API for algorithms, but it's worth it get this right.

We already have the concept of all the algorithms being compiled into a library for each arrow, for example `kwiver_algo_ffmpeg`.  Then the CMake code to make a plugin creates a dynamic library `kwiver_algo_ffmpeg_plugin` for use in the plugin manager.  I think we need to fix the design so that the base libraries can be built static and the plugin libraries add the capabilities to dynamically instantiate those objects.  This may mean we need a major change to how algorithms are constructed from configuration files.

Maybe we should have a meeting to discuss this and include @linus-sherrill

as6520:
I agree that this was not the correct way to fix the issue. However I think that refactoring vital algo would have been beyond the scope of kwiver wheel development where this problem was encountered. This was meant to be a expedient fix to use ffmpeg rather than a long term fix that would fix vital algo.

mleotta:
The problem is that temporary fixes end up not being temporary and complicate the use of the FFMPEG code.  I'd like to see if we can at least identify a plan for a better long term fix before we decide if we need this workaround in the interim.

mleotta:
I believe this PR is no longer relevant so I am going to close it.  Please reopen if that is not the case.

PR #928, opened 2020-02-11
Convert serializers to use PLUGIN_INFO macro.

linus-sherrill:
Added support for custom naming algo factories.


dstoup:
pending dashboard success

PR #929, opened 2020-02-11
make travis call setup_kwiver.sh before running tests

Cookt2:
This branch allows TravisCI on the new python branches to succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,4 +59,5 @@ script:
           ../
   - make -j2
   - make install
+  - . ./setup_KWIVER.sh
   - ctest


Cookt2:
@Erotemic @as6520 @dstoup

as6520:
@Cookt2 I thought that this would target #925 or #924. Did you verify that build in-house?

Cookt2:
> @Cookt2 I thought that this would target #925 or #924. Did you verify that build in-house?

I was going to get this merged into master; then you can rebase off of master. This should be in master anyway, and multiple branches need it.

PR #931, opened 2020-02-11
Add API to optionally load modules

linus-sherrill:
Add parameter to load_all_modules() to select subsets of modules to be loaded.
    
Plugins are distributed in sub-directories off a root directory. The plugin_manager is supplied with the rood directory and adds sub-directories to the search path depending on which plugin types are selected. This shortens the default plugin paths quite a bit.
    
To make this approach work, the build/install paths were changed to group like plugins into separate directories

Plugins are now stored in the following directories:
- processes: <root>/lib/kwiver/plugins/processes
- algorithms: <root>/lib/kwiver/plugins/algorithms
- schedulers: <root>/lib/kwiver/plugins/processes
- applets: <root>/lib/kwiver/plugins/applets
- process-instrumentation: <root>/lib/kwiver/plugins/proc_instr
- plugin explorer extensions: <root>/lib/kwiver/plugins/plugin_explorer
- logger plugins: <root>/lib/kwiver/plugins/logger
- config printer extensions: <root>/lib/kwiver/plugins/modules

Some of the lesser known plugins could be all put into the 'modules' subdirectory, such as process_instrumentation, plugin_explorer, and config extensions, but then you have to decide to give them a meaningful group name or just load them by default.

### DIFF
--- a/CMake/setup_KWIVER.bat.in
+++ b/CMake/setup_KWIVER.bat.in
@@ -8,9 +8,9 @@ if [%1] == [] (
   set config=%1
 )
 
-set PATH=%~dp0/bin/%config%;%~dp0/lib/%config%/@kwiver_plugin_module_subdir@;%~dp0/lib/%config%/@kwiver_plugin_process_subdir@;@EXTRA_WIN32_PATH@;%PATH%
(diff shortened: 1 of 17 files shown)


dstoup:
I like the idea and the code looks good to me.

PR #933, opened 2020-02-12
Add python binding for finalize method

linus-sherrill:
This binding was omitted from the original PR


as6520:
There are a couple of places where the bindings are being tested in sprokit, https://github.com/Kitware/kwiver/blob/87736fe1a556022bec98ed97c740aeb3e150456a/sprokit/tests/bindings/python/sprokit/pipeline/test-process_registry.py#L111 and https://github.com/Kitware/kwiver/blob/87736fe1a556022bec98ed97c740aeb3e150456a/sprokit/tests/bindings/python/sprokit/pipeline/test-run.py#L59. Can these tests be modified to demonstrate finalize?

PR #934, opened 2020-02-17
Add DetectedObjectSetOutputCoco

predicative:
This PR adds a Python implementation for `detected_object_set_output` that writes COCO-format JSON files. It can be used in a pipeline like:
```
process writer :: detected_object_output
  file_name = out.coco.json
  writer:type = coco
  writer:coco:category_start_id = 1  # optional
```

Some points of attention:
- The `"category_start_id"` option was added for an issue @mattdawkins was encountering in VIAME.
  - Matt, what code was causing the issue? As far as I understand, the actual IDs of the categories shouldn't matter, so it would be preferable to fix that code instead of adding the option.
- The properties of the detections represented in the output file are relatively minimal: bounding box, category, and score. However, adding others (e.g. keypoints) might be better for a separate PR.
- No output is actually written until the end, unlike the other detection writers (KW18, (VIAME) CSV, ...). This could be changed, but it would make the code more complicated, and the partial output would be of limited use since e.g. the category labels can only be written at the very end. It still might be useful for memory reasons.
- The code as written requires Python 3.3+. This aligns with KWIVER apparently [requiring 3.4+](https://github.com/Kitware/kwiver/blob/3af282150387daf561a68a4626511fd4047eb995/CMake/kwiver-setup-python.cmake#L130) (for Python 3). However, it would simplify the code some if Python 3.6 was available. It doesn't work with Python 2.7.

### DIFF
--- a/arrows/python/CMakeLists.txt
+++ b/arrows/python/CMakeLists.txt
@@ -7,7 +7,6 @@ kwiver_add_python_module( ${CMAKE_CURRENT_SOURCE_DIR}/simple_image_detector.py
                         kwiver/arrows/python
                         simple_image_detector)
 
-
-
-
-
+kwiver_add_python_module( ${CMAKE_CURRENT_SOURCE_DIR}/detected_object_set_output_coco.py
+                        kwiver/arrows/python
+                        detected_object_set_output_coco)
--- a/arrows/python/detected_object_set_output_coco.py
+++ b/arrows/python/detected_object_set_output_coco.py
@@ -1,0 +1,121 @@
+# ckwg +29
+# Copyright 2020 by Kitware, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice,
+#    this list of conditions and the following disclaimer.
+#
+#  * Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+#  * Neither name of Kitware, Inc. nor the names of any contributors may be used
+#    to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS ``AS IS''
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE AUTHORS OR CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import datetime
+import json
+
+from vital.algo import DetectedObjectSetOutput
+
+class DetectedObjectSetOutputCoco(DetectedObjectSetOutput):
+    """COCO-formatted output for DetectedObjectSets
+
+    See DetectedObjectSetOutput for method information.
+
+    """
+    def __init__(self):
+        DetectedObjectSetOutput.__init__(self)
+        # List of dicts corresponding to elements of the output
+        # "annotations" attribute, minus the "id" attribute
+        self.detections = []
+        # List of image paths
+        self.images = []
+        # Dict mapping category names to numerical IDs
+        self.categories = {}
+        # The current file object or None
+        self.file = None
+
+    def get_configuration(self):
+        cfg = super(DetectedObjectSetOutput, self).get_configuration()
+        return cfg
+
+    def set_configuration(self, cfg_in):
+        cfg = self.get_configuration()
+        cfg.merge_config(cfg_in)
+
+    def check_configuration(self, cfg):
+        return True
+
+    def open(self, file_name):
+        self.file = open(file_name, 'w')
+
+    def close(self):
+        if self.file:
+            self.file.close()
+
+    def write_set(self, detected_object_set, file_name):
+        for det in detected_object_set:
+            bbox = det.bounding_box()
+            d = dict(
+                image_id=len(self.images),
+                bbox=[
+                    bbox.min_x(),
+                    bbox.min_y(),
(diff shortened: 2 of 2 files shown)


Erotemic:
I'm also not a huge fan of "category_start_id". Often, the category ids are known a-priori, and the written categories should align with the known mapping between category names and id. However, in the case where you don't have this information I can see the usefulness of simply incrementing the category it. Perhaps it makes sense to allow the user to pre-specify the category dictionaries, so if the mapping between ids and names is known this implementation would simply use it. 

I'd also like to see docstrings indicating the types of the arguments and return values for "write_set" as well as at least one doctest illustrating a standalone use case for the class. 

Something like:

```python

def _DetectedObjectSet_random():
    """
    This should be a classmethod of DetectedObjectSet to make writing doctests
    easier
    """
    from kwiver.vital.types import BoundingBox, DetectedObject, DetectedObjectSet
    bbox = BoundingBox(0, 10, 100, 50)
    dobj1 = DetectedObject(bbox, 0.2)
    dobj2 = DetectedObject(bbox, 0.5)
    dobj3 = DetectedObject(bbox, 0.4)
    dos = DetectedObjectSet()
    dos.add(dobj1)
    dos.add(dobj2)
    dos.add(dobj3)
    dos.add(dobj3)
    return dos

class DetectedObjectSetOutputCoco(DetectedObjectSetOutput):
    """
    ... 

    Example:
        >>> self = DetectedObjectSetOutputCoco()
        >>> dos = _DetectedObjectSet_random()
        >>> self.write_set(dos, 'foo.bar')
        >>> self.complete()

    """

    def write_set(self, detected_object_set, file_name):
        """
        Args:
            detected_object_set (kwiver.vital.types.DetectedObjectSet):
                Detected objects to write 

            file_name (str):
                the name of the image the detected objects belong to
        """
```

It might be nice for the purposes of testing / demonstration if the output stream could be set to be to stdout or an io.TextIOWrapper for usage within Python.

neal-siekierski:
> I'm also not a huge fan of "category_start_id". Often, the category ids are known a-priori, and the written categories should align with the known mapping between category names and id. However, in the case where you don't have this information I can see the usefulness of simply incrementing the category it. Perhaps it makes sense to allow the user to pre-specify the category dictionaries, so if the mapping between ids and names is known this implementation would simply use it.

The writer currently only uses local information, and `detected_object_type` uses strings, hence the current design. If we want to use existing category IDs, the two main questions I see are how the user should specify those and what the behavior should be if a new category comes up. The answer to the latter is probably something like incrementing, but starting after all the existing IDs.

For the former, we'd want to avoid duplicating work for the user, so it should be possible to specify a mapping using a file the user already has, like Darknet's label file or another COCO file with a `"categories"` property. But we can probably only really pick one format to support.

For what it's worth, I see the category names as primary and the IDs as secondary, so I don't find support for specifying IDs essential, especially with KWIVER not using them internally. I admit that, if you are using other tools that treat the particular ID as significant, being able to preserve IDs is convenient.

> I'd also like to see docstrings indicating the types of the arguments and return values for "write_set" as well as at least one doctest illustrating a standalone use case for the class.

The usage is identical to that of `DetectedObjectSetOutput` / `detected_object_set_output`, hence the reference to the latter's (admittedly sparse) documentation. I could add a doc-test to said base class's doc-string (though that fits better as another PR), but I'm hesitant to repeat information in the subclass that's identical to that in the base class. I can at least add comments to the effect that the various methods are overrides.

> It might be nice for the purposes of testing / demonstration if the output stream could be set to be to stdout or an io.TextIOWrapper for usage within Python.

That's a good idea and I can add that. This will also make my point about not repeating documentation less relevant, since as a result the class and some methods (though not `write_set`) will be different.

Erotemic:
> For what it's worth, I see the category names as primary and the IDs as secondary

I can buy that. I'm ok as long as the string names are persisting. It makes sense that another part of the system can remap the ids as needed. 

>  The usage is identical to that of DetectedObjectSetOutput / detected_object_set_output, hence the reference to the latter's (admittedly sparse) documentation. I could add a doc-test to said base class's doc-string (though that fits better as another PR), but I'm hesitant to repeat information in the subclass that's identical to that in the base class. I can at least add comments to the effect that the various methods are overrides.

I think it makes sense to have at least some repeated information. That way when the user clicks on the sphinx documentation page and they see nothing, they don't have to think to look at the docs for the parent class and try to discern what is different. I think its at least worth having a runnable example in every class. However, a simple link will suffice for the boilerplate docs. For this use 

```python
"""
See :class:`DetectedObjectSetOutput` for method information.
"""
```

so sphinx can create the proper documentation hyperlink. 

Apart from that I'm just looking to see (and run) the doctests. On the off-chance you haven't heard me talk about xdoctest, take a look at how you can use its cli to run for a single doctest in a function or class: https://xdoctest.readthedocs.io/en/latest/

mattdawkins:
> Matt, what code was causing the issue? As far as I understand, the actual IDs of the categories shouldn't matter,

Training mmdet required categories to start with label 1 as 0 was reserved for background

linus-sherrill:
This PR has developed a merge conflict. Other than that, is this approved?

dstoup:
Don't see a dashboard run. That would be important after the conflicts are resolved.

neal-siekierski:
There are still the changes @Erotemic's suggested, namely improved docs and fancier stream handling. I still don't like `category_start_id`; the simplest way to get rid of it would be to always start from 1, since practically I don't think we're going to fix mmdet, there's probably nothing that reserves 1 and up, and starting from 1 isn't too arbitrary. @Erotemic, thoughts?

Erotemic:
@neal-siekierski Responding to your May 18th content. 

As to mmdetection, they changed how they want to encode their category ids. It used to be that background was always `0`, and now its that background is always `len(classes)`. IMO, their decision to enforce a standard here is over restrictive. There is no reason why they can't let the user indicate which class id corresponds to "background". 

For the COCO format in general, the category ids can be arbitrary as long as they are internally consistent. However, if you wish to use two COCO files together, it often helps if the category ids are aligned (sometimes tooling will try to "fix" these by re-assigning ids such that the category names align, but not always). I do think it would be best if there was a way to specify what the category-id-mapping was a-priori, but I guess that can always be rectified later. 

WRT to the comment about maintaining the actual COCO data-structure instead of the proxy, I was thinking something like this: 

```python

def __init__(self):
    # Dict mapping category names to numerical IDs
    self.catname_to_id = {}
    # Core JSON data structure
    self.dataset =  {
        'images': [],
        'categories': [],
        'annotations': [],
        'info': None,
    }


def write_set(self, detected_object_set, file_name):
    dataset = self.dataset

    # Create the "next" image COCO object
    img = {
        'id': len(dataset['images']),
        'file_name': file_name,
    }
    dataset['images'].append(img)

    for det in detected_object_set:
        # Create a new annotation for each detected object
        bbox = det.bounding_box()
        ann = {}
        ann['id'] = len(dataset['annotations'])
        ann['image_id'] = img['id']
        ann['bbox'] = [bbox.min_x(), bbox.min_y(), bbox.width(), bbox.height()]
        ann['score'] = det.confidence()

        if det.type() is not None:
            # Check if the category exists already
            catname = det.type().get_most_likely_class()
            if catname not in self.catname_to_id:
                # Add a new category
                cat = {
                    'id': len(dataset['categories']) + self.category_start_id,
                    'name': catname
                }
                dataset['categories'].append(cat)
                self.catname_to_id[catname] = cat['id']
            ann['category_id'] = self.catname_to_id[catname]
        else:
            ann['category_id'] = None
        dataset['annotations'].append(ann)


def complete(self):
    import datetime
    import json
    now = datetime.datetime.now(datetime.timezone.utc).astimezone()
    self.dataset['info'] = dict(
        year=now.year,
        description="Created by DetectedObjectSetOutputCoco",
        date_created=now.replace(microsecond=0).isoformat(' '),
    )
    json.dump(self.dataset, self.file)

```

Note that the main difference is that I'm maintaining `self.dataset` which is the core json COCO data structure instead of `self.detections`, `self.images`, etc. This should make it easier to extend in the future.

dstoup:
Just noticed the conflict here. Likely that has something to do with all the binding work that was done lately.

PR #935, opened 2020-02-19
Wrap detected_object_set_input for Python

neal-siekierski:
This PR adds wrapping for the `detected_object_set_input` algo. In line with the choices made in #919, its `close` method is made virtual and `stream` is not wrapped. Of note is that `read_set`, which in C++ has a boolean return value and two by-reference output parameters, is wrapped to return a pair or `None`.

@mattdawkins, as I recall `viame/master` changes `read_set` to make `image_path` an in-out parameter (or more precisely, sometimes an input and sometimes an output parameter). That doesn't change the signature in C++, but it corresponds to a different Python signature (`path -> (set, path)` for in-out and `path -> set` for just output). For compatibility with that, I think a separate, for-VIAME patch that adds a Python method `read_set_by_path` that acts as the path-as-input "overload" for `read_set` would be the cleanest approach. This would leave the meaning and use of `read_set` in Python as it is in this PR. (This is relevant to VIAME/VIAME#101.)

### DIFF
--- a/python/kwiver/vital/algo/detected_object_set_input.cxx
+++ b/python/kwiver/vital/algo/detected_object_set_input.cxx
@@ -30,13 +30,15 @@
 #include <utility>
 
 #include <pybind11/pybind11.h>
-#include <vital/bindings/python/vital/algo/trampoline/detected_object_set_input_trampoline.txx>
-#include <vital/bindings/python/vital/algo/detected_object_set_input.h>
+#include <python/kwiver/vital/algo/trampoline/detected_object_set_input_trampoline.txx>
+#include <python/kwiver/vital/algo/detected_object_set_input.h>
 
 namespace py = pybind11;
 
 using dosi = kwiver::vital::algo::detected_object_set_input;
-
+namespace kwiver {
+namespace vital {
+namespace python {
 void detected_object_set_input(py::module &m)
 {
   py::class_< dosi,
@@ -56,4 +58,7 @@ file name, or None if the input is exhausted)")
     .def("open", &dosi::open)
     .def("close", &dosi::close)
     ;
+}
+}
+}
 }
--- a/python/kwiver/vital/algo/detected_object_set_input.cxx
+++ b/python/kwiver/vital/algo/detected_object_set_input.cxx
@@ -1,5 +1,5 @@
 /*ckwg +29
- * Copyright 2020 by Kitware, Inc.
+ * Copyright 2019-2020 by Kitware, Inc.
  * All rights reserved.
  *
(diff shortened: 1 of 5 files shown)


linus-sherrill:
Has this been reviewed and ready to go?

neal-siekierski:
I've added tests of the first sort using `detected_object_set_input_simulator`. However, one of the tests depends on the changes in #952, so as is I expect it to fail. When #952 is merged I can merge in / rebase onto `master`.

neal-siekierski:
... And it puzzlingly succeeded. It looks like the Python tests aren't being run at all. Moreover, the [`CMakeLists.txt`](https://github.com/Kitware/kwiver/blob/072edffcd4df0e93a35756e905bf66a8f12d3bf7/vital/bindings/python/vital/tests/CMakeLists.txt) file in the folder doesn't register any tests, and a quick search suggests that nothing at the CMake level is creating tests for the modules in this folder. Also, for what it's worth, Travis is still using Python 2.7.

neal-siekierski:
@as6520, I've added some tests. Could you take another look?

Also, it would be good to see the test situation clarified (whether by you or someone else).

as6520:
I believe that there is PR in #940 that would change the build the default to python 3. There are two different types of python tests in kwiver, the ctest based tests that are in sprokit and nose/unittest based test in vital. The CI runs the first set of tests but it does not run the second type. @Cookt2  can we add pytest based run for testing vital types. We have started doing that in #925 in the docker image that releases the wheel. I can dm you the command along with the caveats. 

I'll take a look at the tests and get back to you.

as6520:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
There was already a detected_object_set_input wrapped in dev/overhaul, so the diffs might look different due to the rebase. I think all of the changes are there though.

@neal-siekierski @as6520 If at all possible, could you review the changes to double check nothing is lost? I have a copy of the branch's previous state on my github fork of kwiver, so we can always go back.

as6520:
> @neal-siekierski @as6520 If at all possible, could you review the changes to double check nothing is lost? I have a copy of the branch's previous state on my github fork of kwiver, so we can always go back.

Looking through the changes quickly, I think the imports and includes need to be fixed. Apart from that I think the changes are all in and in the correct place. I can work on those fixes and then ask for a review.

as6520:
Jenkins test this please

as6520:
@linus-sherrill This PR is ready to be merged into master.

dstoup (approved):

PR #940, opened 2020-02-26
Default python version to 3

dstoup:


linus-sherrill:
Jenkins test this please

dstoup:
I had some issues with the same test when I tried this a few months ago. I'm not python smart enough to understand the issue though.

624 - test-python-pymodules-pythonpath (Failed)

linus-sherrill:
Jenkins test this please

dstoup:
Very strange. The failing test here is the same as the failing test when I make python 3 the default version. I suspect there is something not right about this test that it would fail when things like these are changed. Trouble is, I have no idea what's going on in that test so I'm not going to be of much use beyond this comment.

PR #945, opened 2020-02-27
Adjust Log Levels

mattdawkins:
VIAME prefers these for displays to users (info level)

### DIFF
--- a/arrows/core/video_input_image_list.cxx
+++ b/arrows/core/video_input_image_list.cxx
@@ -362,7 +362,7 @@ video_input_image_list
 {
   if ( !d->m_image && this->good() )
   {
-    LOG_DEBUG( logger(), "reading image from file \"" << *d->m_current_file << "\"" );
+    LOG_INFO( logger(), "reading image from file \"" << *d->m_current_file << "\"" );
 
     // read image file
     //
--- a/vital/bindings/python/vital/modules/module_loader.py
+++ b/vital/bindings/python/vital/modules/module_loader.py
@@ -49,7 +49,7 @@ def _load_python_module(mod):
                 logger.warn(('Python module "{}" defined {} but '
                              'it is not callable').format(mod, registrar))
 
-    logger.warn(('Python module "{}" does not have registrar method').format(mod))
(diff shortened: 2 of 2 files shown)


arrows/core/video_input_image_list.cxx
```diff
@@ -362,7 +362,7 @@ ::frame_image()
 {
   if ( !d->m_image && this->good() )
   {
-    LOG_DEBUG( logger(), "reading image from file \"" << *d->m_current_file << "\"" );
+    LOG_INFO( logger(), "reading image from file \"" << *d->m_current_file << "\"" );
```
mleotta: This is somewhat subjective.  I'm not going to argue too strongly one way or the other on whether this message should be `debug` or `info`.  However, given that this is *video* input there are likely a lot of these messages that will be reported, and the average user of video input that wants high-level `info` comments probably doesn't want to see a message about every frame that is loaded.

Note that logging is configurable and can be controlled on a per algorithm basis.  So if Viame wants to see these messages it might be better to setting the log level for `arrows.core.video_input_image_list` to `debug` in the Viame logger configuration.

vital/bindings/python/vital/modules/module_loader.py
```diff
@@ -49,7 +49,7 @@ def _load_python_module(mod):
                 logger.warn(('Python module "{}" defined {} but '
                              'it is not callable').format(mod, registrar))
 
-    logger.warn(('Python module "{}" does not have registrar method').format(mod))
+    logger.debug(('Python module "{}" does not have registrar method').format(mod))
```
linus-sherrill: I could see that this warning is useful in determining why your module will not load. If there are too many of these messages, is there another way to limit them such as moving the non-loadable python modules to a different directory?

as6520: Agreed that this warning is useful. One of the changes that was made in the wheel was to start specifying python modules explicitly rather than searching for modules in a package. This is mainly being used for python plugin registration that is being done using entry-point. Maybe we can impose the same requirement for SPROKIT_PYTHON_MODULES based registration.

PR #955, opened 2020-03-17
Dev/pipeline bindings

tao558:


tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

tao558:
Jenkins test this please

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -0,0 +1,131 @@
...
+    {
+      self.add_value<kwiver::vital::any>(port, obj);
+    })
+    .def("add_datum", &adapter_data_set::add_datum)
+
+    // The add_value function is templated.
+    // To bind the function, we must bind explicit instances of it,
+    // each with a different type.
+    // First the native C++ types
+    .def("add_int", &adapter_data_set::add_value<int>)
```
Erotemic, line 80: Is it possible to template pybind11 bindings? 

Looking at this issue it may be possible: https://github.com/pybind/pybind11/issues/199 

It may be better to just explicitly create all possible values, I'm not sure.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -0,0 +1,131 @@
...
+PYBIND11_MODULE(adapter_data_set, m)
+{
+  enum_<adapter_data_set::data_set_type>(m, "DataSetType"
+    , "Type of data set.")
+    .value("data", adapter_data_set::data)
+    .value("end_of_input", adapter_data_set::end_of_input)
+  ;
+
+  class_< adapter_data_set, std::shared_ptr<adapter_data_set > >(m, "AdapterDataSet")
+    .def_static("create", &adapter_data_set::create
```
Erotemic, line 60: Does it make sense to add a `__str__` and `__repr__`  to this class? Not sure what the purpose of an AdapterDataSet is. It might also be nice to have some a class-level docstring with an "Example" doctest block if it is something that we expect users to work with. Note that `xdoctest` supports executing doctests within pybind11 modules, so the example will help with CI testing. I have another class that I did this on if you need an example (see `vital/types/image.cxx`).

python/kwiver/sprokit/adapters/embedded_pipeline.cxx
```diff
@@ -0,0 +1,154 @@
...
+}
+}
+
+namespace ksp = kwiver::sprokit::python;
+
+PYBIND11_MODULE(embedded_pipeline, m)
+{
+  class_< kwiver::embedded_pipeline,
+        std::shared_ptr<kwiver::embedded_pipeline>,
+        ksp::embedded_pipeline_trampoline>(m, "EmbeddedPipeline")
```
Erotemic, line 78: Again, a class-level docstring with an example would be very useful.

python/kwiver/sprokit/tests/sprokit/adapters/test-adapter_data_set.py
```diff
@@ -0,0 +1,272 @@
...
+        test_error("Expected three arguments")
+        sys.exit(1)
+
+    testname = sys.argv[1]
+
+    os.chdir(sys.argv[2])
+
+    sys.path.append(sys.argv[3])
+
+    from kwiver.sprokit.util.test import *
```
Erotemic, line 270: I have a few issues this block: 

1. IMHO `import *` is bad practice. It makes it confusing where things are coming from and it prevents the use of certain static analysis tools. Either `from kwiver.sprokit.util import test as <foobar>` and use `<foobar>.name` or do `from kwiver.sprokit.util.test import (<the>, <members>, <you>, <want>)`. There are rare cases where `import *` is acceptable, but I don't think this is one of them. 

2. This import should not be in the `if __name__ == "__main__":` block. If you were to import the functions from this module and try to call them, they would Error when they reached a name like `test_error`. Is there a problem with putting it at the top of the module?

3. Can we get rid of the `os.chdir`, and `sys.path.append`? They are awful fragile constructs. Usually when I see these patterns there is another more robust way to implement the same functionality.

python/kwiver/sprokit/tests/sprokit/adapters/test-embedded_pipeline.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Now send some data
+    for i in range(10):
+        ds = adapter_data_set.AdapterDataSet.create()
+        val = i
+
+        for port in input_list:
+            ds.add_int(port, val)
+            val += 1
+
```
Erotemic, line 104: Logic can be simplified as such: 

```python
for val, port in enumerate(input_list, start=i):
     ds.add_int(port, val)
```

No need for `val =i` and `val += 1`

python/kwiver/sprokit/tests/sprokit/adapters/test-embedded_pipeline.py
```diff
@@ -0,0 +1,154 @@
...
+    testname = sys.argv[1]
+
+    os.chdir(sys.argv[2])
+
+    sys.path.append(sys.argv[3])
+
+    pipeline_dir = sys.argv[4]
+
+    from kwiver.sprokit.util.test import *
```
Erotemic, line 152: Same import gripes as above.

python/kwiver/sprokit/tests/sprokit/adapters/test-embedded_pipeline.py
```diff
@@ -0,0 +1,154 @@
...
+            break
+        for (port, d) in ods:
+            print("   port:", port, " value:", d.get_int())
+
+
+    ep.wait()
+
+    #######
+    # Still need to test stop()
+    ep = embedded_pipeline.EmbeddedPipeline()
```
Erotemic, line 128: Seems like `EmbeddedPipeline` could use an `__enter__` and `__exit__` method so it could be used with the `with` statement. This is just a note, and nothing needs to be done to address this in this PR.

python/kwiver/sprokit/tests/sprokit/adapters/test-adapter_data_set.py
```diff
@@ -0,0 +1,272 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+def test_import():
+    try:
+        import kwiver.sprokit.adapters.adapter_data_set
+    except:
+        test_error("Failed to import the adapter_data_set module")
+
+
+def check_port_val_equals(expected_val, actual_val, portname):
```
Erotemic, line 38: Is this an unused function?

python/kwiver/sprokit/adapters/__init__.py
```diff
@@ -0,0 +1,4 @@
+from __future__ import absolute_import
```
Erotemic, line 1: Are we still supporting Python 2? (My guess is yes, but I just want to check)

tao558: Yes, we are still supporting Python 2. Python 3 is the default, but users can switch to Python 2 in cmake.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: I believe that in that issue, they are binding explicit instances of it. I haven't seen any other way to do it.

python/kwiver/sprokit/tests/sprokit/adapters/test-adapter_data_set.py
tao558: Yes, resolved in eaa619d6921e23c5d5ac45a5a63c8873c9cadd3d

python/kwiver/sprokit/tests/sprokit/adapters/test-embedded_pipeline.py
tao558: Much better. Resolved in e7dbafbca27d1c95722c142fe333969ce1200319

python/kwiver/sprokit/tests/sprokit/adapters/test-adapter_data_set.py
tao558: To answer point number 3, it turns out that both of those calls are unnecessary. I'll create a branch to fix that for all of sprokit

tao558: This should be handled by #1056

python/kwiver/sprokit/tests/sprokit/adapters/test-embedded_pipeline.py
tao558: This should also be handled by #1056

python/kwiver/sprokit/tests/sprokit/adapters/test-adapter_data_set.py
tao558: Resolved in 90e7c1907abf246285692646feaa83e2e645df20

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: I'm not so sure about `__str__` and `__repr__`. I think they would be good to eventually have, since adapter_data_set (ADS) is essentially a map/python dictionary, but implementing it right now would be difficult and not very beneficial. We would essentially have to loop through the ADS ports/values, call `get_datum<vital::any>` on the values to avoid having to deal with specific template types, cast those to py::objects, then call py::str() on the casted values (if that will work). A fair amount of vital types don't have `__str__` or operator<< (this is still a WIP), so we would likely be getting things along the lines of "timestamp object at 0x....." most of the time.

python/kwiver/sprokit/adapters/embedded_pipeline.cxx
tao558: Successful docstring added in 4c92416d1d4373d10a7e85ded265c3b5aef134fe

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: Docstrings added in 4c92416d1d4373d10a7e85ded265c3b5aef134fe

tao558:
Jenkins test this please

python/kwiver/sprokit/adapters/adapter_data_set.cxx
Erotemic: We should maintain an issue where we outline "Pythonic" improvements that could be made to these bindings. 

If ADS is effectively a dict, then it should have a dict-like interface with    `__iter__`, `__len__`, `__contains__`, `__getitem__`, `items`, `keys`, `values`, `get`,  and if it should be writable it should have:   `__delitem__`, `__setitem__`, `update`, and perhaps: `copy`. 

For `__str__` and `__repr__` I think its useful for objects like this to behave as a dict  in that respect in order to make debug outputs readable. There are exceptions to this rule though, like if we expect ADS to contain a ton of info, then I typically like to print a summary like number of items. In general though I think its always good practice for a `__str__` and `__repr__` to give some indication as to the contents of an object. The default implementations are very restrictive. 

In this case, I think the keys might have important information even if the values were relatively sparse.

tao558: I've created a few issues for vital types that could use some extra pythonic functionality. I think this would be a good idea in this case.

As for str and repr, we may be able to print the type of the value for each key.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -0,0 +1,147 @@
...
+    .def("get_port_data_f2f_homography", &adapter_data_set::get_port_data<kwiver::vital::f2f_homography>)
+    ;
+    ads.doc() = R"(
+        Python bindings for kwiver::adapter::adapter_data_set
+
+        Example:
+            >>> from kwiver.sprokit.adapters import adapter_data_set
+            >>> # Following ads has type "data". We can add/get data to/from ports
+            >>> ads = adapter_data_set.AdapterDataSet.create()
+            >>> assert ads.type() == adapter_data_set.DataSetType.data
```
Erotemic, line 138: would isinstance be more applicable here?

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -0,0 +1,147 @@
...
+            >>> # Can add as a general python object
+            >>> ads.add_value("port1", "a_string")
+            >>> # Can also add by specifying type
+            >>> ads.add_int("port2", 5)
+            >>> # Get both values
+            >>> assert ads.get_port_data("port1") == "a_string"
```
Erotemic, line 144: Yeah, Python `__getitem__` and `__setitem__` would be really useful for this class. Making this as dict-like as possible would make it much more Pythonic.

python/kwiver/sprokit/tests/sprokit/adapters/test-adapter_data_set.py
```diff
@@ -0,0 +1,258 @@
...
+    OVERWRITE_PORT = "test_overwrite_port"
+    ads = adapter_data_set.AdapterDataSet.create()
+
+    # Overwriting with same datum
+    ads.add_datum(OVERWRITE_PORT, datum.new("d2"))
+    overwrite_helper(ads.add_datum, ads.get_port_data, datum.new("d3"), "datum_string", OVERWRITE_PORT)
+
+    # Overwriting with completely different types
+    overwrite_helper(ads.add_datum, ads.get_port_data, datum.new(12), "datum_int", OVERWRITE_PORT)
+    overwrite_helper(ads.add_string_vector, ads.get_port_data_string_vector, datum.VectorString(["baz", "qux"]), "string_vector", OVERWRITE_PORT)
```
Erotemic, line 214: Just noting that it would be useful for there to be something like `datum.coerce(d)`, which takes `d` and returns the datum of the appropriate type depending on the input Python type (or TypeError if unknown). Also `ads.add` where it is able to coerce the type correctly would be useful. Not sure if these already exist and you are using explicit calls, but if not these would be goods thing to add to Pythonic improvements.

tao558:
Jenkins test this please

tao558:
Jenkins test this please

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: Added in 1c40fdd8f27f207f3620e362328f94bf75ee9e11

python/kwiver/sprokit/tests/sprokit/adapters/test-adapter_data_set.py
tao558: Implemented for both ADS and datum in 1c40fdd8f27f207f3620e362328f94bf75ee9e11

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: `__iter__` was present in the original bindings. In 1c40fdd8f27f207f3620e362328f94bf75ee9e11, I added `__str__`, `__repr__`, `__nice__`, `__getitem__`, `__setitem__` and `__len__`. Most of the rest of those would require changes to the actual adapter_data_set class to be efficient (simply wrapping std::map's functions). If you'd prefer that I add the rest I would be happy to do so, otherwise I can create an issue for the remaining methods.

tao558:
Jenkins test this please

tao558:
Jenkins test this please

linus-sherrill (commented):
I did a quick review and did not look at everything in depth, but generally, this looks good.

tao558:
Jenkins test this please

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -112,30 +112,30 @@ py::object get_port_data_correct_type(adapter_data_set &self, ::sprokit::process
 {
   kwiver::vital::any const any = self.get_port_data<kwiver::vital::any>(port);
 
-  #define GET_OBJECT(TYPE) \
-  if (typeid(TYPE).hash_code() == any.type().hash_code()) \
+  #define ADS_GET_OBJECT(TYPE) \
+  if (any.is_type<TYPE>()) \
```
Erotemic, line 116: I don't understand the C++ here enough to know what is happening. Can I get an explanation of old versus new behavior?

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -0,0 +1,296 @@
...
+}
+
+}
+}
+}
+
+
+PYBIND11_MODULE(adapter_data_set, m)
+{
+  py::bind_vector<std::vector<unsigned char>, std::shared_ptr<std::vector<unsigned char>>>(m, "VectorUChar", py::module_local(true));
```
Erotemic, line 154: Comment here describing these would be good.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -0,0 +1,296 @@
...
+    {
+      return py::make_iterator(self.cbegin(), self.cend());
+    }, py::keep_alive<0,1>())
+
+    // Members
+    .def("type", &adapter_data_set::type)
+    .def("is_end_of_data", &adapter_data_set::is_end_of_data)
+
+    // General add_value which adds any type, and __setitem__
+    .def("add_value", &kwiver::sprokit::python::add_value_correct_type)
```
Erotemic, line 178: This should have a docstring indicating it is equivalent to `__getitem__`, the same with the `get_port_data` function.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
```diff
@@ -0,0 +1,296 @@
...
+    .def("_get_port_data_corner_points", &adapter_data_set::get_port_data<kwiver::vital::geo_polygon>)
+    .def("_get_port_data_f2f_homography", &adapter_data_set::get_port_data<kwiver::vital::f2f_homography>)
+    .def("__nice__", [](const adapter_data_set& self) -> std::string {
+    auto locals = py::dict(py::arg("self")=self);
+      py::exec(R"(
+          retval = 'size={}'.format(len(self))
+      )", py::globals(), locals);
+      return locals["retval"].cast<std::string>();
+      })
+  .def("__repr__", [](py::object& self) -> std::string {
```
Erotemic, line 249: Doesn't have to happen here, but it might be useful to make a separate pybind11 class that other pybind11 classes can inherit from similar to [NiceRepr](https://ubelt.readthedocs.io/en/0.9.1/_modules/ubelt/util_mixins.html#NiceRepr) in ubelt. (although note that in this case the `__str__` method will still exist along with `__nice__`.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: So in an ideal world, the implementation with hash codes would just work. However, under lib C++, this comparison will fail if two equivalent types are defined in different shared libraries. This is what was causing the Mac errors. Is_type compares the names of the type. Using is_type is probably the best option regardless, since it's in vital.

Erotemic: I see, `vital.any.type()` and `typeid(.)` return a `std::type_info`, which contains the `hash_code` method.   

When you say two libs define equivalent types, you mean that two libs share type definitions that have the same names (and maybe the same methods)? In that case I can understand why the `hash_code` would be different. Its like in Python if you defined the same class in two files, using isinstance(foo.Bar(), baz.Bar) is going to (and should) return False even if Bar is functionally the same. 

It seems slightly dangerous to only rely on the names here given that its very easy for another lib to define a class with the same name, but I guess that's the behavior we want?  

It might be worth adding a comment discussing this behavior in case this comes back to bite us in the future.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: How does a0322134e3f4a072f38fcba0b4da0bab9d227a23 look?

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: Added in a0322134e3f4a072f38fcba0b4da0bab9d227a23. Is that clear?

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: > When you say two libs define equivalent types, you mean that two libs share type definitions that have the same names (and maybe the same methods)?

Not quite, sorry for not being clear. When 2 pybind modules have bindings for the same C++ type, _that_ `type_info`
/ `type_info.hash_code()`  comparison fails. In our example, the pybind modules `datum` and `adapter_data_set` both provide bindings for `std::vector<double>` (and the other vector instantiations). Consider the following code:

`> ads = AdapterDataSet.create()`
` # Notice that we use datum.VectorDouble(), not adapter_data_set.VectorDouble()`
`> d = datum.new(datum.VectorDouble([5.2, -3.14]))`
`> ads["p1"] = d`
`> ads["p1"]`

That last call to `get_port_data_correct_type` is what would cause issues. In that function, we eventually check if the object at that port is an `std::vector<double>`, which in this case, should return true. The issue is that the `std::vector<double>` stored at that port was created from the `datum` module, not the `adapter_data_set` module. The `std::type_info`'s and their `hash_codes()` are not the same, even though the underlying C++ type is the same, so the comparison would return false.

If you'd still prefer that I add a comment, I'm happy to do so.

python/kwiver/sprokit/adapters/adapter_data_set.cxx
Erotemic: Yes, this is helpful.

tao558:
Jenkins test this please

as6520 (approved):

dstoup (commented):
I didn't dig in to the python, trusting our experts have already flushed that out. The C++ and CMake look good. I made one minor comment but otherwise I'm happy if the Python folks are happy.

CMake/utils/kwiver-utils-tests-python.cmake
```diff
@@ -65,16 +65,8 @@ endfunction ()
...
-  set(python_chdir          ".")
 
-  if (CMAKE_CONFIGURATION_TYPES)
-    set(python_module_path      "${kwiver_python_output_path}/$<CONFIGURATION>/${python_sitename}")
-    set(python_chdir           "$<CONFIGURATION>")
-  endif ()
-
-  kwiver_add_test(python-${group} ${instance}
-    "${python_chdir}" "${python_module_path}" ${ARGN})
+  kwiver_add_test(python-${group} ${instance} ${ARGN})
```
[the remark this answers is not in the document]
tao558: I think you're right. Removed in 4559547b3fff00163f7198dd063c6a60a83c96a1

tao558:
Jenkins test this please

tao558:
Jenkins test this please

python/kwiver/sprokit/adapters/adapter_data_set.cxx
Erotemic: I think that is roughly what I was thinking. Two pybind modules would effectively be different C++ shared libraries (i.e. different .so files). In each of those modules you create a new python class VectorDouble that has the same underlying C++ implementation, but because it was defined separately in each pybind module, it gets a different type hash code. I think its a non-trivial enough issue that a comment pointing back to this conversation is warranted.

tao558:
Jenkins test this please

python/kwiver/sprokit/adapters/adapter_data_set.cxx
tao558: Added a comment in b2ca3d8828aa87126cb981b57a759eda20e25002.

We are comparing the underlying C++ types, not the Python types/names.

tao558:
Jenkins test this please

tao558:
Jenkins test this please

PR #960, opened 2020-03-20
database_query::set_temporal_bounds now throws logic_error appropriately

tao558:
This PR fixes database_query::set_temporal_bounds. The comments indicate that the function should throw an std::logic_error if the upper bound parameter is less than the lower bound parameter, but this was not implemented. This PR adds the exception, and tests to show that the function now works as expected.

### DIFF
--- a/vital/types/database_query.h
+++ b/vital/types/database_query.h
@@ -45,6 +45,7 @@
 #include <vital/vital_config.h>
 
 #include <memory>
+#include <stdexcept>
 
 namespace kwiver {
 namespace vital {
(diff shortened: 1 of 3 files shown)


vital/types/database_query.h
```diff
@@ -45,6 +45,7 @@
 #include <vital/vital_config.h>
 
 #include <memory>
+#include <stdexcept>
```
mattdawkins: Did this need to be added to the .h instead of the .cxx?

linus-sherrill: I would recommend moving this include to the implementation since it is not needed in the interface definition.

tao558: Yeah my mistake. I'll push this change right away

PR #964, opened 2020-03-31
Normalizing config for algorithms in processes.

linus-sherrill:
Standardize config entry for specifying algorithm type. Rolled documentation into this entry so all processes that instantiate an algorithm get the same extended documentation. The goal is to make it easy to standardize the documentation and config handling.

- Convert throw to VITAL_THROW where applicable

- Clean up some inconsistencies how algo's are specified.

- Remove outdated attach_logger() call from processes.


mleotta:
I did not have time to review in detail, and I'm not up to speed on sprokit.  However, I am in favor of more consistently using VITAL_THROW.

dstoup:
Jenkins test this please

PR #967, opened 2020-04-01
Dev/add null checks track descriptor

tao558:
This PR adds logic for throwing an std::logic_error in track_descriptor::at and track_descriptor::descriptor_size. The exceptions guard against dereferencing the "data_" member variable if its still NULL, which causes a segfault. These changes are only compiled for debug builds.

track_descriptor::has_descriptor has also been modified to return false if the data_ member is still NULL

### DIFF
--- a/vital/types/track_descriptor.cxx
+++ b/vital/types/track_descriptor.cxx
@@ -30,7 +30,6 @@
 
 #include "track_descriptor.h"
 
-#include <assert.h>
 #include <sstream>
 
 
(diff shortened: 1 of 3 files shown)


vital/types/track_descriptor.cxx
```diff
@@ -153,6 +157,15 @@ double&
 track_descriptor
 ::at( const size_t idx )
 {
+  #ifdef DEBUG_CHECK_DATA_NULL
+  if (!this->data_)
+  {
+    std::stringstream msg;
+    msg << "Attempted to access raw data of null descriptor ptr";
+    throw std::logic_error( msg.str() );
```
linus-sherrill, line 165: If the `msg` is just a simple string, you can just put that string in the initializer for `std::logic_error( "Attempted...ptr" );`

And below too.

PR #970, opened 2020-04-09
Fix inconsistent python install paths

linus-sherrill:
There are two different ways of composing the python install path in the kwiver-utils-python.cmake  file and they are different. Usually this does not cause a problem, but sometimes it does. In any event, why are there two different ways of composing this path? One approach uses platform specific logic to determine components of the path, the other does not. If all this additional logic is needed, then it should be refactored so both places can benefit.

This change centralizes the install path creation.

### DIFF
--- a/CMake/utils/kwiver-utils-python.cmake
+++ b/CMake/utils/kwiver-utils-python.cmake
@@ -123,16 +123,11 @@ function (kwiver_add_python_module path     modpath    module)
   if (WIN32)
     if (python_noarch)
       return ()
-    else ()
-      set(python_install_path lib)
     endif ()
   else ()
     if (python_noarch)
       set(python_noarchdir /noarch)
-      set(python_install_path lib)
       set(python_arch u)
-    else ()
-      set(python_install_path "lib${LIB_SUFFIX}")
     endif ()
   endif ()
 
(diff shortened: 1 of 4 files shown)


linus-sherrill:
Strangely enough, I discovered that 
`distutils.sysconfig_get_python_lib(prefix='/my/install/root/')`
and
`distutils.sysconfig_get_python_lib(prefix='')`
produce different results. One specifies `site-packages` and the other `dist-packages`
Will leave the implementation as is.

Purg:
I to this day do not know why debian/ubuntu decided that they needed a differently named site-packages directory... I'm would love to be enlightened...

linus-sherrill:
Does this look good enough to merge?

CMake/utils/kwiver-utils-python.cmake
```diff
@@ -124,15 +124,12 @@ function (kwiver_add_python_module path     modpath    module)
     if (python_noarch)
       return ()
     else ()
-      set(python_install_path lib)
```
Purg, line 127: Is this and the one below now empty "else" statements?

linus-sherrill: Good point - fixed

PR #979, opened 2020-04-22
Change local_cartesian test suite name, change variable in conversion test

tao558:
This PR makes a few small changes to the tests for the kwiver vital type local_cartesian:

1.) Change the first argument to the TEST macro from geo_point to local_cartesian. This was causing an issue where local_cartesian test cases would look like they were being run under the geo_point test suite

2.) Change the local_cartesian variable being used from lines 131 on from "lc_lla" to "lc_utm", which was previously not used in the test. This object is meant to be tested from line 131 on instead of lc_lla. The accuracy of the location checks needed to be decreased from `10^-8` to `10^-7` to allow these new tests to pass

### DIFF
--- a/vital/tests/test_local_cartesian.cxx
+++ b/vital/tests/test_local_cartesian.cxx
@@ -60,7 +60,7 @@ main(int argc, char* argv[])
 }
 
 // ----------------------------------------------------------------------------
-TEST(geo_point, constructor)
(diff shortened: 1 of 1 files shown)


tao558:
Jenkins test this please

linus-sherrill:
Jenkins test this please

PR #980, opened 2020-04-23
Dev/update geo mgrs

tao558:
This PR makes a few changes to the kwiver vital type geo_MGRS:

1.) In the header file for the type, VITAL_EXPORT calls were added in the class declaration and before the insertion operator. Previously, using the type in any code outside the header and implementation files would cause linker errors, which is solved by this change.

2.) A test suite is added to test the vital type, and to ensure that the above linker errors don't occur.

### DIFF
--- a/vital/tests/test_geo_MGRS.cxx
+++ b/vital/tests/test_geo_MGRS.cxx
@@ -1,0 +1,139 @@
+/*ckwg +29
+ * Copyright 2020 by Kitware, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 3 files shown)


tao558:
Jenkins test this please

vital/tests/test_geo_MGRS.cxx
```diff
@@ -0,0 +1,139 @@
+/*ckwg +29
+ * Copyright 2016-2017 by Kitware, Inc.
```
linus-sherrill, line 2: Oops - wrong date

tao558: Good catch. Should be all set now.

linus-sherrill:
Jenkins test this please

linus-sherrill:
Jenkins test this please

tao558:
Jenkins test this please

PR #983, opened 2020-04-24
Deprecate sprokit CMake support

linus-sherrill:
Convert sprokit CMake support to the equivalent KWIVER CMake support.

Sprokit has some cmake support code that originates when sprokit was a stand-alone package. Now that it is part of kwiver, it should use that CMake support serivces. Having a duplicate set in sprokit is confusing and causes maintenance problems.


linus-sherrill:
Should support for pkgconfig be removed? There is incomplete support in sprokit and nowhere else. It is apparently not used by our customers.

dstoup:
Is this PR still relevant? I know there was some recent work that may have done a little of it. Might be worth rebasing on master and getting this work landed.

linus-sherrill:
I think some of this work was also done by a Python branch. It would be good to rebase on master and see if any changes are needed.

PR #984, opened 2020-04-27
Downgrade message to Warning.

linus-sherrill:
This message should be a warning since it can proceed with the default logger when the specified back end can not be loaded. The "ERROR" in the message would cause testing failures due to the testing framework looking for that keyword.


dstoup:
Sorry for the delay, I'm just getting done with meetings for the day. Thanks for looking into this!

PR #992, opened 2020-05-08
Fix: INTERFACE_LIBRARIES doesn't have LINK_LIBRARIES property

danlipsa:


mleotta:
@danlipsa what were failures you noticed that prompted this change? 

Since this is a bug fix, we probably want to rebase this on the `release` branch and target for merging into `release`.

danlipsa:
The message was that an interface library (Threads::Threads) does not have the LINK_LIBRARIES property. This was when adding kwiver_algo_cuda plugin.
Not sure why this does not occurs for you. Either cuda version or cmake version (I use the latest) may make a difference.

I'll rebase and target for merging for release.

danlipsa:
it was an error message at configure time saying that INTERFACE_LIBRARIES don't have LINK_LIBRARIES properties. I can print which is that interface library. Not sure why you are not seeing this. It may be either from cuda 10 that uses an interface library or from cmake I am using 3.17

PR #993, opened 2020-05-09
Fix: INTERFACE_LIBRARIES doesn't have LINK_LIBRARIES property

danlipsa:


danlipsa:
This is the same as 
https://github.com/Kitware/kwiver/pull/992
targeted for release

PR #994, opened 2020-05-14
Add dep on config with vital-config

dstoup:
This PR fixes a sporadic failure in Visual Studio, where the vital_config library tries to build before the vital/version.h is configured and 'installed'. It sets a dependency on the vital_config to the kwiver_config custom target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,19 +224,6 @@ endif ()
 #---------------------------------------------------------------------
 #
 
-# Use Git (if available) to add Git hash info to the version header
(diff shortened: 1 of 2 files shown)


dstoup:
@tao558 This the fix to vital/version.h not found that we were looking into.

dstoup:
Jenkins test this please

dstoup:
Any objections? Dashboards pass, so I would like to merge this soon so tonight's will be clean.

tao558:
I'm just catching up now, but I have no objections.

Issue #757, opened 2019-02-25
Remove algorithm_impl


mleotta:
@as6520 pointed out that it looks like the `algorithm_impl` class is no longer needed:

https://github.com/Kitware/kwiver/blob/master/vital/algo/algorithm.h#L346

From what I can see I agree.  This appears to be a relic of the old days before the new plugin loader when each algorithm needed to be able to clone itself.  Now that the `clone()` function is removed the `algorithm_impl` is empty and useless.  It makes things more complicated and should be removed.

We should removed `algorithm_impl` and make all algorithms derive directly from `algorithm_def`.  For deeper class hierarchies, like the OpenCV detectors, the child algorithms should be able to derive directly from their intended parent without an `algorithm_impl` injected in the middle.

mleotta:
@kfieldho We discussed this earlier.  @linus-sherrill should probably confirm that this change makes sense, but he doesn't need to be the one to make the change.

linus-sherrill:
I agree that the `algorithm_impl` is no longer needed.

linus-sherrill:
Addressed in PR #997

PR #997, opened 2020-05-18
Removing algorithm_impl template.

linus-sherrill:
This change addresses issue #757 where this class has been deemed unnecessary.

This template class was left behind but marked as deprecated so as not to break other projects.


dstoup:
These changes looks right to me.

Issue #583, opened 2018-08-02
Add KWIVER_PYTHON_MAJOR_VERSION to the kwiver-config.cmake


aaron-bray:
It is defaulted to 2, my custom application is setup for 3
(I set fletch to 3, and kwiver to 3)

But when my application does the find_package(kwiver) 

I get this error:

 -- KWIVER_PYTHON_MAJOR_VERSION = 2
5>  -- PYTHON_VERSION = 3.6
5>  CMake Error at C:/Programming/builds/project_d/external/kwiver/CMake/kwiver-setup-python.cmake:169 (message):
5>    Requested python "2" but got "3.6"
5>  Call Stack (most recent call first):
5>    C:/Programming/builds/project_d/external/kwiver/CMake/kwiver-utils.cmake:10 (include)
5>    CMakeLists.txt:67 (include)

To fix this, I provide this variable to my ExternalProject_Add

CMAKE_CACHE_ARGS
    -Dkwiver_DIR:PATH=${kwiver_DIR}
    -DKWIVER_PYTHON_MAJOR_VERSION:STRING=3

this should really be added to kwiver-config with what version kwiver was built with

PR #999, opened 2020-05-19
Add python major version to kwiver-config.cmake

linus-sherrill:
Add KWIVER_PYTHON_MAJOR_VERSION to the kwiver-config.cmake file. This fix addresses issue #583 for interfacing with the kwiver build and install products.


dstoup:
The change looks great. Is major version sufficient or will we need minor version at some point? Not sure why we would really, major should be enough to use the correct FindPython{2|3} but figured it was worth the ask.

linus-sherrill:
I thought about adding the complete python version, but the issue only mentioned the major version. It can be added. What is the consensus?

linus-sherrill:
I added the full PYTHON version, as it may be useful in the future.

mleotta:
This is sort of a bug fix.  We are preparing an release very soon.  I'm wondering if this change belongs their too.

PR #1001, opened 2020-05-21
Convert test SEGFAULT to failure.

linus-sherrill:
These changes check for the failure condition in the ffmpeg tests and just fail the test rather than dropping core.


arrows/tests/test_video_input.h
```diff
@@ -176,6 +176,8 @@ void test_seek_frame( kwiver::vital::algo::video_input& vi )
 
     auto img = vi.frame_image();
 
+    ASSERT_TRUE( !! img );
```
dstoup, line 179: Looks good to me though I don't understand that syntax....

PR #1006, opened 2020-06-02
Test PR over python overhaul that is rebased to check for CI pass/fail

Purg:
This is the result of a `rebase -i --rebase-merges --onto master f75454864 dev/overhaul-python-crall` from the `dev/overhaul-python-crall` branch in order to linearize the messy history of the branch (which included 4 different master merged from different points in time).

This PR is not intended to be merged per say but to test out that this did not explode CI before considering jumping to this result.


Erotemic:
So I understand, the idea is we want to linearize the history, and this is testing if that is feasible?

as6520:
Yeah originally that was the plan. Since dev/overhaul has landed I believe the changes in the branch are already in master.

PR #1016, opened 2020-07-02
Use an avfilter pipeline in the FFMPEG reader

mleotta:
My motivation for this change was to restore the deinterlacing option that we used to have with older versions of FFMPEG.  This is important for properly reading interlaced video like the VIRAT data.  In modern FFMPEG there is no explicit deinterlacing functions.  Instead there is a generic filtering pipeline with `libavfilter` that includes several different deinterlacing filters to choose from.  The filter pipeline is very similar in concept to Sprokit, but focused on audio and video data only.  So with this change we can now inject  "yadif" (yet another deinterlacing filter) into the decoder which will deinterlace only if interlacing is detected.

This change is much more powerful than just deinterlacing.  The avfilter pipeline has a large pool of filters to choose from (blurring, cropping, flipping, color correction, etc).  You can now easily add a pipeline with any number of these features by setting the `filter_desc` string in the ffmpeg_video_input config.  The video will be filtered as it comes out of the video reader.  For example `yadif=dint=1, hflip, gblur=sigma=3` will deinterlace (the `dint=1` means only if needed), horizontally flip the image, and apply a Gaussian blur with sigma of 3.  We can do all of these things with VXL or OpenCV after reading the image, but we haven't wrapped all of these things as KWIVER algorithms, so the avfilter may be more convenient.

The current limitation of feature is that we are only supporting filters that operate on each frame independently.  Any filters that combine multiple frames or change the temporal frame sampling will not currently work correctly.  These things conflict with our design that focuses on frame accurate video seeking.

### DIFF
--- a/CMake/FindFFMPEG.cmake
+++ b/CMake/FindFFMPEG.cmake
@@ -64,6 +64,13 @@ else()
     /usr/local/lib64
   )
 
+  find_library( FFMPEG_avfilter_LIBRARY avfilter
+    /usr/lib
+    /usr/local/lib
+    /usr/lib64
+    /usr/local/lib64
+  )
+
   find_library( FFMPEG_swscale_LIBRARY swscale
     /usr/lib
     /usr/local/lib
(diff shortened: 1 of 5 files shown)


mleotta:
There have been no objections in the last week, and all checks have passed, so I'm going to merge this.

PR #1019, opened 2020-07-08
Fix warnings caused by catching polymorphic type exception by value.

dstoup:


dstoup:
All CI passing, merging.

Issue #718, opened 2019-01-14
Serialization (Protobuf) of vital::image is incomplete/broken


mleotta:
The `vital::image` class has two modes of operation: It either owns it's data or it is a view into someone else's data.  If it is a view into someone else's data the `vital::image::memory()` will be NULL.  As written this will cause serialization to segfault because we are always serializing the `vital::image_memory` object, which is not guaranteed to exist.  The easiest fix here is to check if the memory is NULL and deep copy the data into a new image with owned memory before serializing the `vital::image::memory`.  Alternatively, you could avoid the deep copy by traversing the image and serializing directly, but that involves more code.

PR #1020, opened 2020-07-08
Provide consistent data references in ocv_to_vital's returned image

neal-siekierski:
From the commit:

> If the else block in `ocv_to_vital(const cv::Mat&, ColorMode)` (in `kwiver::arrows::ocv::image_container`) is triggered, then `img.data` no longer points to memory managed by the variable `memory`.  Therefore we use `img_clone.data` in that case instead.

I ran into this when running a debug build of TeleSculptor. As previously created, the returned `image` ran afoul of [this check](https://github.com/vxl/vxl/blob/0bb0ca92867408caec298cef05412ed85c6d56b7/core/vil/vil_image_view.hxx#L85-L86) in VXL because the first pixel pointer was out of range of the data array.

I'm not very familiar with KWIVER's image infrastructure, so it would be good if someone could double-check this.

### DIFF
--- a/arrows/ocv/image_container.cxx
+++ b/arrows/ocv/image_container.cxx
@@ -115,28 +115,25 @@ image_container
   // if the cv::Mat has reference counted memory then wrap it to keep a
   // counted reference too it.  If it doesn't own its memory, then the
   // vital image won't take ownership either
-  image_memory_sptr memory;
-  void* data;
+  cv::Mat img2;
 #ifndef KWIVER_HAS_OPENCV_VER_3
   if ( !img.refcount )
 #else
   if ( !img.u )
 #endif
   {
-    data = img.data;
-    memory = std::make_shared<mat_image_memory>(img);
+    img2 = img;
   }
   else
   {
-    cv::Mat img_clone = img.clone();
-    data = img_clone.data;
-    memory = std::make_shared<mat_image_memory>(img_clone);
+    img2 = img.clone();
   }
 
-  return image(memory, data,
-               img.cols, img.rows, img.channels(),
-               img.channels(), img.step1(), 1,
-               ocv_to_vital(img.type()));
+  image_memory_sptr memory = std::make_shared<mat_image_memory>(img2);
+  return image(memory, img2.data,
+               img2.cols, img2.rows, img2.channels(),
+               img2.channels(), img2.step1(), 1,
+               ocv_to_vital(img2.type()));
 }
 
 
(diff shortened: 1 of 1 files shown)


as6520:
@neal-siekierski I am not familiar with image vital type. @linus-sherrill Could you please suggest reviewers for the PR?

mleotta:
This change looks like the right fix to address this immediate bug, but looking at the code, I wonder if we should remove that `else` case instead.  It looks like `else` case was added a couple of years ago by @as6520 as part of some serialization work.  I don't think the `else` case should be there.  The documentation of `ocv_to_vital` is as follows:
```
This function constructs a vital::image from a cv::Mat and wraps the same
memory.  If the memory is owned by the cv::Mat this function will use a
mat_image_memory class to retain the original memory and reference
counting.  If the cv::Mat does not own the memory, the vital::image will
point to the same memory but also not take ownership.  That is
vital::image::memory() will return nullptr.
```
The added `else` case instead makes a deep copy of the image which owns its own memory.  That is not what this function is supposed to do.  It should always make a shallow copy, which is fast.  If you need an image with memory management you can always deep copy either the cv::Mat before passing to this function or the vital::image that is returned by this function.  `ocv_to_vital` should allow you to get a converted view into existing memory whether it is owned or not.

mleotta:
I'd prefer to revert the change from 2 years ago that added the else condition, but we should hear from @as6520 about why it was added in the first place.  It's likely something else will have to fixed elsewhere if we remove the `else`.

In general, a user needs to understand that an image, whether cv::Mat or vital::image or vil_image, may own it's memory or it may be a view into someone else's memory.  If a case arises where an image that owns it's memory is needed, it's up to the user to check if the memory is owned and deep copy if not.

as6520:
Thanks for the reminder @mleotta,  I think you had captured this problem in #718. Based on https://github.com/Kitware/kwiver/commit/d9132a9a9a211c52aa1a8f70e4bcf89925724f35, this might not be an issue anymore since @linus-sherrill is making a copy of the image in serialization. So it might be okay to revert it.

mleotta:
By chance I just ran into this same bug in my own work.  I'm going to push up an alternate fix for this issue which I think is more correct.  There is another bug I found that was my fault (4 years ago), which probably lead @as6520 to add the `else` clause as a work around.  So in the end this was all my fault.

neal-siekierski:
> I'm going to push up an alternate fix for this issue which I think is more correct.

That sounds fine to me. Feel free to close this once that's up.

neal-siekierski:
Closing in favor of #1025.

PR #1022, opened 2020-07-10
Started next release

mleotta:


mleotta:
This is a trivial change that is part of the post-release book keeping.  We need to setup the release branch with new release notes and updated version number to start excepting patches in preparation for a patch release.

PR #1024, opened 2020-07-10
Dev/finish activity serialization

tao558:
This PR adds protobuf serialization for the vital type "activity". Tests for the JSON and protobuf serialization are also added.

In order for the JSON tests to pass, the implementations of load and save for an activity were modified. Activity has two fields that are pointers, and in the previous implementation, it was assumed that they were always non-null. When one of these fields was null, a null pointer was dereferenced which caused a seg-fault.

Other small improvements were made as well:
1.) In a few unrelated tests, iterators were not being incremented.
2.) Renamed a misnamed unrelated test

### DIFF
--- a/arrows/serialize/json/load_save.cxx
+++ b/arrows/serialize/json/load_save.cxx
@@ -664,17 +664,14 @@ void load( cereal::JSONInputArchive& archive,
   // Optional parameters aren't supported
   // for JSON. activity_type and participants may or may not exist,
   // so we check if an exception is thrown when we look for them.
-  // If an exception is thrown, we don't set those fields.
-  bool has_act_type = true;
-  bool has_participants = true;
-
+  // If an exception is thrown, we set those fields to null.
   try
   {
     load( archive, *act_type );
   }
   catch( cereal::Exception& )
   {
-    has_act_type = false;
+    act_type = nullptr;
   }
 
   try
@@ -683,26 +680,11 @@ void load( cereal::JSONInputArchive& archive,
   }
   catch( cereal::Exception& )
   {
-    has_participants = false;
+    participants = nullptr;
   }
 
-  if ( has_act_type )
-  {
-    activity.set_activity_type( act_type );
-  }
-  else
-  {
-    activity.set_activity_type( nullptr );
-  }
-
-  if ( has_participants )
-  {
-    activity.set_participants( participants );
-  }
-  else
-  {
-    activity.set_participants( nullptr );
-  }
+  activity.set_activity_type( act_type );
+  activity.set_participants( participants );
 }
 
 } // end namespace
--- a/arrows/serialize/json/load_save.cxx
+++ b/arrows/serialize/json/load_save.cxx
@@ -624,10 +624,15 @@ void save( cereal::JSONOutputArchive& archive,
     archive( ::cereal::make_nvp( "id", activity.id() ),
              ::cereal::make_nvp( "label", activity.label() ),
              ::cereal::make_nvp( "confidence", activity.confidence() ) );
-    save( archive, *activity.activity_type());
     save( archive, activity.start());
     save( archive, activity.end());
-    save( archive, *activity.participants() );
+
+    // These may be null
+    if ( activity.activity_type() )
+      save( archive, *activity.activity_type());
+
+    if ( activity.participants() )
+      save( archive, *activity.participants() );
 }
 
 void load( cereal::JSONInputArchive& archive,
@@ -636,7 +641,7 @@ void load( cereal::JSONInputArchive& archive,
   kwiver::vital::activity_id_t id;
   kwiver::vital::activity_label_t label;
   kwiver::vital::activity_confidence_t confidence;
-  kwiver::vital::class_map_sptr class_map =
+  kwiver::vital::class_map_sptr act_type =
                            std::make_shared< kwiver::vital::class_map >();
   kwiver::vital::object_track_set_sptr participants =
                            std::make_shared< kwiver::vital::object_track_set >();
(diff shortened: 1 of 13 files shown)


tao558:
Jenkins test this please

arrows/serialize/json/load_save.cxx
```diff
@@ -624,10 +624,15 @@ void save( cereal::JSONOutputArchive& archive,
...
              ::cereal::make_nvp( "label", activity.label() ),
              ::cereal::make_nvp( "confidence", activity.confidence() ) );
-    save( archive, *activity.activity_type());
     save( archive, activity.start());
     save( archive, activity.end());
-    save( archive, *activity.participants() );
+
+    // These may be null
+    if ( activity.activity_type() )
+      save( archive, *activity.activity_type());
```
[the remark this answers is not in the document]
tao558: Fixed in 55651293d54bc6f866d9db56f45c41d631739c0d

arrows/serialize/json/load_save.cxx
```diff
@@ -646,15 +655,54 @@ void load( cereal::JSONInputArchive& archive,
...
+    has_participants = false;
+  }
+
+  if ( has_act_type )
+  {
+    activity.set_activity_type( act_type );
+  }
+  else
+  {
+    activity.set_activity_type( nullptr );
```
as6520, line 695: Can we set activity type to nullptr [here](https://github.com/Kitware/kwiver/pull/1024/files#diff-09735ca8ba7ffc48e8004498c9f7d14bR677)?

arrows/serialize/json/load_save.cxx
```diff
@@ -646,15 +655,54 @@ void load( cereal::JSONInputArchive& archive,
...
+  }
+
+  if ( has_participants )
+  {
+    activity.set_participants( participants );
+  }
+  else
+  {
+    activity.set_participants( nullptr );
```
as6520, line 704: Can we set participants to nullptr [here](https://github.com/Kitware/kwiver/pull/1024/files#diff-09735ca8ba7ffc48e8004498c9f7d14bR686)?

vital/types/protobuf/activity.proto
```diff
@@ -0,0 +1,47 @@
...
+package kwiver.protobuf;
+
+import "class_map.proto";
+import "object_track_set.proto";
+import "timestamp.proto";
+
+message activity {
+  required int64 id = 1;
+  required string label = 2;
+  optional class_map classifications = 3;
```
as6520, line 42: Shouldn't this be called activity_type?

tao558: I don't believe so. This has the same fields as the activity vital type. activity_type was refactored out in favor of class_map, which already has protobuf bindings

arrows/serialize/json/load_save.cxx
tao558: Yes, great idea.

vital/types/protobuf/activity.proto
as6520: I was referring to `classifications`, should `classifications` be renamed to `activity_type`?

arrows/serialize/json/load_save.cxx
tao558: Fixed in 560077cce47e58452637078afd1b6d92ed82c0ef

arrows/serialize/json/load_save.cxx
tao558: Fixed in 560077cce47e58452637078afd1b6d92ed82c0ef

vital/types/protobuf/activity.proto
tao558: In that case yeah it definitely could be. I'll change it.

tao558: Fixed in e6f9cce61663bf4ac8cd3aa68cd96ab59c8beeac

arrows/serialize/json/tests/test_serialize.cxx
```diff
@@ -194,6 +339,8 @@ TEST( serialize, detected_object )
     {
       EXPECT_EQ( *(o_it->first), *(d_it->first) );
       EXPECT_EQ( o_it->second, d_it->second );
+      o_it++;
```
dstoup, line 342: Considering these are tests and are less subject to performance requirements, you can disregard this comment with regards to this PR, but in general, it's best to pre-increment when possible. Post-increment requires an extra copy which can get costly when the iterator objects are large or when there are a lot of them.

arrows/serialize/json/tests/test_load_save.cxx
```diff
@@ -71,6 +72,170 @@ int main(int argc, char** argv)
...
+  EXPECT_EQ( act.id(), act_dser.id() );
+  EXPECT_EQ( act.label(), act_dser.label() );
+  EXPECT_DOUBLE_EQ( act.confidence(), act_dser.confidence() );
+  EXPECT_EQ( act.start(), act_dser.start() );
+  EXPECT_EQ( act.end(), act_dser.end() );
+
+  // Check values in the retrieved class map
+  auto const act_type = act.activity_type();
+  auto const act_type_dser = act_dser.activity_type();
+  EXPECT_EQ( act_type->size(), act_type_dser->size() );
```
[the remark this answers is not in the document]
tao558: I did this intentionally to make it easier to read those 3 lines. The 2nd arguments are all lined up for those expect_double_eq calls. If you would still prefer I change it, I am happy to do so.

arrows/serialize/json/load_save.cxx
```diff
@@ -646,14 +655,35 @@ void load( cereal::JSONInputArchive& archive,
...
+
+  // Optional parameters aren't supported
+  // for JSON. activity_type and participants may or may not exist,
+  // so we check if an exception is thrown when we look for them.
+  // If an exception is thrown, we set those fields to null.
+  try
+  {
+    load( archive, *act_type );
+  }
+  catch( cereal::Exception& )
```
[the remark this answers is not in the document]
tao558: The latter. We're first checking whether those fields are set. If they're not, that signifies that they should be set to null. Fixed in c51b579f145749e25bac297306d50f8db025a625

arrows/serialize/json/tests/test_serialize.cxx
tao558: Good to know! This was easy enough to change in this case: a7e446f368e2708cefd14cdd367d795516350dda

tao558:
Jenkins test this please

PR #1028, opened 2020-07-15
Fix camera init early termination bug

mleotta:


mleotta:
The Windows failure seems unrelated

dstoup:
The outstanding Linux build did succeed. It just failed to report back.

PR #1029, opened 2020-07-17
Fix reliability of OpenCV Estimate PNP 

mleotta:
There was an outlier RANSAC loop around the OpenCV RANSAC PNP algorithm.  Letting OpenCV handle the RANSAC by itself produces better results and the code is much simpler.  Run time seems about the same.


linus-sherrill:
Looks ok to me - pending CI success

dstoup:
The outstanding Mac build did succeed but failed to report back.

Issue #1021, opened 2020-07-10
KWIVER 1.5 build failure on windows


aaron-bray:
I had to add :
`#include <stdexcept>`
here:
https://github.com/Kitware/kwiver/blob/5a7a959a1df79d1657774b36d2c35be5884850c6/vital/types/track.cxx#L39

to get kwiver 1.5 to compile on windows MSVC 2019 (16.6.2)
due to :
`KWIVER\src\vital\types\track.cxx(138,16): error C2039: 'logic_error': is not a member of 'std'`

Note, I built both the basic KWIVER (no added cmake options) and with ARROWS and PROJ

Issue #1035, opened 2020-07-30
MSVC 2019 cannot compile variant.hpp


mleotta:
When building sprokit in MSVC 2019 there are numerous errors generated in `vital/internal/variant/variant.hpp`.  It appears that the `MPARK_CPP14_CONSTEXPR` define evaluates to `true` on this compiler, but it is not able to correctly compile the C++14 code in the `#ifdef`.  I assume earlier MSVC compilers used the `#else` case.

It looks like this file was ported from Boost.  Maybe we need to get an updated copy from a newer release of Boost?

Examples of errors:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C3546	'...': there are no parameter packs available to expand (compiling source file C:\Users\matt.leotta\source\repos\kwiver\sprokit\src\sprokit\pipeline_util\pipe_parser.cxx)	sprokit_pipeline_util	C:\Users\matt.leotta\source\repos\kwiver\vital\internal\variant\variant.hpp	389	
Error	C2955	'kwiver::vital::detail::traits': use of class template requires template argument list (compiling source file C:\Users\matt.leotta\source\repos\kwiver\sprokit\src\sprokit\pipeline_util\pipe_parser.cxx)	sprokit_pipeline_util	C:\Users\matt.leotta\source\repos\kwiver\vital\internal\variant\variant.hpp	389	
Error	C2119	'<range>$L0': the type for 'auto &&' cannot be deduced from an empty initializer (compiling source file C:\Users\matt.leotta\source\repos\kwiver\sprokit\src\sprokit\pipeline_util\pipe_parser.cxx)	sprokit_pipeline_util	C:\Users\matt.leotta\source\repos\kwiver\vital\internal\variant\variant.hpp	389	
Error	C2672	'begin': no matching overloaded function found (compiling source file C:\Users\matt.leotta\source\repos\kwiver\sprokit\src\sprokit\pipeline_util\pipe_parser.cxx)	sprokit_pipeline_util	C:\Users\matt.leotta\source\repos\kwiver\vital\internal\variant\variant.hpp	389	
Error	C2893	Failed to specialize function template 'unknown-type std::begin(_Container &)' (compiling source file C:\Users\matt.leotta\source\repos\kwiver\sprokit\src\sprokit\pipeline_util\pipe_parser.cxx)	sprokit_pipeline_util	C:\Users\matt.leotta\source\repos\kwiver\vital\internal\variant\variant.hpp	389	
[...]

PR #1036, opened 2020-07-30
Add missing header needed for MSVC 2019

mleotta:
This fix addresses #1021


mleotta:
@dstoup a trickier MSVC 2019 issue is #1035 if you want to tackle that instead :)

dstoup:
The Window build succeeded but failed to report back.

PR #1044, opened 2020-07-31
Add small patch to variant.hpp to fix VS 2019 builds

dstoup:
Copying the variant.hpp file caused issues with Linux. I believe this small patch works on VS 2019


mleotta:
I've tested on VS 2019, and it works for me.  Thanks!

PR #1053, opened 2020-08-14
Let kwiver recognize Visual Studio 2019 in its setup_KWIVER.bat file.

dstoup:
Kwiver needs to recognize Visual Studio 2019 or it sets up the wrong PATH in the setup_KWIVER.bat file. That causes some of the dashboard tests to fail to build because they can't find OpenCV.


dstoup:
Putting this on release

PR #1054, opened 2020-08-14
Kwiver support vs2019

dstoup:
Force kwiver to recognize Visual Studio 2019 and apply that to its setup_KWIVER.bat. Without it, some tests are failing to run during the build since the can't find OpenCV.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,9 +47,11 @@ mark_as_advanced(KWIVER_DEFAULT_LIBRARY_DIR)
 
 #
 # kwiver module config file.
-set(KWIVER_CONFIG_FILE          "${KWIVER_BINARY_DIR}/kwiver-config.cmake")
+set(KWIVER_CONFIG_FILE "${KWIVER_BINARY_DIR}/kwiver-config.cmake")
 if(CMAKE_CXX_COMPILER_ID MATCHES MSVC)
- if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.10)
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.20)
+    set(_vcVersion vc16)
+  elseif(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.10)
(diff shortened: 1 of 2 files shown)


mleotta:
@dstoup even with this change I'm having trouble getting KWIVER to build all the GTEST tests in VS2019.  It works for me in VS2017.  Are you having this issue?  It's the same old problem of KWIVER not having the environment set to be able to run the test executables, which need to be run at build time because of the way we use GTEST.

I know @aaron-bray did something a while back that made this work, at least for VS2017.  I'm wondering if that needs to be updated.

dstoup:
> @dstoup even with this change I'm having trouble getting KWIVER to build all the GTEST tests in VS2019. It works for me in VS2017. Are you having this issue? It's the same old problem of KWIVER not having the environment set to be able to run the test executables, which need to be run at build time because of the way we use GTEST.
> 
> I know @aaron-bray did something a while back that made this work, at least for VS2017. I'm wondering if that needs to be updated.

They all build for me. Here are the caveats:
I tested the change on master for testing, not release, even though I targetted release for the PR.
I always have to do some awful trickery to make any of the GTest tests build on any VS version. I run CMake config and generate, then I open cmd and run setup_KWIVER.bat to get the env correct then run cmake-gui from that shell, reconfig/regenerate (not sure if I really need to do that), and use the "Open Project" button to bring up Visual Studio. That seems to get the env set up correctly in VS as well.

But, I hate it and wish there was something better we could do. It would be great to hear from @aaron-bray  about it. I know @mwoehlke-kitware spent some time with Brad but, I think was told there wasn't much that could be done in CMake about it.

FWIW, we do something similar to that trick to make the Jenkins dashboards work on the tests. Basically it amounts to running cmake once to generate the setup script, calling it, then proceeding with the usual CTest script.

dstoup:
@mleotta  I should mention ... even though it's not related to the GTest tests directly, I can't build kwiver on VS 2019 at all without my Fletch branch that patches Darknet for pthread support. It was the dashboard failures on that PR that led me to this issue.

PR #691, opened 2018-12-14
Add PyTorch processes with descriptors and TUT tracker

mattdawkins:

PR #1055, opened 2020-08-17
Add tests for dev/add-pytorch-arrow (#691)

neal-siekierski:
This PR has the minimal set of tests I've created for #691. I've created a separate PR to make keeping track of any discussion easier.

All the tests do is make sure that the four processes added by #691 can at least run on dummy input. I'd appreciate some quick feedback on a couple things:
- @as6520 or @tao558, are things correctly set up for the tests to be run?
- @mattdawkins, the `try`-`except` blocks around the processes' `._step` implementations make it harder for the tests to detect errors. I recall you being fine with my removing it in `srnn_tracker.py` at another point; would there be any issue with doing so here as well? (I might elect to keep them around `srnn_tracker.py` though in order to make integrating the more recently released changes easier.)

Thanks!

### DIFF
--- a/python/kwiver/sprokit/processes/pytorch/alexnet_descriptors.py
+++ b/python/kwiver/sprokit/processes/pytorch/alexnet_descriptors.py
@@ -159,7 +159,8 @@ class AlexNetDescriptors(KwiverProcess):
             print( repr( e ) )
             import traceback
             print( traceback.format_exc() )
-            #sys.stdout.flush()
+            sys.stdout.flush()
+            raise
 
     def __del__(self):
         print('!!!!alexnet tracking Deleting python process!!!!')
--- a/python/kwiver/sprokit/processes/pytorch/resnet_augmentation.py
+++ b/python/kwiver/sprokit/processes/pytorch/resnet_augmentation.py
@@ -282,7 +282,8 @@ class DataAugmentation(KwiverProcess):
             print( repr( e ) )
             import traceback
             print( traceback.format_exc() )
-            #sys.stdout.flush()
+            sys.stdout.flush()
+            raise
 
     def __del__(self):
         print('!!!!Deleting augmentation python process!!!!')
--- a/python/kwiver/sprokit/processes/pytorch/resnet_descriptors.py
+++ b/python/kwiver/sprokit/processes/pytorch/resnet_descriptors.py
@@ -163,7 +163,8 @@ class ResnetDescriptors(KwiverProcess):
             print( repr( e ) )
             import traceback
             print( traceback.format_exc() )
-            #sys.stdout.flush()
+            sys.stdout.flush()
+            raise
 
     def __del__(self):
         print('!!!!Resnet tracking Deleting python process!!!!')
--- a/python/kwiver/sprokit/processes/pytorch/srnn_tracker.py
+++ b/python/kwiver/sprokit/processes/pytorch/srnn_tracker.py
@@ -427,7 +427,8 @@ class SRNNTracker(KwiverProcess):
             print( repr( e ) )
             import traceback
             print( traceback.format_exc() )
-            #sys.stdout.flush()
+            sys.stdout.flush()
+            raise
 
 
 # ==================================================================
--- a/python/kwiver/sprokit/tests/processes/CMakeLists.txt
+++ b/python/kwiver/sprokit/tests/processes/CMakeLists.txt
@@ -17,3 +17,7 @@ kwiver_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/pythonpath_test_process.py
 kwiver_add_python_module(${CMAKE_CURRENT_SOURCE_DIR}/pythonpath_test_scheduler.py
   sprokit/tests/processes
   pythonpath_test_scheduler)
+
+if(KWIVER_ENABLE_PYTORCH)
+  add_subdirectory( pytorch )
+endif()
--- a/python/kwiver/sprokit/tests/processes/pytorch/CMakeLists.txt
+++ b/python/kwiver/sprokit/tests/processes/pytorch/CMakeLists.txt
@@ -1,0 +1,7 @@
+kwiver_add_python_module( ${CMAKE_CURRENT_SOURCE_DIR}/__init__.py
+  sprokit/tests/processes/pytorch
+  __init__ )
+
+kwiver_add_python_module( ${CMAKE_CURRENT_SOURCE_DIR}/test_processes.py
+  sprokit/tests/processes/pytorch
+  test_processes )
--- a/python/kwiver/sprokit/tests/processes/pytorch/test_processes.py
+++ b/python/kwiver/sprokit/tests/processes/pytorch/test_processes.py
@@ -1,0 +1,154 @@
+"""Tests for the processes in kwiver.sprokit.processes.pytorch
+
+As a minor issue, all the code to be tested here is in KWIVER
+processes, which would make KWIVER's embedded pipeline functionality
+the preferred way to test them.  However, said functionality isn't yet
+wrapped for Python, so we're left invoking "kwiver runner" as a
+subprocess.
+
+"""
+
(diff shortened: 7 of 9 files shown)


neal-siekierski:
@mattdawkins, to summarize what you told me yesterday (let me know if I've mixed anything up), the reason the `._step()` methods have their bodies wrapped in a `try`-`except` block is that otherwise any errors would effectively stall the pipeline and the traceback would never get printed (at least on Windows), right?

I've been able to reproduce something similar on Linux (under certain conditions) and plan to make an issue for it. As I understand it, in order to support testing while making sure the tracebacks get printed, it would be a sufficient workaround to flush `stdout` and then re-raise the exception. Thoughts?

tao558:
Regarding the tests, I think the 

> This PR has the minimal set of tests I've created for #691. I've created a separate PR to make keeping track of any discussion easier.
> 
> All the tests do is make sure that the four processes added by #691 can at least run on dummy input. I'd appreciate some quick feedback on a couple things:
> 
> * @as6520 or @tao558, are things correctly set up for the tests to be run?

Looks good to me. These are set up correctly to run nosetests, and probably pytest as well. I'd just make sure that the dependent modules are in the requirements.txt and setup.py.

@as6520 any other thoughts?

as6520:
One minor concern, the tests associated with sprokit are present in https://github.com/Kitware/kwiver/tree/master/python/kwiver/sprokit/tests, should these tests be moved to https://github.com/Kitware/kwiver/tree/master/python/kwiver/sprokit/tests/processes?

neal-siekierski:
Thanks for the feedback!

Regarding what @tao558 wrote:

> I'd just make sure that the dependent modules are in the requirements.txt and setup.py.

Good catch. The requirements are `numpy`, `scipy`, `torch`, `torchvision`, and `opencv-python`. `numpy` is already listed in `requirements.txt` and `setup.py`, `scipy` is listed only in `requirements.txt` (where it isn't pinned), and the others aren't currently there.

I could just go and add the missing ones, but an unconditional dependency on `torch` in particular strikes me as undesirable, as KWIVER has plenty of functionality without the new, dependent code and the current Linux binaries are over 700 MB (see [here](https://pypi.org/project/torch/#files)).

There's the "extras" feature of `setuptools` ([link](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)), which might help here. However, enabling or disabling PyTorch support requires changing the value `KWIVER_ENABLE_PYTORCH` on the CMake side, and I don't know if extras can change build flags. The new code could perhaps be changed to be compatible with runtime feature detection, which is what the intent of the "extras" mechanism seems to be. I'd appreciate others' thoughts on this.

Regarding what @as6520 wrote:

> One minor concern, the tests associated with sprokit are present in https://github.com/Kitware/kwiver/tree/master/python/kwiver/sprokit/tests, should these tests be moved to https://github.com/Kitware/kwiver/tree/master/python/kwiver/sprokit/tests/processes?

I agree it makes sense to move the tests to an existing location. I'll put them under `/python/kwiver/sprokit/tests/processes/pytorch`. (By the way, it looks like the `.../tests/processes` folder doesn't currently have any tests per se, only helper processes and such.)

@mattdawkins, if you don't object I'll also make the change I mentioned above (flush stdout and then reraise in the `except` block). I can reproduce a similar issue (on my Linux setup) without the `try`-`except` by manually inducing an error when stderr is not a TTY; reintroducing the `try`-`except` with the changes I mentioned prints the error.

neal-siekierski:
All right, I've performed the move. I've also made the `try`-`except` changes. (I have a version with only the move locally if we decide we don't want them.) If the new changes look good, then that should only leave the dependency issue.

as6520:
> Good catch. The requirements are numpy, scipy, torch, torchvision, and opencv-python. numpy is already listed in requirements.txt and setup.py, scipy is listed only in requirements.txt (where it isn't pinned), and the others aren't currently there.

I would recommend pinning torch and torchvision since mixing version tends to causes problems in my experience.  Although Fb is fairly good about releasing the version two together so that they don't go out of sync and cause problems. 

> There's the "extras" feature of setuptools (link), which might help here. However, enabling or disabling PyTorch support requires changing the value KWIVER_ENABLE_PYTORCH on the CMake side, and I don't know if extras can change build flags. The new code could perhaps be changed to be compatible with runtime feature detection, which is what the intent of the "extras" mechanism seems to be. I'd appreciate others' thoughts on this.

The cmake args are mainly added in setup.py ([eg](https://github.com/Kitware/kwiver/blob/master/python/setup.py#L65)) or can be provided as command line parameters when the wheel is being built. However after the build, the wheel is essentially a tar file. So disabling torch if it is required in the wheel would work. We can try to use extras feature, however if the arrow is present in then it might crash due to import errors when kwiver tools are used. Thus I would recommend adding the two packages to requirements.

An alternate solution would be to create an independent package for the tracker ([eg](https://github.com/as6520/external_arrow)), publishing it on pypi ([same package](https://pypi.org/project/external-arrow/)) and registering the arrow as an [entrypoint](https://github.com/as6520/external_arrow/blob/master/setup.py#L39). Thus torch would become the requirement for the package rather than the kwiver. However I don't think that would be applicable here.

neal-siekierski:
> I would recommend pinning torch and torchvision since mixing version tends to causes problems in my experience.

I agree that pinning packages generally makes sense. (I also mentioned it partially to note the inconsistency between `requirements.txt` and `setup.py` as they currently are in `master`.)

> The cmake args are mainly added in setup.py ([eg](https://github.com/Kitware/kwiver/blob/master/python/setup.py#L65)) or can be provided as command line parameters when the wheel is being built. However after the build, the wheel is essentially a tar file. So disabling torch if it is required in the wheel would work.

That's about what I expected, thanks.

> We can try to use extras feature, however if the arrow is present in then it might crash due to import errors when kwiver tools are used. Thus I would recommend adding the two packages to requirements.

Working around the import errors is essentially what I had in mind when I wrote this:

> The new code could perhaps be changed to be compatible with runtime feature detection, which is what the intent of the "extras" mechanism seems to be.

But are import errors actually a concern?  There are no imports elsewhere of `kwiver.sprokit.processes.pytorch`, and as I recall plugin discovery can deal with an `ImportError`.  If they're not a problem, then the wheel could always include this code and we could make direct use of extras.  The remaining issue would be making sure the tests are only run when `torch` etc. are available.  `pytest` and `nose` appear to handle skipping tests in different ways ([`pytest`](https://docs.pytest.org/en/stable/skipping.html#skipping-test-functions), [`nose`](https://nose.readthedocs.io/en/latest/plugins/skip.html)), though; were you trying to support both?

If supporting extras ends up looking to be too much of a hassle, though, adding the various unconditional dependencies seems fine for now.

> An alternate solution would be to create an independent package for the tracker ([eg](https://github.com/as6520/external_arrow)), publishing it on pypi ([same package](https://pypi.org/project/external-arrow/)) and registering the arrow as an [entrypoint](https://github.com/as6520/external_arrow/blob/master/setup.py#L39). Thus torch would become the requirement for the package rather than the kwiver. However I don't think that would be applicable here.

Making the code here into a separate package (either by letting this repository build multiple packages or extracting the code to a separate repository) indeed strikes me as beyond the scope of this and the original PR, though it's good to keep such things in mind for the future.

neal-siekierski:
I've gone ahead and added the other dependencies unconditionally, including adding `scipy` to `setup.py`. I've pinned all the new entries in `requirements.txt` to the versions I've been using locally. I've also added the relevant CMake flag to `setup.py`.

I think that takes care of everything for now. @as6520, did I miss anything? If not, once CI runs I'll add this to #691, add release notes, and look to merge.

as6520:
LGTM

as6520:
btw does this depend on #691?

neal-siekierski:
> LGTM

Great, thanks!

> btw does this depend on #691?

It's set to merge into #691's branch, so it will become a part of it.

PR #1058, opened 2020-08-19
Add explicit equivalent operator for iterator to fix windows build

mattdawkins:
I get windows builds without this "more than one operator != matches these operands" due to lack of an explicit operator== on iterator type

### DIFF
--- a/vital/iterator.h
+++ b/vital/iterator.h
@@ -151,6 +151,36 @@ public:
   }
 
   /**
+   * Equality operator
+   *
+   * Two iterators are considered equal if their current iteration references
+   * are equal in value or both iterators in a past-end state. If one
+   * iterator is in a past-end state and the other is not, they two iterators
+   * will never be equal.
+   *
+   * \returns If the two iterators are equal to each other.
+   */
+  virtual
+  bool equal_to( base_iterator<T, Tb> const & other ) const
+  {
+    // If both iterators are at their ends, they are equal.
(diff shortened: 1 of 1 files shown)


vital/iterator.h
```diff
@@ -160,26 +160,43 @@ class base_iterator
    *
    * \returns If the two iterators are equal to each other.
    */
-  friend bool operator==( base_iterator<T, Tb> const & lhs,
-                          base_iterator<T, Tb> const & rhs )
+  virtual
+  bool equal_to( base_iterator<T, Tb> const & other ) const
```
Purg, line 164: Would it make sense to make this private or protected? I could see it as redundant or confusing if there are multiple avenues exposed for checking equality.

mattdawkins: I'm fine with making it protected, the derived class needs to assess it to get around another ambigious overload of operator==

dstoup:
You get this Windows build error on master? I would be interested in knowing what configuration gives you this error since the dashboards are passing. It would be good to change or add dashboards to expose these kinds of issues.

mattdawkins:
VC command line on zeah when built in a super-build, I hit these with both master and viame/master, though I think there's multiple VS17 compiler versions on there

mattdawkins:
https://github.com/VIAME/VIAME/blob/master/cmake/build_server_windows_cpu.bat has explicit windows baths and toolkit it's using if that's helpful.

PR #1061, opened 2020-08-31
[VIAME] Merge pre-overhaul master

neal-siekierski:
@mattdawkins, this is the branch I mentioned. It merges 1a6759a851ef0dba91298ab696c8bdd3767390c6, the last commit to `master` before the Python overhaul branch was merged, into `viame/next`. Some things to note:
- I did this in stages, so there are multiple merge commits. Each merge was either clean or merged only one conflicting upstream merge commit (usually a PR to `master`). This was to make it easier for me, but it should also make it easier to check that each upstream change was handled correctly.
- Per https://github.com/Kitware/kwiver/pull/1061#issuecomment-685956890, I can build this with VIAME and run a couple examples, but it appears that changes will need to be made to VIAME.


mattdawkins:
I need to get a new release out this week I don't want to merge this for just because it would be too much risk, though after will look at it and test in dashboards

neal-siekierski:
I've been able to successfully build this with VIAME (an older version, VIAME/VIAME@8088545af77569c20ab9a140e835ab31f7dead20) and run the default detection pipeline from the annotation GUI, though it appears there's been some change to the build system as the modules (`viame_core_plugin.so`, `viame_hello_world_plugin.so`, and `viame_opencv_plugin.so`) are ending up directly in `install/lib` instead of `install/lib/viame/modules` like they used to (`.../viame/processes` is fine, though). If I work around that (by symlinking `viame_core_plugin.so` into a directory in `KWIVER_PLUGIN_PATH`), I can also run the `object_detection` example.

PR #1062, opened 2020-08-31
fix race condition by using mutex

judajake:


dstoup:
Jenkins test this please

PR #1064, opened 2020-09-01
add bounds checks

judajake:


dstoup:
Jenkins test this please

vital/kwiversys/CommandLineArguments.cxx
```diff
@@ -616,7 +616,17 @@ void CommandLineArguments::GenerateHelp()
       {
       str << std::endl;
       char argument[100];
-      sprintf(argument, "%s", sit->c_str());
+      if(sit->size() < 100-13)
```
linus-sherrill, line 619: I know this is a little late since this has been merged, but I see some things that could be improved.
1) Where does the "13" come from? It would be good to explain the origin of this magic number.
2) Why is the argument limited to 100 char? This is a modern system and this could be 1024 characters or more.
3) There is a lot of 'C' style length comparisons and other stringy logic. This could be eliminated by using a `std::stringstream` since the constructed string gets sent to a stream in the end.

PR #1067, opened 2020-09-02
Fortify fixes

judajake:


dstoup:
Jenkins test this please

dstoup:
These fixes are for the case where we are rightfully returning a size_t but GDAL uses unsigned, right?

vital/util/file_md5.cxx
```diff
@@ -64,7 +65,9 @@ file_md5( const string& fn )
     while ( is )
     {
       is.read( reinterpret_cast< char* >( buf ), bufsize );
-      if ( is.gcount() )
+      //gcount() should never be negative, but fortify does not seem to
```
dstoup, line 68: It's ok to keep this here, but kwiversys::MD5.c:md5_append:383 does the check and returns `if (nbytes <= 0)`

Edit* Noticed the diff and that we always had the if stmt here. This is fine.

vital/util/transform_image.h
```diff
@@ -90,9 +90,9 @@ void transform_image( image_of<T>& img, OP op )
   bool wBh = std::abs(img.w_step()) < std::abs(img.h_step()),
        dBh = std::abs(img.d_step()) < std::abs(img.h_step()),
        dBw = std::abs(img.d_step()) < std::abs(img.w_step());
-  size_t w_idx = ( ! wBh ) + dBw,
-         h_idx = wBh + dBh,
-         d_idx = ( ! dBw ) + ( ! dBh );
+  size_t w_idx = static_cast<size_t>( ! wBh ) + static_cast<size_t>(dBw),
```
dstoup, line 93: As we discussed, this change looks fine but we really should consider a more robust way of doing this.

dstoup:
I made the comments for the record. I think all of these fixes are relevant and correct. I'll merge if no one objects.

PR #1071, opened 2020-09-08
Fixed a bug in kwiver_add_library

mleotta:
Removed ${CMAKE_INSTALL_PREFIX} from destination paths.

The addition of this prefix was breaking some use of kwiver_add_library
in external projects that build on KWIVER.  Some CPack packagers require
these paths to be relative.  When absolute paths are needed, CMake will
automatically prepend CMAKE_INSTALL_PREFIX, so there should be no
need to make this explicit.


mleotta:
@tao558 Let me know if I'm missing something here about why CMAKE_INSTALL_PREFIX was added during the recent Python work.  I don't have a really deep understanding of how this works, but I do know that this change was breaking my TeleSculptor packaging and reverting it fixes the problem.

tao558:
> @tao558 Let me know if I'm missing something here about why CMAKE_INSTALL_PREFIX was added during the recent Python work. I don't have a really deep understanding of how this works, but I do know that this change was breaking my TeleSculptor packaging and reverting it fixes the problem.

This might be a better question for @as6520, as those changes were his. I'm not sure why that commit has me as both the author and committer.

I squashed adjacent commits by the same committer to make the dev/overhaul branch smaller. Looking at the branch again, it looks like only this commit didn't retain the original commiter. Sorry about that.

as6520:
I had added `CMAKE_INSTALL_PREFIX` for scikit-build, since the installation process required explicit path. Maybe we can do a conditional add based on the cmake variable SKBUILD so that the prefix is only added when scikit build is driving the build. Would that fix the build for telesculptor?

mleotta:
I think that would fix my issue, but this is already in a `if (NOT SKBUILD OR NOT target_type STREQUAL "STATIC_LIBRARY")` block.  So does this mean you have a case where you are building for scikit-build and you are building a dynamic library and that's when you need this absolute path?  I don't know anything about scikit-build, but [CMake docs for "install"](https://cmake.org/cmake/help/latest/command/install.html) say that these destinations are interpreted as relative to `CMAKE_INSTALL_PREFIX` if they are relative paths.  Having them relative allows them to be relocated (e.g. with DESTDIR).  Can you confirm that the explicit `CMAKE_INSTALL_PREFIX` is really needed for scikit-build on the current master branch?

as6520:
Sure let me verify if explicitly specifying `CMAKE_INSTALL_PREFIX` is required.

as6520:
An update with the build. I tried a couple of variants to check if CMAKE_INSTALL_PREFIX was explicitly required. When I used master to build the wheel, I was successful (with the caveat that ffmpeg arrow had to be disabled, it is failing due to an unrelated fPIC issue).  On this branch, I ran into the following error
```
  File "/usr/local/lib/python3.7/site-packages/skbuild/setuptools_wrap.py", line 586, in setup
    languages=cmake_languages
  File "/usr/local/lib/python3.7/site-packages/skbuild/cmaker.py", line 242, in configure
    CMaker.check_for_bad_installs()                                             
  File "/usr/local/lib/python3.7/site-packages/skbuild/cmaker.py", line 474, in check_for_bad_installs
    ("      " + _install) for _install in bad_installs)
  CMake-installed files must be within the project root.
      Violating Files:
      /vital/types/types.so
      /vital/types/activity.so
      /vital/types/bounding_box.so
      /vital/types/camera.so
      /vital/types/camera_intrinsics.so
      /vital/types/camera_map.so
      /vital/types/class_map.so
      /vital/types/color.so
      /vital/types/covariance.so
      /vital/types/descriptor.so     
      /vital/types/descriptor_set.so                                            
      /vital/types/detected_object.so
      /vital/types/detected_object_set.so
      /vital/types/eigen.so
      /vital/types/feature.so                                                   
      /vital/types/homography.so    
      /vital/types/homography_f2f.so
      /vital/types/landmark.so
      /vital/types/rotation.so
      /vital/types/similarity.so
      /vital/types/timestamp.so
      /vital/types/track.so
      /vital/types/track_set.so
      /vital/types/object_track_set.so
      /vital/config/config.so
      /vital/modules/modules.so
      /vital/algo/algorithm.so
      /vital/algo/algorithm_factory.so
      /sprokit/pipeline/datum.so
      /sprokit/pipeline/edge.so
      /sprokit/pipeline/pipeline.so
      /sprokit/pipeline/process.so
      /sprokit/pipeline/process_cluster.so
      /sprokit/pipeline/process_factory.so
      /sprokit/pipeline/scheduler.so
      /sprokit/pipeline/scheduler_factory.so
      /sprokit/pipeline/stamp.so
      /sprokit/pipeline/utils.so
      /sprokit/pipeline/version.so
      /sprokit/pipeline_util/bake.so
      /sprokit/pipeline_util/export_.so
      /sprokit/pipeline_util/load.so
      /arrows/serialize/json/json.so
```
This error occurs when I am explicitly specifying the CMAKE_INSTALL_PREFIX as a command line parameter. If I do not specify the parameter, the number  of violating files increase. Since only python bindings are present in the set of violating files, I can look further into this issue to see if CMAKE_INSTALL_PREFIX is being ignored by kwiver-utils-python.cmake. If that fix is successful, then we can revert `CMAKE_INSTALL_PREFIX`

mleotta:
@as6520 To me this looks like a bug in the `check_for_bad_installs` function in skbuild.  It explicitly searches and replaces `${CMAKE_INSTALL_PREFIX}` with the install prefix, but it does not check if the paths are relative.  When CMake finds a relative path it will automatically prepends `${CMAKE_INSTALL_PREFIX}`.  So relative paths should not be considered bad installs.  I'm not sure why the paths in the error message have a leading `/`.  That makes them look absolute.  I don't know if that `/` is coming from skbuild or from CMake.

Anyway, the workaround is probably to have a special case for skbuild in our code that uses `${CMAKE_INSTALL_PREFIX}` if building for skbuild and otherwise not.

as6520:
@mleotta I suspect that this might be coming from kwiver-utils-python.cmake since all the bad install seems to be python bindings. I am looking into it to verify if a workaround is required, however if this is critical for telesculptor, we can merge this branch and I can fix the scikit build issues in a followup branch.

mleotta:
@as6520 I've modified this PR so that it should work for both our uses.  If this works for you then let's get this merged.  If you later find a bug elsewhere that would allow you to simplify and reduce the special cases for skbuild, then we can apply that clean-up on a later branch.

as6520:
Sure I am building the wheel with the most recent commit.

as6520:
The build was successful. LGTM

as6520 (approved):

mleotta:
Great, I will merge once the Windows CI build finishes

PR #1080, opened 2020-09-21
Cleanup plugin path generation.

linus-sherrill:
The plugin path only points to the base directory for all the
plugin types. This shortens the path considerably.

### DIFF
--- a/arrows/core/applets/CMakeLists.txt
+++ b/arrows/core/applets/CMakeLists.txt
@@ -20,12 +20,6 @@ set( headers
 # Add our plugin path to the default list
 include_directories( ${CMAKE_CURRENT_BINARY_DIR} )
 
-###
-# Add path to applets so plugin loader can find them
-if(NOT SKBUILD)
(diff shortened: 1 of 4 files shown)


arrows/core/applets/CMakeLists.txt
```diff
@@ -20,12 +20,6 @@ set( headers
 # Add our plugin path to the default list
 include_directories( ${CMAKE_CURRENT_BINARY_DIR} )
 
-###
-# Add path to applets so plugin loader can find them
```
dstoup, line 24: Not sure about removing these SKBUILD blocks. Ameya may need to chime in, unless I'm missing that they moved somewhere else already. He put those in to control the wheel build.

dstoup (commented):
@as6520 Thoughts on the SKBUILD blocks?

arrows/core/applets/CMakeLists.txt
linus-sherrill: I think this may be a merge issue. The intent is to consolidate where and how the paths are calculated and communicated to the run-time into one place in the CMake file and in one include file.

linus-sherrill:
This branch has been merged into PR #1092

PR #1081, opened 2020-09-23
Remediating some warnings when using KWIVER_CPP_EXTRA option.

linus-sherrill:
Specifically addressing 
- unused parameters
- shadowed variables
- Missing case elements of a switch

### DIFF
--- a/vital/algo/bundle_adjust.cxx
+++ b/vital/algo/bundle_adjust.cxx
@@ -37,6 +37,7 @@
 #include <vital/algo/bundle_adjust.h>
 #include <vital/algo/algorithm.txx>
 #include <vital/logger/logger.h>
+#include <vital/vital_config.h>
 
 namespace kwiver {
 namespace vital {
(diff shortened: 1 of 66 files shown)


vital/algo/bundle_adjust.cxx
```diff
@@ -62,8 +62,8 @@ ::optimize(
   kwiver::vital::simple_camera_perspective_map &cameras,
   kwiver::vital::landmark_map::map_landmark_t &landmarks,
   vital::feature_track_set_sptr tracks,
-  const std::set<vital::frame_id_t>& fixed_cameras,
-  const std::set<vital::landmark_id_t>& fixed_landmarks,
+  const std::set<vital::frame_id_t>& /* fixed_cameras */,
```
dstoup, line 65: Why use VITAL_UNUSED in some places and commenting out in others?

dstoup (commented):
LGTM

vital/algo/bundle_adjust.cxx
linus-sherrill: Don't know why this was handled as a comment. Has been updated to use VITAL_UNUSED

linus-sherrill:
If this cleans up some classes of warnings, we can move some flags, like warnings for unused parameters and shadowing to the main set of options.

dstoup:
> If this cleans up some classes of warnings, we can move some flags, like warnings for unused parameters and shadowing to the main set of options.

I agree! I would be great to move more flags into the main set of warnings. Probably good to do it even if they aren't completely cleaned up with this PR.  Once they show up in the builds we can clean the rest up petty quickly.

PR #1085, opened 2020-09-30
Patch to support JSON serialization of boolean metadata

russelldj:
For reference, here is a snippet of the serialized data taken from [here](https://data.kitware.com/#item/56f580488d777f753209c72f).

`{
            "tag": 16,
            "type": "bool",
            "name": "Is frame a key frame",
            "value": false
        },
`

### DIFF
--- a/arrows/serialize/json/load_save_metadata.cxx
+++ b/arrows/serialize/json/load_save_metadata.cxx
@@ -91,10 +91,13 @@ struct meta_item
     {
       // bool metadata passes the is_integral() check but cannot be cast
       // to uint64_t
-      if ( trait.tag_type() == typeid( bool ) ) {
+      if ( trait.tag_type() == typeid( bool ) )
+      {
         const bool value = kwiver::vital::any_cast< bool > ( this->item_value );
         archive( CEREAL_NVP( value ) );
-      } else {
+      }
+      else
+      {
         const uint64_t value = kwiver::vital::any_cast< uint64_t > ( this->item_value );
         archive( CEREAL_NVP( value ) );
       }
--- a/arrows/serialize/json/load_save_metadata.cxx
+++ b/arrows/serialize/json/load_save_metadata.cxx
@@ -81,7 +81,7 @@ struct meta_item
     archive(  ::cereal::make_nvp( "type", item_value.type_name() ) );
     archive(  ::cereal::make_nvp( "name", trait.name() ) );
 
-    // This si a switch on the item data type
+    // This is a switch on the item data type
     if ( trait.is_floating_point() )
     {
       const double value = kwiver::vital::any_cast< double > ( this->item_value );
@@ -89,8 +89,15 @@ struct meta_item
     }
     else if ( trait.is_integral() )
     {
-      const uint64_t value = kwiver::vital::any_cast< uint64_t > ( this->item_value );
-      archive( CEREAL_NVP( value ) );
+      // bool metadata passes the is_integral() check but cannot be cast
+      // to uint64_t
+      if ( trait.tag_type() == typeid( bool ) ) {
+        const bool value = kwiver::vital::any_cast< bool > ( this->item_value );
+        archive( CEREAL_NVP( value ) );
+      } else {
+        const uint64_t value = kwiver::vital::any_cast< uint64_t > ( this->item_value );
+        archive( CEREAL_NVP( value ) );
+      }
     }
     else if ( trait.tag_type() == typeid( std::string ) )
     {
--- a/arrows/serialize/json/load_save_metadata.cxx
+++ b/arrows/serialize/json/load_save_metadata.cxx
@@ -145,9 +145,19 @@ struct meta_item
     }
     else if ( trait.is_integral() )
     {
-      uint64_t value;
-      archive( CEREAL_NVP( value ) );
-      this->item_value = kwiver::vital::any( value );
+      // is_integral() returns true for a bool, which needs to be handled differently
+      if ( trait.tag_type() == typeid( bool ) )
+      {
+        bool value;
+        archive( CEREAL_NVP( value ) );
+        this->item_value = kwiver::vital::any( value );
+      }
+      else
+      {
+        uint64_t value;
+        archive( CEREAL_NVP( value ) );
+        this->item_value = kwiver::vital::any( value );
+      }
     }
     else if ( trait.tag_type() == typeid( std::string ) )
     {
--- a/arrows/serialize/json/tests/test_load_save.cxx
+++ b/arrows/serialize/json/tests/test_load_save.cxx
@@ -227,7 +227,7 @@ kwiver::vital::metadata create_meta_collection()
   }
 
   {
-    const auto& info = traits.find( kwiver::vital::VIDEO_KEY_FRAME );
+    const auto& info = traits.find( kwiver::vital::VITAL_VIDEO_KEY_FRAME );
     auto* item = info.create_metadata_item( kwiver::vital::any(true ) );
     meta.add( item );
   }
--- a/arrows/serialize/json/tests/test_load_save.cxx
+++ b/arrows/serialize/json/tests/test_load_save.cxx
@@ -227,7 +227,7 @@ kwiver::vital::metadata create_meta_collection()
   }
 
   {
-    const auto& info = traits.find( kwiver::vital::VITAL_VIDEO_KEY_FRAME );
+    const auto& info = traits.find( kwiver::vital::VITAL_META_VIDEO_KEY_FRAME );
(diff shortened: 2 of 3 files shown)


linus-sherrill:
Does the protobuf implementation have the same problem?

arrows/serialize/json/load_save_metadata.cxx
```diff
@@ -81,16 +81,23 @@ struct meta_item
...
       const double value = kwiver::vital::any_cast< double > ( this->item_value );
       archive( CEREAL_NVP( value ) );
     }
     else if ( trait.is_integral() )
     {
-      const uint64_t value = kwiver::vital::any_cast< uint64_t > ( this->item_value );
-      archive( CEREAL_NVP( value ) );
+      // bool metadata passes the is_integral() check but cannot be cast
+      // to uint64_t
+      if ( trait.tag_type() == typeid( bool ) ) {
```
mleotta, line 94: KWIVER C++ style prefers the braces for scope, `{` and `}`, on new lines.

russelldj:
I have not gotten a chance to test protobuf yet, should I do that?

arrows/serialize/json/load_save_metadata.cxx
```diff
@@ -81,16 +81,23 @@ struct meta_item
...
+        const bool value = kwiver::vital::any_cast< bool > ( this->item_value );
+        archive( CEREAL_NVP( value ) );
+      } else {
+        const uint64_t value = kwiver::vital::any_cast< uint64_t > ( this->item_value );
```
mleotta, line 98: This may more of a question for @linus-sherrill, but do we need a special case for signed integers as well?  I expect this code may work as is, but the values of negative values may be written incorrectly in ASCII form in a JSON file since they will be cast to an unsigned int first.  We may even parse these value correctly if we read them back into an uint64 and then cast back to int, but I expect other JSON readers outside of KWIVER would not be so lucky.

linus-sherrill: The original concept for the uint64 as the metadata type was just to hold 64 bits of untyped data and delegate extracting (casting) the actual data type to the customer (as opposed to adding support for {uint,unt}{8,16,32,64} types). A cast to the desired signed/unsigned type should reinterpret the bits to get the correct value. The bit pattern get serialized/deserialized and ends up the same at the other end. (JSON should encode this as type uint64_t.) The JSON generated by the save() method only has to be decoded by the load() method. The JSON format used here was only designed to be a data marshaling format used to move data between endpoints, but could be created by save() and parsed by something other than the load() method.  I think the JSON encoding is self consistent in that it states the type is uint64, even though it may be interpreted as a different type further up the stack. 

I don't know if it is burdensome to get the uint64_t from the metadata_item and them cast it to the expected integer type, but we could add an additional templated accessor method which would include the needed casting.

linus-sherrill:
It would be great if you could look at the protobuf implementation to see if it has the same problem. We try to keep all/both serialization implementations synchronized.

arrows/serialize/json/load_save_metadata.cxx
mleotta: @linus-sherrill in this case we are specifically looking for a JSON representation of metadata that can be exported outside of KWIVER and used by other readers.  I agree that for protobuf we can encode things however we want because our protobuf reader is the only thing reading it.  However, JSON is generally intended to be both human readable and more portable across tools.  Can we updated the JSON reader to correctly read and write signed types? or does this JSON serializer need to be consistent with protobuf?

Note that I don't think we need to encode all combinations of bit depths for integers as long as we cast up to the bigger bit depth.  It's really just a signed/unsigned issue.  Negative numbers will not cast to meaningful values as uint64.  So maybe we need both uint64 and int64.

linus-sherrill: The JSON serializer does not need to be consistent with the protobuf format. The only hard requirement with the serializers is that deserializing serialized data returns the original value. I think that 

    int i = -1;
    unit64_t ui = static_cast<uint64_t>(i);
    // The unsigned representation for -1 is a really big number (0xFFFFFFFFFFFFFFFF)
    int j = static_cast<int>(ui);   // j should be -1
    int16_t k = static_cast<int16_t>(ui); // k should be -1

So it should be ok for the consumer to cast the result into the desired integer type with a static cast.

When converting signed to unsigned or a larger size integer, the high order sign bit is propagated through the additional high order bits. When converting to a integer of smaller size, the high order bits are truncated. This is not the case when converting floating point. I'm not sure if it works this way in python.

We could add a int64_t case to the serializer, although the only integer format that is currently used in the metadata is uint64_t. It is only by convention that a very limited set of types are used in the metadata.

mleotta:
@russelldj I changed the merge target here from `master` to `release`.  It looks like the code was already based on `release` so no rebasing was needed.  While there are other things proposed changes discussed in the comments, I think this change, as it is, addresses the immediate bug and should be merged first.  However, I see one potential issue.  It doesn't look like the JSON reader code has been updated to also support this new bool case.  We are now writing out bool values as "true" or "false" in the JSON, but still trying to read them back into uint64, which I expect will be problematic.

You should update the reader code with the analogous bool case.  I'd also recommend adding a unit test that reads and writes a bool value in JSON. It may need to test both true and false values.  Hopefully this is just a small change on existing unit tests, though I haven't looked to see what unit tests we have for this.

russelldj:
@mleotta, I added the reader. When looking at the unit tests, it does not appear that we test any other atomic types. Would you suggest creating a new test just for bools, or add it to some aggregate structure like the [metadata collection](https://github.com/Kitware/kwiver/blob/d84cf780aafd1a3208b03fc62da0051c88ae62e8/arrows/serialize/json/tests/test_load_save.cxx#L388)?

mleotta:
I don't think we need a separate test for each data type.  That is a good idea, but it probably a lot more work.  I do suggest that we have a test that writes and reads back in a metadata item or collection of metadata items and we should construction that test data in such a way that it contains all the core types, and possibly multiple examples covering different scenarios (e.g. both positive and negative values, true and false for bool).  It will be harder to determine what caused the failure when the test fails, but at least we will have more code coverage.

russelldj:
I will work on the approach you suggested, that makes a lot of sense to me. I don't fully understand it, but it appears that there's a pretty heavily-parameterized test for [covariance items](https://github.com/Kitware/kwiver/blob/d84cf780aafd1a3208b03fc62da0051c88ae62e8/arrows/serialize/json/tests/test_load_save.cxx#L803), which individually tests many it different situations. I don't know if I have the skill to replicate it (or if it's even possible for most/all atomic types) but maybe that's something interesting to think about longer term.

russelldj:
Apparently I didn't have tests built locally and am now having some issues with other ones (Ceres for now) building. Is it useful to build locally or should I be able to figure it out solely from travis outputs?

linus-sherrill:
I find it helpful to build and run the tests locally. It helps get through CI without too many cycles and sometimes smokes out collateral damage from your change.

russelldj:
Sorry, I shouldn't have deleted that comment. I found the mistake but that's still useful to know for the future.

arrows/serialize/json/tests/test_load_save.cxx
```diff
@@ -226,6 +226,12 @@ kwiver::vital::metadata create_meta_collection()
     meta.add( item );
   }
 
+  {
+    const auto& info = traits.find( kwiver::vital::VIDEO_KEY_FRAME );
```
dstoup, line 230: This is the line Travis is choking on. If you get it to build and repush, I will make sure the other CI is run as well.

dstoup (changes requested):

arrows/serialize/json/tests/test_load_save.cxx
mleotta: @russelldj It probably needs to be `kwiver::vital::VITAL_VIDEO_KEY_FRAME`.  The macros in metadata_tags.h prepend `VITAL_` to each of those identifiers when making the enum.

linus-sherrill: Has this been fixed?

russelldj: I've been having issues getting GTests to build locally and haven't had much time to dig into it. But Matt L. suggested I push up this change.

russelldj:
Jenkins test this please

mleotta:
@russelldj it looks like I had it wrong too.  The Travis CI build is telling us it should be `kwiver::vital::VITAL_META_VIDEO_KEY_FRAME`

russelldj:
Ack. I also should have caught that. Updated.

linus-sherrill:
Jenkins test this please

linus-sherrill (approved):
Pending a clean CI build

PR #1086, opened 2020-10-01
Make MVG arrow

mleotta:
This branch make a Multi-View Geometry (MVG) arrow (previously known as SFM), which will be the new home of 3D vision algorithms and applets that previously had gone into the core arrow.  This is the first step in cleaning up core, which has become overcrowded with a diverse mix of functionality.  Like core, the sfm arrow does not add new dependencies.  This new arrow just adds a better functional separation of code and allows more modularity for turning subsets of functionality on or off.

This initial branch is intentionally small and only moves one applet, one algo, and one utility from core.  This is to keep the code review simple and quick and to make sure the new arrow lands quickly.  Once this branch lands, additional branches will be submitted to move more functionality from core to mvg.


russelldj (approved):
I can't comment too much on organization but this all looks good to me. My one concern, which I believe you brought up previously, was that `sfm` is too narrow of a name. Some slightly broader ideas that come to mind are `core3d` (though I guess that clashes with the IARPA program), `kw3d`, `recon3d`,  `base3d`, and `3dmodel`.

mleotta:
Yeah, that's a good point.  I'm wondering now if "MVG" (multi view geometry) is a better more broad name than "SFM" (structure from motion).  However, I don't want people to confuse this with openMVG, which is another third party library that we could add an arrow for in the future.

Not everything in this library will be 3D, (e.g. feature tracking, homographies, fundamental matrices, etc.) but I think it will all relate to geometry across multiple views.

Anyone else have suggestions on the naming?

danlipsa (commented):
Looks good! +1

judajake (approved):

mleotta:
I've renamed this new arrow from SFM to MVG.  I think Multi-View Geometry better captures the scope of things that will go here, including feature point tracking, homography/fundamental/essential matrix estimation, multi-view stereo, depth map fusion, and, of course, structure from motion.  Not all of these things are strictly 3D vision, but they all involve geometric relations across multiple views.

PR #1088, opened 2020-10-04
Moved more classes from core to MVG arrow

mleotta:


mleotta:
This PR moves several more files from the core arrow to the MVG (Multi-View Geometry) arrow.  This branch is entirely about refactoring.  There are no algorithmic changes.  Mostly it consists of
 - Relocating files from `arrows/core` to `arrows/mvg`
 - Updating include paths across KWIVER to reflect these changes
 - Replacing `CORE` with `MVG` in include guards and export macros
 - Changing namespace of moved files to `kwiver::arrows::mvg`
 - Updating libraries to link to `kwiver_algo_mvg` as needed

In several cases there were issues with the include guards and namespace of the old files, so these were corrected as well.  For some reason much of the older code was in `kwiver::arrows` instead of `kwiver::arrow::core`.  Many unit tests outside of core also used these moved files in testing, which is why many files outside of core and MVG are also touched.

This is not all the files that should be moved from core to MVG.  I'm continuing to do this in batches to make the code review more manageable.

dstoup (approved):

russelldj (approved):

mleotta:
CI checks are all passing and I see three reviewer approvals, so I will merge.

PR #1089, opened 2020-10-04
Clean arrows library links

mleotta:
Most arrows were linking the core arrow (kwiver_algo_core) when they did not use any symbols from core arrow.  Instead they needed to only link to vital_algo.  In some cases private linkage to core was required, and in several cases only some tests needed to link to core.


dstoup (approved):

mleotta:
If we merge this one first I will resolve conflicts with #1088, which contains some similar but conflicting changes.

linus-sherrill (approved):

mleotta:
I think 2 reviews are enough.  I'm going to merge this.

PR #1092, opened 2020-10-06
Dev/search paths

linus-sherrill:
Centralize compile time search paths.
    
Move all compile time search paths into a single include file. The search paths for processes, clusters, and pipelines are consolidated into the kwiver-include-paths.h file. Previously there were three different include files and associated cmake support code. Also the handling of these search paths has been standardized, since they were not all the same.


linus-sherrill:
Jenkins, test this please

linus-sherrill:
Any comments? This has passed CI and is ready to merge.

mleotta:
I took a quick look and I think this is okay.  This is mostly impacting the location of sprokit config files, right?  I like the path structure of having different types of config files in different sub directories under "config".  We should probably move the existing files in "config" to "config/algo" and then very soon (next week) we will also want to add "config/applets" for applet default configurations.  These changes could happen in a separate PR.  We will also have to verify that configs can find each other for inclusion when in these different subdirectories.  Both applet and process configs may want to include algo configs.

linus-sherrill:
The sprokit default path for includes in pipe files used to be "${CMAKE_INSTALL_PREFIX}/share/sprokit/pipelines/include", but I don't think anything was ever put there. The cluster loader would look in "${CMAKE_INSTALL_PREFIX}/share/sprokit/pipelines/clusters", but we have not used clusters enough to actually have any in that directory. It made sense to pull these directories up to the top level.

PR #1095, opened 2020-10-07
Updated the config files to account for code moved from core to MVG

mleotta:


mleotta:
@judajake you will likely need these config updates for your applet

mleotta:
I've tested these config changes in TeleSculptor and they work as expected

PR #1096, opened 2020-10-07
Add init cameras landmarks

judajake:
Moved the init cameras and landmarks gui tool from TeleSculptor to kwiver

### DIFF
--- a/arrows/mvg/applets/init_cameras_landmarks.cxx
+++ b/arrows/mvg/applets/init_cameras_landmarks.cxx
@@ -75,31 +75,6 @@ typedef kwiversys::CommandLineArguments argT;
 
 kwiver::vital::logger_handle_t main_logger( kwiver::vital::get_logger( "init_cameras_landmarks" ) );
 
-//-------------------------------------------------------------------
-std::string
-frameName(kwiver::vital::frame_id_t frame,
-          kwiver::vital::metadata_vector const& mdv)
-{
-  using kwiver::vital::basename_from_metadata;
-  for (auto const& md : mdv)
-  {
-    if (md->has(kwiver::vital::VITAL_META_IMAGE_URI) ||
-      md->has(kwiver::vital::VITAL_META_VIDEO_URI))
-    {
-      return basename_from_metadata(md, frame);
-    }
-  }
-  return basename_from_metadata(nullptr, frame);
-}
-
-std::string
-frameName(kwiver::vital::frame_id_t frame,
-          kwiver::vital::metadata_sptr md)
-{
-  return kwiver::vital::basename_from_metadata(md, frame);
-}
-
-
 // ------------------------------------------------------------------
 kwiver::vital::config_block_sptr default_config()
 {
@@ -243,7 +218,7 @@ public:
     {
       return HELP;
     }
-    if ( cmd_args.count("config") )
+    if ( cmd_args.count("config") > 0 )
     {
       opt_config = cmd_args["config"].as<std::string>();
     }
@@ -330,6 +305,10 @@ public:
 
   void load_tracks( )
   {
+    if(config == nullptr)
+    {
+      return;
+    }
     kwiver::vital::path_t in_tracks_path =
       config->get_value<kwiver::vital::path_t>("input_tracks_file");
     feature_track_set_ptr =
@@ -391,18 +370,17 @@ public:
       GET_K_CONFIG(double, aspect_ratio),
       GET_K_CONFIG(double, skew));
 
-    auto baseCamera = simple_camera_perspective();
-    baseCamera.set_intrinsics(K);
+    auto base_camera = simple_camera_perspective();
+    base_camera.set_intrinsics(K);
     auto md = sfm_constraint_ptr->get_metadata()->metadata();
-    kwiver::vital::camera_map::map_camera_t camMap;
     if (!md.empty())
     {
-      std::map<frame_id_t, metadata_sptr> mdMap;
+      std::map<frame_id_t, metadata_sptr> md_map;
 
-      for (auto const& mdIter : md)
+      for (auto const& md_iter : md)
       {
         // TODO: just using first element of metadata vector for now
-        mdMap[mdIter.first] = mdIter.second[0];
+        md_map[md_iter.first] = md_iter.second[0];
       }
 
       bool init_cams_with_metadata =
@@ -413,7 +391,7 @@ public:
         auto im = first_frame;
         K->set_image_width(static_cast<unsigned>(im->width()));
         K->set_image_height(static_cast<unsigned>(im->height()));
-        baseCamera.set_intrinsics(K);
+        base_camera.set_intrinsics(K);
 
         bool init_intrinsics_with_metadata =
         config->get_value<bool>("initialize_intrinsics_with_metadata", true);
@@ -422,7 +400,7 @@ public:
           // find the first metadata that gives valid intrinsics
           // and put this in baseCamera as a backup for when
           // a particular metadata packet is missing data
-          for (auto mdp : mdMap)
+          for (auto mdp : md_map)
           {
             auto md_K =
               intrinsics_from_metadata(*mdp.second,
@@ -430,16 +408,18 @@ public:
                                        static_cast<unsigned>(im->height()));
             if (md_K != nullptr)
             {
-              baseCamera.set_intrinsics(md_K);
+              base_camera.set_intrinsics(md_K);
               break;
             }
           }
         }
 
         local_geo_cs lgcs = sfm_constraint_ptr->get_local_geo_cs();
-        camMap =
-          initialize_cameras_with_metadata(mdMap, baseCamera, lgcs,
+        kwiver::vital::camera_map::map_camera_t cam_map =
+          initialize_cameras_with_metadata(md_map, base_camera, lgcs,
                                            init_intrinsics_with_metadata);
+        camera_map_ptr =
+          std::make_shared<kwiver::vital::simple_camera_map>(cam_map);
 
         sfm_constraint_ptr->set_local_geo_cs(lgcs);
       }
@@ -448,7 +428,7 @@ public:
 
   bool write_cameras()
   {
-    std::string output_cameras_director =
+    std::string output_cameras_directory =
     config->get_value<std::string>("output_cameras_directory");
 
     for( auto iter: camera_map_ptr->cameras())
@@ -456,7 +436,7 @@ public:
       int fn = iter.first;
       camera_sptr cam = iter.second;
       std::string out_fname =
-        output_cameras_director + "/" + get_filename(fn) + ".krtd";
+        output_cameras_directory + "/" + get_filename(fn) + ".krtd";
       kwiver::vital::path_t out_path(out_fname);
       auto cam_ptr = std::dynamic_pointer_cast<camera_perspective>(cam);
       write_krtd_file( *cam_ptr, out_path );
@@ -532,12 +512,12 @@ public:
       auto mdv = videoMetadataMap->get_vector(frame_id);
       if (!mdv.empty())
       {
-        return frameName(frame_id, mdv);
+        return basename_from_metadata(mdv, frame_id);
       }
     }
     auto dummy_md = std::make_shared<kwiver::vital::metadata>();
     dummy_md->add<kwiver::vital::VITAL_META_VIDEO_URI>(std::string(video_name));
-    return frameName(frame_id, dummy_md);
+    return basename_from_metadata(dummy_md, frame_id);
   }
 };
 
@@ -630,9 +610,12 @@ add_command_options()
     cxxopts::value<std::string>() )
   ( "v,video", "Input video", cxxopts::value<std::string>() )
   ( "t,tracks", "Input tracks", cxxopts::value<std::string>() )
-  ( "a,carmera", "Output directory for cameras", cxxopts::value<std::string>() )
+  ( "k,camera", "Output directory for cameras", cxxopts::value<std::string>() )
   ( "l,landmarks", "Output landmarks file", cxxopts::value<std::string>() )
   ;
+
+  //If we want to remove tracks reading from the config, we should then add this
+  //m_cmd_options->parse_positional("tracks");
 }
 
 // ============================================================================
--- a/arrows/mvg/applets/init_cameras_landmarks.cxx
+++ b/arrows/mvg/applets/init_cameras_landmarks.cxx
@@ -38,8 +38,10 @@
 #include <vital/config/config_block_io.h>
 #include <vital/config/config_block.h>
 #include <vital/config/config_parser.h>
+#include <vital/io/camera_from_metadata.h>
 #include <vital/io/camera_io.h>
 #include <vital/io/landmark_map_io.h>
+#include <vital/io/metadata_io.h>
 #include <vital/io/track_set_io.h>
 #include <vital/plugin_loader/plugin_manager.h>
 #include <vital/util/get_paths.h>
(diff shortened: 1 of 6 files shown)


arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,513 @@
...
+// ----------------------------------------------------------------------------
+void
+init_cameras_landmarks::
+add_command_options()
+{
+  m_cmd_options->custom_help( wrap_text( "[options]\n" ) );
+
+  m_cmd_options->add_options()
+  ( "h,help",     "Display applet usage" )
+  ( "c,config",   "Configuration file for tool", cxxopts::value<std::string>() )
```
mleotta, line 499: I think it would be useful to add command line option here to provide input and output paths.  If specified, these would override what is in the config file.  In fact, the "config" option would rarely be needed at all if a default config is loaded automatically and files paths are specified on the command line.

We should probably also provide defaults for the output file names so that all you really need to provide is the input files on the command line

arrows/mvg/applets/init_cameras_landmarks.h
```diff
@@ -0,0 +1,62 @@
...
+namespace arrows {
+namespace mvg {
+
+class KWIVER_ALGO_MVG_APPLETS_EXPORT init_cameras_landmarks
+  : public kwiver::tools::kwiver_applet
+{
+public:
+  init_cameras_landmarks();
+
+  PLUGIN_INFO( "init_cameras_landmarks",
```
mleotta, line 47: Not something to worry about now, but we may ultimately want to change the name of this tool, and probably change the name of the algorithm it calls.  It's not longer about "initializing" as it was in the early days of MAP-Tk.  This is the full estimation process.  So "estimate_cameras_landmarks" is more correct.  However, we might just call it "sfm" for structure-from-motion because that is the more commonly used term for this.

No need to change anything now, I just wanted to put that out for discussion.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,513 @@
...
+    write_ply_file( landmark_map_ptr, out_landmarks_path );
+    return true;
+  }
+
+  void run_algorithm()
+  {
+    // If camera_map_ptr is Null the initialize algorithm will create all
+    // cameras.  If not Null it will only create cameras if they are in the map
+    // but Null.  So we need to add placeholders for missing cameras to the map
+    if (camera_map_ptr)
```
mleotta, line 370: Just an FYI, this block (and the landmarks block below) are about handling the case when you are re-running this algorithm after you've already created cameras and landmarks previously.  Right now this is not doing anything because we are not loading in previously existing cameras (or landmarks).  We may want to support that in the future but it is low priority.  This tool will primarily be used to create cameras and landmarks from scratch.  However, it does add another argument for changing the configuration name from "output_camera_directory" to "camera_directory".

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,513 @@
...
+    return true;
+  }
+
+  bool write_landmarks()
+  {
+    kwiver::vital::path_t out_landmarks_path =
+      config->get_value<kwiver::vital::path_t>("output_landmarks_filename");
+
+    // verify that we can open the output file for writing
+    // so that we don't find a problem only after spending
```
mleotta, line 350: As the comment indicates, this check would be better run before running all the computation.  I suggest moving this file write access check to the `check_config` function that we error out earlier if the file path is not writable.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,513 @@
...
+        sfm_constraint_ptr->set_metadata(metadata_map);
+      }
+    }
+  }
+
+  bool write_cameras()
+  {
+    std::string output_cameras_director =
+    config->get_value<std::string>("output_cameras_director");
+    std::string camera_prefix = config->get_value<std::string>("camera_prefix");
```
mleotta, line 328: See also other comment about camera prefix.  This was the old way of naming output files.  We now use the video metadata to set the name of these output files.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,513 @@
...
+      video_input::get_nested_algo_configuration("video_reader", config,
+                                                 video_reader);
+
+      video_reader->open( config->get_value<std::string>("video_source") );
+      if (video_reader->get_implementation_capabilities()
+        .has_capability(video_input::HAS_METADATA))
+      {
+        kwiver::vital::timestamp currentTimestamp;
+        kwiver::vital::metadata_map::map_metadata_t map_t;
+        while (video_reader->next_frame(currentTimestamp))
```
mleotta, line 307: Shortcut: the `video_input` class has a `metadata_map()` member that does this loop for you, and it may do it more efficiently, depending on which back end you have.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,513 @@
...
+
+  void load_tracks( )
+  {
+    kwiver::vital::path_t in_tracks_path =
+      config->get_value<kwiver::vital::path_t>("input_tracks_file");
+    feature_track_set_ptr =
+      kwiver::vital::read_feature_track_file(in_tracks_path);
+  }
+
+  void load_sfm_constraint( )
```
mleotta, line 284: As is obvious here, we only need the video for its metadata in this tool.  In the future it would be better to write out the metadata map to a file during tracking and just read that back in.  TeleSculptor doesn't even do that yet, but it should.

Note that this does not need to be address now, it more of a direction for future improvement.

mleotta (changes requested):
This is looking pretty good.  I've made some suggestions of some small things to fix.  There are also some comments about more forward looking ideas that don't need to be address on this PR.  The priority for now is to get this working and producing results that match what comes out of TeleSculptor.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,513 @@
...
+    {
+      LOG_ERROR(main_logger, "Configuration not valid.");
+      config = nullptr;
+      return FAIL;
+    }
+
+    return SUCCESS;
+  }
+
+  void register_algorithms()
```
linus-sherrill, line 251: This is not needed (and has no effect) because all plugins are loaded by the main kwiver program. Applets are loadable, so everything must be loaded before this applet can run.

russelldj (approved):

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+namespace {
+
+typedef kwiversys::SystemTools ST;
+typedef kwiversys::CommandLineArguments argT;
+
+kwiver::vital::logger_handle_t main_logger( kwiver::vital::get_logger( "init_cameras_landmarks" ) );
+
+//-------------------------------------------------------------------
+std::string
+frameName(kwiver::vital::frame_id_t frame,
```
mleotta, line 80: We should probably just move this helper function to the same header and source file that contains `basename_from_metadata` and just make this a version of `basename_from_metadata` that takes a `metadata_vector`.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+    {
+      opt_config = cmd_args["config"].as<std::string>();
+    }
+    if ( cmd_args.count("output-config") > 0 )
+    {
+      opt_out_config = cmd_args["output-config"].as<std::string>();
+    }
+
+    // Set up top level configuration w/ defaults where applicable.
+    config = default_config();
```
mleotta, line 256: See #1102 for and example of how to replace this function with a pointer to a default config file that we can put in `kwiver/config/applets/`.  Of course that branch will need to land first

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+    auto K = std::make_shared<simple_camera_intrinsics>(
+      GET_K_CONFIG(double, focal_length),
+      GET_K_CONFIG(kwiver::vital::vector_2d, principal_point),
+      GET_K_CONFIG(double, aspect_ratio),
+      GET_K_CONFIG(double, skew));
+
+    auto baseCamera = simple_camera_perspective();
+    baseCamera.set_intrinsics(K);
+    auto md = sfm_constraint_ptr->get_metadata()->metadata();
+    kwiver::vital::camera_map::map_camera_t camMap;
```
mleotta, line 397: might be good to rename variables avoiding camelCase now that we are firmly in KWIVER land and not in a Qt app where the styles are all intermixed.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+                                           init_intrinsics_with_metadata);
+
+        sfm_constraint_ptr->set_local_geo_cs(lgcs);
+      }
+    }
+  }
+
+  bool write_cameras()
+  {
+    std::string output_cameras_director =
```
mleotta, line 451: director -> directory

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+  m_cmd_options->add_options()
+  ( "h,help",     "Display applet usage" )
+  ( "c,config",   "Configuration file for tool", cxxopts::value<std::string>() )
+  ( "o,output-config",
+    "Output a configuration. This may be seeded with a "
+    "configuration file from -c/--config.",
+    cxxopts::value<std::string>() )
+  ( "v,video", "Input video", cxxopts::value<std::string>() )
+  ( "t,tracks", "Input tracks", cxxopts::value<std::string>() )
+  ( "a,carmera", "Output directory for cameras", cxxopts::value<std::string>() )
```
mleotta, line 633: The default value for this should probably be "results/krtd".

Maybe we should use "-k" as the single letter option rather than "-a"?  Nether is great, but k is linked to KRTD.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+  ( "l,landmarks", "Output landmarks file", cxxopts::value<std::string>() )
```
mleotta, line 634: The default value for this should probably be "results/landmarks.ply"

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+
+  m_cmd_options->add_options()
+  ( "h,help",     "Display applet usage" )
+  ( "c,config",   "Configuration file for tool", cxxopts::value<std::string>() )
+  ( "o,output-config",
+    "Output a configuration. This may be seeded with a "
+    "configuration file from -c/--config.",
+    cxxopts::value<std::string>() )
+  ( "v,video", "Input video", cxxopts::value<std::string>() )
+  ( "t,tracks", "Input tracks", cxxopts::value<std::string>() )
```
mleotta, line 632: I would suggest making tracks a positional argument since it is the required input.  The video input is optional and used to load metadata if desired, so that could stay as is.

mleotta (changes requested):
Looks really close now.  Just a few mostly stylistics suggestions.

mleotta:
Jenkins test this please

arrows/mvg/applets/init_cameras_landmarks.cxx
russelldj: carmera->camera

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+
+  void clear_ptrs()
+  {
+    camera_map_ptr = nullptr;
+    landmark_map_ptr = nullptr;
+    feature_track_set_ptr = nullptr;
+    sfm_constraint_ptr = nullptr;
+  }
+
+  void load_tracks( )
```
russelldj, line 331: Should this also have a check for the `config` being a `nullptr`?

dstoup:
Jenkins test this please

dstoup:
Since I decided to search for the Windows errors to make sure it wasn't another configuration issue, I figured I should post them here for your convenience.

**First error**
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\memory(2082): error C2027: use of undefined type 'kwiver::arrows::mvg::init_cameras_landmarks::priv' (compiling source file C:\workspace\KwiverWindowsPR\arrows\mvg\applets\register_applets.cxx) [C:\workspace\KwiverWindowsPR\build\arrows\mvg\applets\kwiver_algo_mvg_applets.vcxproj]
  C:\workspace\KwiverWindowsPR\arrows/mvg/applets/init_cameras_landmarks.h(54): note: see declaration of 'kwiver::arrows::mvg::init_cameras_landmarks::priv' (compiling source file C:\workspace\KwiverWindowsPR\arrows\mvg\applets\register_applets.cxx)

**Second error**
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\memory(2082): error C2338: can't delete an incomplete type (compiling source file C:\workspace\KwiverWindowsPR\arrows\mvg\applets\register_applets.cxx) [C:\workspace\KwiverWindowsPR\build\arrows\mvg\applets\kwiver_algo_mvg_applets.vcxproj]
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\memory(2084): warning C4150: deletion of pointer to incomplete type 'kwiver::arrows::mvg::init_cameras_landmarks::priv'; no destructor called (compiling source file C:\workspace\KwiverWindowsPR\arrows\mvg\applets\register_applets.cxx) [C:\workspace\KwiverWindowsPR\build\arrows\mvg\applets\kwiver_algo_mvg_applets.vcxproj]
  C:\workspace\KwiverWindowsPR\arrows/mvg/applets/init_cameras_landmarks.h(54): note: see declaration of 'kwiver::arrows::mvg::init_cameras_landmarks::priv' (compiling source file C:\workspace\KwiverWindowsPR\arrows\mvg\applets\register_applets.cxx)

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+  commandline_mode process_command_line(cxxopts::ParseResult& cmd_args)
+  {
+    static std::string opt_config;
+    static std::string opt_out_config;
+
+    if ( cmd_args["help"].as<bool>() )
+    {
+      return HELP;
+    }
+    if ( cmd_args.count("config") )
```
russelldj, line 246: The other ones seem to have the ` > 0` check for readability.

arrows/mvg/applets/init_cameras_landmarks.h
mleotta: For now we should change underscores to hyphens in this name as that is our convention for applet names.  So `init-cameras-landmarks`

arrows/mvg/applets/init_cameras_landmarks.cxx
judajake: I think this is partly dependent on whether or not we want to have read in capabilities via the config file.  If it is required from the command line, then we cannot have it in the config file.

arrows/mvg/applets/init_cameras_landmarks.cxx
```diff
@@ -0,0 +1,644 @@
...
+            if (md_K != nullptr)
+            {
+              baseCamera.set_intrinsics(md_K);
+              break;
+            }
+          }
+        }
+
+        local_geo_cs lgcs = sfm_constraint_ptr->get_local_geo_cs();
+        camMap =
```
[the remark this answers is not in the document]
judajake: camMap as camera_map_sptr instead of camera_map_t does not compile.  Granted, since it is not used right now, I do not need to keep the return from the function.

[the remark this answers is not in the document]
mleotta: Ah, we may need to package that camera map into a sptr.  Something like 
```c++
camera_map_ptr = std::make_shared<simple_camera_map>(camMap);
```
syntax may be slight off, I didn't test this.

mleotta:
Jenkins test this please

mleotta (approved):
As long as the CI builds pass, I approve of merging this.  There are additional things to improve, but these will come in a follow-up branch.

russelldj (approved):

russelldj:
Sorry, that was not intentional.

mleotta:
The Windows build that was failing previously is now passing.  That was the main barrier to merging.  The Linux CI build is currently broken on all branches for unrelated reasons, but both Travis CI and CentOS (which are both Linux) are passing.  So this is good enough for me.  I'm going to merge.

PR #1098, opened 2020-10-07
Fix for compilation with CUDA 10

chetnieter:
Added -fPIC flag to fix compilation with CUDA 10 on linux.


mleotta:
Looks good to me.

Issue #1048, opened 2020-08-10
Python required by vital_type_converters but shouldn't be


dstoup:
The vital_type_converters in sprokit/processes/bindings/c/ calls #include <python/kwiver/vital/util/python.h> which includes Python.h. But, the library isn't protected by if(KWIVER_ENABLE_PYTHON). That class is handling both python and C++ type conversion.

The end result is that you will get a build failure if you've disabled python. If it passes that error, I suspect there will be a link error when it tries to link vital_python_util.

In file included from /home/david/kitware/code/vision/kwiver/src/sprokit/processes/bindings/c/vital_type_converters.h:46,
                 from /home/david/kitware/code/vision/kwiver/src/sprokit/processes/bindings/c/vital_type_converters.cxx:36:
/home/david/kitware/code/vision/kwiver/src/python/kwiver/vital/util/python.h:58:12: fatal error: Python.h: No such file or directory
   58 |   #include <Python.h>
      |            ^~~~~~~~~~

dstoup:
Fixed in #1099

PR #1099, opened 2020-10-07
Remove unused vital_type_converters

dstoup:
This PR also fixes issue #1048

I opted to delete the code since it's not being used in kwiver. It's been replaced by the recent pybind11 effort. I did a build with almost everything enabled, CUDA being the one big exception though. It did build and tests all pass. I also greped the code and didn't see vital_type_converters being referenced anywhere.


tao558 (approved):

mleotta:
Probably out of scope for this branch, but do we need any of the C bindings anymore?  That is, can we remove C bindings from vital and remove the whole CMake option to build C bindings?  Maybe this is something to do after the remainder of the new pybind11 bindings land.  From what I remember the C bindings were only there to enable the old Python bindings.  They probably haven't been maintained very well and are incomplete.

johnwparent (approved):

as6520:
I don't think C bindings are being used anymore and we can probably remove them completely.

dstoup:
Thanks for the info on C bindings. I was wondering the same thing while making this PR. I will get this one landed so we squash the bug, then I'll try another PR to remove them all.

PR #1100, opened 2020-10-09
Sync track features

danlipsa:


danlipsa:
@mleotta Please review

arrows/mvg/applets/track_features.cxx
```diff
@@ -305,25 +397,38 @@ run()
...
+      auto const num_frames = static_cast<kwiver::vital::frame_id_t>(
+        video_reader->num_frames());
+      valid_frames.reserve(num_frames);
+      for (kwiver::vital::frame_id_t f = 1; f <= num_frames; ++f)
+      {
+        valid_frames.push_back(f);
+      }
+    }
+
+    std::vector<kwiver::vital::frame_id_t> camera_frames;
```
mleotta, line 427: I had forgotten that previously the `camera_frames` was just for restarting previous results that already had cameras.  More commonly, it was for cases where metadata was used to create initial guess for the cameras.  That is, this helps align which frames we track to which frames have metadata.

For now I suggested with initialize `camera_frames` with the frames of `md_map` if metadata is being used or empty if not.  That's not exactly what the GUI does.  The GUI converts metadata to cameras first.  It's possible that some frames with metadata will not have sufficient metadata to make a camera, so you will get fewer camera frames.  However, in most cases, `camera_frames` will be the same as the frames on which we have metadata.

mleotta (changes requested):
This looks like a good step in the right direction, but I have some further suggestions.

danlipsa:
@mleotta Please review. This the MR that addresses your comments and also uses only the selected_frames to trach features. It gets different tracks.txt than the gui though - maybe because of the cameras.

mleotta (approved):
Looks good to me, but we will have to resolved conflicts with #1101 before both can merge.

There are additional possibly improvements, like loading default config from an installed file and additional command line options to set input and output files, but these can come in a new branch.

mleotta:
Jenkins test this please

danlipsa:
Jenkins test this please

danlipsa:
@mleotta I fixed the conflicts. Is this ready to merge?

mleotta:
Jenkins test this please

mleotta:
@danlipsa I will merge when all the CI tests come back clean.

mleotta:
I'm not sure why travis CI is failing but it seems unrelated to this branch.  I will merge.

PR #1101, opened 2020-10-09
Removed plugin loader calls from applet.

linus-sherrill:
There is no need to load plugins from an applet. An applet is a plugin, so all plugins have to be already loaded before an applet can be instantiated. 

Also minor changes:
-Shorten name spaces
-converted some types to auto const

### DIFF
--- a/arrows/mvg/applets/track_features.cxx
+++ b/arrows/mvg/applets/track_features.cxx
@@ -222,11 +222,14 @@ int
 track_features::
 run()
 {
+  namespace kv = ::kwiver::vital;
+  namespace kva = ::kwiver::vital::algo;
+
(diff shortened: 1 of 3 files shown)


mleotta:
This looks okay to me, but it will likely conflict with other changes that @danlipsa is working on.  Also, we have other applets, like dump-klv, that would likely benefit from the same changes.

@danlipsa, if we merge this into master can you resolve the conflicts with your branch?

mleotta (approved):
This looks okay to me.  See other comments I made about other applets and conflicts with other branches.

danlipsa:
Sounds good. I'll apply this commit on top of mine.

PR #1102, opened 2020-10-13
Use config files for default applet configuration

mleotta:


mleotta:
This is not quite working yet, but illustrates what I had in mind for loading default applet configuration from installed config files

config/CMakeLists.txt
```diff
@@ -36,10 +38,11 @@ add_custom_target(kwiver-data-config ALL DEPENDS "${out_dir}")
 foreach(file ${config_files})
   get_filename_component(basename ${file} NAME_WE)
   get_filename_component(filename ${file} NAME)
+  get_filename_component(dir ${file} DIRECTORY)
   set(target ${basename})
   kwiver_configure_file(${target}
-    "${CMAKE_CURRENT_SOURCE_DIR}/${filename}"
-    "${out_dir}/${filename}"
+    "${CMAKE_CURRENT_SOURCE_DIR}/${dir}/${filename}"
+    "${out_dir}/${dir}/${filename}"
```
mleotta, line 45: This doesn't work correctly for installing files in sub-directories.  For one, the directory is not created without a custom target like the one used above.  Maybe I should put a new CMakeLists in the applets sub-directory and copy this installation code?

config/applets/dump_klv.conf
```diff
@@ -0,0 +1,4 @@
+# Default configuration for core dump-klv applet
+
+video_reader:type = ffmpeg
+video_reader:vidl_ffmpeg:time_source = misp
```
mleotta, line 4: `time_source` was copied from the hard coded default config, but this no longer does anything unless the reader type is set to `vidl_ffmpeg` instead of `ffmpeg`.

config/CMakeLists.txt
russelldj: I'm not sure if I understood you correctly, but I do like the idea of keeping configs near the relevant code so they are more likely to be updated appropriately. It looks like the KLV reader already has a sample config file, it's just not installed anywhere.

mleotta: I forgot about that sample.  We should probably remove that since you wouldn't use that anymore.  In the olden days you needed to pass a config file like this into the tool before it would do anything.  In the new way the config files are installed and found and loaded automatically.

To your point about keeping the config files near the source code, I can see the benefit of that.  On the other hand, these configurations are for the "global" kwiver system.  They are not specific to one tool or algorithm, rather they describe how to configure kwiver components together as a whole.  For example, dump-klv configuration is setting up ffmpeg configuration, which lives in a different arrow.  In many cases these config files include other config files, so it's helpful to have all the kwiver configuration files in one place to browse through them and see how they interact.

mleotta:
I've revised and this should be working now.  No longer WIP.

mleotta:
Jenkins test this please

mleotta:
All, it would help to land this branch today so that we can start rolling this configuration mechanism into the other applets.  Any objections?

russelldj (approved):

mleotta:
No objections received and I got one approval.  I'm going to merge so we can build on this.  If there are objections we can address them in a new PR.

PR #1106, opened 2020-10-16
Revise track features applet

mleotta:
This PR has further revision to the track features applets.  Primarily

1. refactoring into smaller functions using the PIMPL pattern following the approach used in #1096 
2. adding command line argument for specifying input and output files
3. loading default configuration for algorithm from installed configuration files
4. providing the missing default configuration file to load


danlipsa:
LGTM +1

russelldj (approved):

arrows/mvg/applets/track_features.cxx
```diff
@@ -296,101 +244,225 @@ select_frames(std::vector<kwiver::vital::frame_id_t> const& valid_frames,
...
+      "\n"
+      "Leave this blank if no image masking is desired.");
+
+    config->set_value("output_homography_file", homography_file,
+      "Optional path to a file to write source-to-reference "
+      "homographies for each frame. Leave blank to disable this "
+      "output. The output_homography_generator algorithm type "
+      "only needs to be set if this is set.");
+
+    config->set_value("invert_masks", false,
```
danlipsa, line 405: @mleotta This is already in track-features.conf. Why do you set it again in default_config()

mleotta: It's not strictly necessary for parameters to be in both places.  The reason to have them in both places is primarily for documentation.  I like having all parameters defined here in the code so that a developer can see what all the parameters are in one location.  It also means that if the installed config files can't be found for whatever reason the executable can still output what all the expected parameters are.  I like having the parameters also defined in the installed config files because this is what is easiest for end users to look at, copy, and modify after KWIVER is installed.  We could leave parameters like this out of the config file and the defaults would be the same, but then end users would be less likely to notice that this parameter even exists.  It's common practice to copy these installed config files into your working directory and modify them.  Of course the downside of having these specified in two places is keeping them up to date and in sync.  That's a legitimate problem that I don't have a good answer for.

danlipsa: Sounds good. Thanks Matt.

PR #1107, opened 2020-10-16
Enable the MVG arrow in the Docker build

mleotta:


dstoup:
Does this flag always default to OFF?
Also, is there any reason we should not have it running on every dashboard? If we should, then I will add that flag to our CI builds.
Finally, it looks like that flag is not included in the travis build. It is probably worth adding it either with this PR or another, once travis is fixed.

Otherwise, approved!

dstoup (approved):

mleotta:
No it defaults to ON, so I suppose this change isn't really needed.  I just know we were missing FFMPEG for a while in Docker and I wanted to make sure we were not missing MVG, but you are right that we probably don't need to explicitly turn it on and that we should be consistent in Travis CI and other testing configurations.  What do you think?  Should I just leave this out and go with the default or be explicit about it being on?

dstoup:
> ut you are right that we probably don't need to explicitly turn it on and that we should be consistent in Travis CI and other testing c

I think it's probably best to keep the flag explicitly ON, just in case. That's consistent with all of the other feature flags.

dstoup:
The Windows build is failing for a common and unrelated issue. Good to go unless you want to kick of the travis build again once the fix lands. I will try to investigate that common test failure on this dashboard today.

dstoup (approved):

mleotta:
This change will not impact any of our CI builds because none of them build the Docker image (that I know of). 

I will merge.

PR #1110, opened 2020-10-20
Add VTK arrow

mleotta:
This branch is to create a new VTK arrow in KWIVER.  For now I've just added one utility function that was ported from TeleSculptor.  We plan to add more to this arrow soon, but I want to land this simple branch first so we have a placeholder to add additional VTK functionality.


arrows/vtk/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  ${plugin_vtk_headers}
+  ${plugin_vtk_sources}
+  )
+
+target_link_libraries( kwiver_algo_vtk
+  PUBLIC               vital
+  PRIVATE              vtkFiltersSources
+  )
+
+#algorithms_create_plugin( kwiver_algo_mvg
```
linus-sherrill, line 31: I know this is just a comment, but it has the wrong library name.

arrows/vtk/depth_utils.cxx
```diff
@@ -0,0 +1,148 @@
...
+  vtkIdType pt_id = 0;
+
+  for (int y = nj - 1; y >= 0; y--)
+  {
+    for (int x = 0; x < ni; x++)
+    {
+      if (weight_img.size() > 0)
+      {
+        if (weight_img(x, y) > 127)
+          weight->SetValue(pt_id, 1.0);
```
linus-sherrill, line 105: According to style guide, all code blocks need braces.

arrows/vtk/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  )
+
+kwiver_add_library( kwiver_algo_vtk
+  ${plugin_vtk_headers}
+  ${plugin_vtk_sources}
+  )
+
+target_link_libraries( kwiver_algo_vtk
+  PUBLIC               vital
+  PRIVATE              vtkFiltersSources
```
dstoup, line 28: I think vtkFiltersSources needs to be PUBLIC since it's references in the header, depth_utils.h, thus exposed in the public API.

mleotta: I wasn't directly using the right libraries either.  It's fixed now.

dstoup: Thanks ... I wondered about that as well since the libraries didn't seem to match up with the headers but since I don't know much about VTK library organization, I took that one on faith :)

mleotta: Yeah, I don't know a good way of figuring out which VTK headers belong to which libraries other than by going into the VTK source code and searching for header file names to figure out where they came from, which is what I had to do here.

dstoup:
Jenkins test this please

dstoup (approved):
Pending dashboards

dstoup:
It's probably a good time to get a VTK upgrade in Fletch. Do you have a specific version I should add or just go with the latest release?

mleotta:
I just need VTK 8.2 for TeleSculptor.  The only newer version appears to be VTK 9.0.  I haven't tried that yet.  Presumably it will work, possibly with some minor tweaks.

CMake/kwiver-depends-VTK.cmake
```diff
@@ -0,0 +1,21 @@
...
+  )
+
+if( KWIVER_ENABLE_VTK )
+    find_package(VTK REQUIRED
+        COMPONENTS
+        vtkCommonCore
+        vtkCommonDataModel
+        )
+    if(VTK_VERSION VERSION_LESS 8.2)
+        message(FATAL_ERROR "${PROJECT_NAME} supports VTK >= v8.2 "
```
linus-sherrill, line 15: Should "supports" -> "requires"? Not sure what context this is used in.

arrows/vtk/depth_utils.cxx
```diff
@@ -0,0 +1,136 @@
...
+  const bool has_mask = mask_img.size() > 0;
+  const bool has_uncertainty = uncertainty_img.size() > 0;
+  for (int y = nj - 1; y >= 0; y--)
+  {
+    for (int x = 0; x < ni; x++)
+    {
+      depths->SetValue(pt_id, depth_img(x, y));
+
+      color->SetTuple3(pt_id,
+                       (int)color_img(x + i0, y + j0, 0),
```
linus-sherrill, line 108: These will generate some old style cast warnings.

linus-sherrill (approved):
Two minor comments. Nothing required.

dstoup:
> I just need VTK 8.2 for TeleSculptor. The only newer version appears to be VTK 9.0. I haven't tried that yet. Presumably it will work, possibly with some minor tweaks.

Thanks ... I noticed that after my post. I misread the version spec as > 8.2. I have had this upgrade on my plate for a while anyway, do going to finish. I will leave 8.2 as the default for now though, until you get some testing done.

dstoup:
I'm a bit confused about the linux build. The log and the dashboard say it passed but Jenkins says it failed. I am wondering if it's caused by the test

The following tests FAILED:
	755 - Download (Not Run)

We filter out 'Not Run' on Windows because there are so many. Not sure if this is a new issue or what it is actually downloading for that matter. I will dig in some.

mleotta:
Whatever that "Download" test is that is not running, it seems completely unrelated to this branch.  Can we merge?

PR #1112, opened 2020-10-20
Remove loading plugins from KLV applet.

linus-sherrill:
There is no point in loading the plugins from an applet since it is itself a plugin. All plugins have to be already loaded for the runner to find this applet.


mleotta (approved):
We already removed plugin loading in a recently merged branch, so this is mostly a no-op.  The change is now just updating the copyright year and removing a blank line.  These are still reasonable changes, so I approve, even if the name of the PR is no longer accurate.

mleotta:
Only comments and whitespace are changed in the diff, so I see no need to re-run Jenkins on this for the Linux build.

PR #1120, opened 2020-10-21
Fix example darknet sha512

dstoup:
The sha512 for the darknet model download in examples is out of date.


dstoup:
Well, I've locally tried every available sha512 and none of them make the 'Download' test succeed. @aaron-bray I think it's likely that something else is going wrong here but I'm not sure what it is. I can disable examples on this dashboard until we can figure it out but I don't love having it lurking as a disabled failure. Thoughts?

aaron-bray:
Hmm
Its possible you have to update the sha in the file and rerun all of cmake to get it to update the download script...

I will see if I can find some time to take a look

dstoup:
> Hmm
> Its possible you have to update the sha in the file and rerun all of cmake to get it to update the download script...
> 
> I will see if I can find some time to take a look

I thought about that too. I did rerun CMake but didn't remove everything and do it fresh every time. The dashboard does though so the sha512 you provided did fail here on a clean build.

dstoup:
Closing this PR. None of the available sha512 hashes solve the issue which leads me to believe that something larger is going wrong. I wrote it up as an issue, #1122

PR #1121, opened 2020-10-22
Refactor applet config validation

mleotta:
Applets reuse the same pattern repeatedly of checking configuration for input or output files and directories.  Depending on whether the input or output is required or optional the logic is slightly different, but generally we want to see if the parameter is set and if it points to a valid file or directory on disk.  For outputs, we may even want to create missing directories and test that we have permission to write to output files.

This PR refactors current applets to move this logic into a set of reusable utility functions that any applet can reuse.

### DIFF
--- a/vital/applets/config_validation.h
+++ b/vital/applets/config_validation.h
@@ -1,0 +1,88 @@
+/*ckwg +29
+ * Copyright 2020 by Kitware, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  * Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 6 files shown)


vital/applets/config_validation.h
```diff
@@ -0,0 +1,170 @@
...
+#define KWIVER_TOOLS_CONFIG_VALIDATION_H
+
+#include <vital/applets/kwiver_tools_applet_export.h>
+
+#include <vital/config/config_block.h>
+#include <vital/logger/logger.h>
+
+#include <string>
+
+namespace kwiver {
```
mleotta, line 41: @linus-sherrill I was following the established naming here, but this namespace and library name seems wrong given that this is in `vital\applets`.  Shouldn't the namespace be `kwiver::vital` or `kwiver::vital::applets`?

dstoup (approved):
I had the same question as Linus. Otherwise, looks good!

mleotta:
I'd like to merge this as is (consistent with existing namespaces) and then maybe we can address the namespace and library naming on a new branch.  Those naming changes would extend beyond the files modified in this branch.  @linus-sherrill do you have any objection to that plan?

vital/applets/config_validation.h
linus-sherrill: These are applet support, so `kwiver::vital::applet` would be appropriate.

linus-sherrill (approved):
The refactoring looks like a big improvement.

PR #1124, opened 2020-10-23
WIP: Disable VTK when a bad version is found

dstoup:
When we find VTK but the version isn't usable, disable VTK rather than failing.

I realize that this PR is a bit of a break in the current kwiver patter for third party library support but I wanted to pose it as a possible alternative since we have a growing number of packages that need newer versions that some of the supported fletch versions.

This PR isn't critical since I can easily change the dashboard scripts that build VIVIA, for example but again, wanted to see if there were objections to changing the pattern for supported packages in kwiver.

### DIFF
--- a/CMake/kwiver-depends-VTK.cmake
+++ b/CMake/kwiver-depends-VTK.cmake
@@ -6,16 +6,16 @@ option( KWIVER_ENABLE_VTK
   )
 
 if( KWIVER_ENABLE_VTK )
-    find_package(VTK REQUIRED
+    find_package(VTK 8.2
         COMPONENTS
         vtkCommonCore
         vtkCommonDataModel
         )
     if(VTK_VERSION VERSION_LESS 8.2)
-        message(FATAL_ERROR "${PROJECT_NAME} supports VTK >= v8.2 "
-            "(Found ${VTK_VERSION})")
-    endif()
-
-  include(${VTK_USE_FILE})
-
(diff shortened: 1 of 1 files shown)


dstoup:
I realize that this PR is a bit of a break in the current kwiver patter for third party library support but I wanted to pose it as a possible alternative since we have a growing number of packages that need newer versions that some of the supported fletch versions.

This PR isn't critical since I can easily change the dashboard scripts that build VIVIA, for example but again, wanted to see if there were objections to changing the pattern for supported packages in kwiver.

mleotta (commented):
I don't understand why we would want to do this.  Can you explain?  It seems to me that if someone turns on a package and they don't have the required dependencies that should be an error.  They can always manually turn it back off if they decide to proceed without that package.  The problem with the proposed approach is that you just get a warning (which will likely be missed) and then you spend time building everything without a package you asked for.  Then you might wonder why it is missing after an apparently successful build.

dstoup:
> I don't understand why we would want to do this. Can you explain? It seems to me that if someone turns on a package and they don't have the required dependencies that should be an error. They can always manually turn it back off if they decide to proceed without that package. The problem with the proposed approach is that you just get a warning (which will likely be missed) and then you spend time building everything without a package you asked for. Then you might wonder why it is missing after an apparently successful build.

It was mostly for support with superbuilds.  I will close and disable kwiver's VTK in scripts.

PR #1125, opened 2020-10-23
fixed logic errors in config validation

mleotta:
It turns out some of the logic was wrong in the recently merged branch on configuration validation

### DIFF
--- a/vital/applets/config_validation.cxx
+++ b/vital/applets/config_validation.cxx
@@ -48,10 +48,10 @@ validate_required_input_file(std::string const& name,
                              kv::logger_handle_t logger)
 {
   if (!config.has_value(name) ||
-      config.get_value<std::string>(name) != "")
+      config.get_value<std::string>(name) == "")
   {
     LOG_ERROR(logger, "Configuration value for " << name
-                     << " is missing but required");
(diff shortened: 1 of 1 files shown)


dstoup (approved):
LGTM

russelldj (approved):

mleotta:
Jenkins test this please

mleotta:
Any idea what happened to the CI on this branch?  It says the Windows build is still pending, but there is a message saying the Windows build succeeded.  Can we merge?

russelldj:
Unfortunately I have no idea about CI. However, I noticed this doesn't contain either `validate_optional_input_dir` or `validate_required_input_dir`. Was this intentional? If not, should it be included in this commit?

mleotta:
Good point on the missing input directory variants.  I should have add them previously, but we didn't have an applet that need input directories, so I guess that's why I didn't add them.  However, we do have a need for this case.  I'd rather add it in a new PR because this PR is about fixing logic bugs in the functions that we already merged.

linus-sherrill:
Sometimes CI gets stuck. In this case, there is an entry that the Windows build succeeded and that overrides what is in the summary list.

PR #1130, opened 2020-10-28
Add missing utilities for config validation for input directories

mleotta:
Previously we provided functions for input files, output files, and output directories.  This PR completes the set by adding input directories.

### DIFF
--- a/vital/applets/config_validation.cxx
+++ b/vital/applets/config_validation.cxx
@@ -150,15 +150,9 @@ validate_required_input_dir(std::string const& name,
                             kv::config_block const& config,
                             kv::logger_handle_t logger)
 {
-  if (!config.has_value(name) ||
-      config.get_value<std::string>(name) == "")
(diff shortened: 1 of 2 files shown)


russelldj:
LGTM. It looks like the errors are different between `validate_optional_output_dir` and `validate_optional_input_dir`. If that was intentional I think that makes perfect sense. If not, and we want analogous behavior between the tools, `validate_optional_input_dir(name, config, logger)` could call `validate_optional_`**`output`**`_dir(name, config, logger, `**`false`**`)` instead of being implemented independently.

russelldj (approved):

dstoup (approved):
LGTM

linus-sherrill (approved):

mleotta:
@russelldj makes a good point.  I'll do that.

russelldj (approved):

PR #1134, opened 2020-10-30
If trace set as default c++ log level use debug in python

mattdawkins:
trace doesn't exist in python


linus-sherrill (approved):
We should also look into removing the sprokit logging support and only use the vital implementation.

dstoup:
> We should also look into removing the sprokit logging support and only use the vital implementation.



> We should also look into removing the sprokit logging support and only use the vital implementation.

I agree 100%. My biggest concern with this PR was the duplication of the code in two places. Removing the sprokit logging will fix that.

dstoup (approved):
I think this looks good now. The premise makes sense and you've consolidated the functionality. Also looks like dashboards are clean.

as6520 (approved):

PR #1135, opened 2020-11-04
Color mesh

danlipsa:
@mleotta @russelldj The this the color mesh tool. Please review.

### DIFF
--- a/arrows/vtk/applets/CMakeLists.txt
+++ b/arrows/vtk/applets/CMakeLists.txt
@@ -1,0 +1,30 @@
+#
+# CMake file for the arrows.mvg.applets
+#
+
+set( sources
+  register_applets.cxx
+  color_mesh.cxx
+  )
+
+set( headers
+  color_mesh.h
+  )
+
+###
+# Add our plugin path to the default list
+include_directories( ${CMAKE_CURRENT_BINARY_DIR} )
+
+# Add applet plugin
+kwiver_add_plugin( kwiver_algo_vtk_applets
+  SUBDIR       ${kwiver_plugin_applets_subdir}
+  SOURCES      ${sources} ${headers}
+  PRIVATE      kwiver_algo_core
+               kwiver_algo_vtk
+               kwiver_tools_applet
+               kwiversys
+               vtkIOGeometry
+               vtkIOPLY
+               vtkIOXML
+
+)
--- a/arrows/vtk/applets/CMakeLists.txt
+++ b/arrows/vtk/applets/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-# CMake file for the arrows.mvg.applets
+# CMake file for the arrows.vtk.applets
 #
 
 set( sources
@@ -10,10 +10,6 @@ set( sources
 set( headers
   color_mesh.h
   )
-
-###
-# Add our plugin path to the default list
-include_directories( ${CMAKE_CURRENT_BINARY_DIR} )
 
 # Add applet plugin
 kwiver_add_plugin( kwiver_algo_vtk_applets
--- a/arrows/vtk/applets/color_mesh.cxx
+++ b/arrows/vtk/applets/color_mesh.cxx
@@ -51,6 +51,7 @@
 #include <vtkOBJReader.h>
 #include <vtkPLYReader.h>
 #include <vtkPLYWriter.h>
+#include <vtkPointData.h>
 #include <vtkXMLPolyDataReader.h>
 #include <vtkXMLPolyDataWriter.h>
 
@@ -121,7 +122,7 @@ public:
   std::string cameras_dir_;
   std::string mask_file_;
   std::string output_mesh_;
-  std::string ply_color_ = "mean";
+  std::string active_attribute_ = "mean";
   int frame_ = -1;
   int frame_sampling_ = 1;
   bool all_frames_ = false;
@@ -203,10 +204,10 @@ public:
       all_frames_ = cmd_args["all-frames"].as<bool>();
       config_->set_value("all_frames", all_frames_);
     }
-    if (cmd_args.count("ply-color"))
+    if (cmd_args.count("active-attribute"))
     {
-      ply_color_ = cmd_args["ply-color"].as<std::string>();
-      config_->set_value("ply_color", ply_color_);
+      active_attribute_ = cmd_args["active-attribute"].as<std::string>();
+      config_->set_value("active_attribute", active_attribute_);
     }
 
     bool valid_config = check_config(config_);
@@ -296,7 +297,7 @@ public:
       "remove_occluded", true,
       "Remove occluded points if parameter is true.");
     config->set_value(
-      "ply_color", ply_color_,
+      "active_attribute", active_attribute_,
       "What is saved in a PLY file, mean or median.");
     config->set_value(
       "remove_masked", true,
@@ -389,7 +390,7 @@ public:
   }
 
   bool save_mesh(vtkSmartPointer<vtkPolyData> mesh,
-                 char const * output_path, char const* ply_color)
+                 char const * output_path)
   {
     std::string ext = ST::GetFilenameExtension(output_path);
     std::string filename_noext = ST::GetFilenameWithoutExtension(output_path);
@@ -406,7 +407,7 @@ public:
     {
       vtkNew<vtkPLYWriter> writer;
       writer->SetFileName(output_path);
-      writer->SetArrayName(ply_color);
+      writer->SetArrayName(mesh->GetPointData()->GetScalars()->GetName());
       writer->AddInputDataObject(mesh);
       writer->Write();
       return true;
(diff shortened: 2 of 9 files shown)


mleotta:
This branch builds on #1128.  It will be easier to review this one once the other branch lands.

mleotta:
I've merged in #1128.  Please merge in the latest master or rebase on it so that this branch only shows new additions.  Also, if you could address the minor style issues commented on #1128 on this branch, that would be nice.

danlipsa:
@mleotta @russelldj I rebased on master and pushed a new version.

arrows/vtk/applets/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+
+# Add applet plugin
+kwiver_add_plugin( kwiver_algo_vtk_applets
+  SUBDIR       ${kwiver_plugin_applets_subdir}
+  SOURCES      ${sources} ${headers}
+  PRIVATE      kwiver_algo_core
+               kwiver_algo_vtk
+               kwiver_tools_applet
+               kwiversys
+               ${VTK_LIBRARIES}
```
russelldj, line 26: In some of the other applets I believe we link against specific vtk sub-libraries rather than all of them. This certainly seems cleaner, but I was wondering if there's a reason for the other approach?

arrows/vtk/applets/color_mesh.cxx
```diff
@@ -0,0 +1,527 @@
...
+
+class color_mesh::priv : public kwiver::arrows::vtk::mesh_coloration
+{
+public:
+  priv(color_mesh* parent) : p(parent)
+  {
+
+  }
+
+  color_mesh* p;
```
russelldj, line 112: Now that we have `kvt::find_configuration` this isn't used anywhere.

arrows/vtk/applets/color_mesh.cxx
```diff
@@ -0,0 +1,527 @@
...
+      vtkSmartPointer<vtkPolyData> mesh = reader->GetOutput();
+      return mesh;
+    }
+    else
+    {
+      return nullptr;
+    }
+  }
+
+  void save_mesh(vtkSmartPointer<vtkPolyData> mesh, const char* output_path)
```
russelldj, line 370: Did you mention that you are going to write `.obj` and `.ply` as well?

arrows/vtk/applets/color_mesh.cxx
```diff
@@ -0,0 +1,527 @@
...
+    "the average color or with a color for a particular camera is produced.") );
+
+  m_cmd_options->positional_help("\n  input-mesh  - input mesh file."
+                                 "\n  video-file  - input video file."
+                                 "\n  cameras-dir  - input camera directory."
+                                 "\n  output-mesh - output mesh file."
+  );
+
+  m_cmd_options->add_options()
+    ( "a,average-color",
```
russelldj, line 483: Do we have options for median as a config parameter?

russelldj (approved):
Looks good to me, just a few small questions.

arrows/vtk/applets/CMakeLists.txt
danlipsa: Probably the link command would take less time, and also its nice to see all the libraries (if there are not too many). I'll change it.

arrows/vtk/applets/color_mesh.cxx
danlipsa: I don't think so - this is what was originally supported by TeleSculptor. We can certainly add it though.

arrows/vtk/applets/color_mesh.cxx
danlipsa: When that is set, it saves both average and median. We could name it something different?

russelldj: What you're doing seems good to me.

arrows/vtk/applets/color_mesh.cxx
mleotta: @danlipsa please add support for writing to `.ply`. or `.vtp`.  You can decide which writer to use based on provided file extension.  `.ply` is a common output format for TeleSculptor because it can be opened by many more 3rd party programs than `.vtp`.  `.ply` supports color per vertex which we used here when we compute the median or mean color.  `'ply` will not support saving colors for multiple frames.  That was a feature requested by one customer and is not commonly used.  We can skip `.obj` for now because it does not support color per vertex.  However, if we ever get around to using texture maps instead of color per vertex, `.obj` supports texture maps, and I believe `.ply` does not.

arrows/vtk/applets/color_mesh.cxx
mleotta: The common case is computing the combined (e.g. mean or median) color.  So instead of an option for this I think it should save the combined/average color by default unless you used an option, like `--frame`, that indicates otherwise.  We could replace this option with "a,all-colors" instead and have it save colors of all frames if specified.  This would only work if the output file was a `.vtp`.

Note that when the output is not `.vtp` only one color mode (e.g. median or mean) can be saved.  So we should also have an option or configuration parameter to select which of the color arrays is saved.  In TeleSculptor we save whichever one is selected for visualization.

arrows/vtk/applets/color_mesh.cxx
danlipsa: Sounds good. I'll add PLY writing only for the average color. If the user wants PLY with color for multiple frames I'll print an error saying to use VTP.

danlipsa: Or maybe just save it as VTP and print a warning.

danlipsa: Done

arrows/vtk/applets/color_mesh.cxx
danlipsa: Sounds good.  I used "a,all-frames" and "p,ply_color" where ply_color can be mean or median.

arrows/vtk/applets/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+#
+# CMake file for the arrows.mvg.applets
```
russelldj, line 2: arrows.vtk.applets

arrows/vtk/applets/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+  color_mesh.cxx
+  )
+
+set( headers
+  color_mesh.h
+  )
+
+###
+# Add our plugin path to the default list
+include_directories( ${CMAKE_CURRENT_BINARY_DIR} )
```
russelldj, line 16: @mleotta has told me he doesn't think this is needed.

arrows/vtk/applets/register_applets.cxx
```diff
@@ -0,0 +1,67 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * \file
+ * \brief register mvg applets into a plugin
```
russelldj, line 33: vtk applets

arrows/vtk/applets/register_applets.cxx
```diff
@@ -0,0 +1,67 @@
...
+
+#include <arrows/vtk/applets/kwiver_algo_vtk_applets_export.h>
+#include <vital/plugin_loader/plugin_loader.h>
+#include <vital/applets/applet_registrar.h>
+
+#include <arrows/vtk/applets/color_mesh.h>
+
+namespace kwiver {
+namespace arrows {
+namespace mvg {
```
russelldj, line 44: namespace vtk

arrows/vtk/applets/register_applets.cxx
```diff
@@ -0,0 +1,67 @@
...
+    return;
+  }
+
+  // -- register applets --
+  reg.register_tool< color_mesh >();
+
+  reg.mark_module_as_loaded();
+}
+
+} // end namespace mvg
```
russelldj, line 65: here as well

russelldj:
The namespaces and exports need to be fixed in `color_mesh.h` and `color_mesh.cxx` as well.

arrows/vtk/applets/color_mesh.cxx
```diff
@@ -0,0 +1,579 @@
...
+    ( "h,help",     "Display applet usage" )
+    ( "m,mask-file",
+      "An input mask video or list of mask images to indicate "
+      "which pixels to ignore.",
+      cxxopts::value<std::string>())
+    ( "o,output-config",
+      "Output a configuration. This may be seeded with a "
+      "configuration file from -c/--config.",
+      cxxopts::value<std::string>() )
+    ( "p,ply-color",
```
mleotta, line 550: As discussed, I'd recommend calling this "active-color-mode" or something similar.  I'm not sure what the abbreviated flag should be since we've already used "a", "c", and "m".

danlipsa:
@russelldj @mleotta David, I think this is ready if you need to merge it in the common branch.

mleotta:
Jenkins test this please

mleotta (approved):

PR #1137, opened 2020-11-09
Dev/refactor pytest

johnwparent:
Nose, the current test runner for the Python bindings has been deprecated for quite some time. This branch implements a transition to using pytest in CMake code.
Additionally, one vital/type binding has been transitioned from nose as a proof of concept. The remainder of the bindings can be refactored before they're merged in. The transition can be supported by the tool: https://github.com/pytest-dev/nose2pytest. 
Pytest is perfectly capable of running tests written for nose, so this change should not drop, or cause any new failures in the python bindings tests


python/kwiver/vital/tests/test_activity_type.py
```diff
@@ -34,7 +34,6 @@
 
 """
 import unittest
-import nose.tools as nt
```
tao558, line 37: Was this meant to be removed?

python/kwiver/vital/tests/test_camera_intrinsics.py
```diff
@@ -51,8 +50,8 @@ def test_full_init(self):
         CameraIntrinsics(10, (5, 5), 1.2, 3.1, [4, 5, 6])
 
     def test_get_focal_length(self):
-        ntools.assert_equal(CameraIntrinsics().focal_length, 1.)
-        ntools.assert_equal(CameraIntrinsics(5.2).focal_length, 5.2)
+        assert(CameraIntrinsics().focal_length == 1.)
```
tao558, line 53: Nitpick: `Pytest` has `assertEqual`: https://docs.python.org/3/library/unittest.html

Not a huge deal to use `assert` here and everywhere else, but it might be cleaner. In general, anything that `Nose` had, `Pytest` has, but in camelCase instead of snake_case

python/kwiver/vital/tests/test_camera_intrinsics.py
```diff
@@ -109,10 +104,10 @@ def test_as_matrix(self):
     def test_equal(self):
         ci1 = CameraIntrinsics()
         ci2 = CameraIntrinsics()
-        ntools.assert_true(ci1 == ci2)
-        ntools.assert_false(ci1 != ci2)
```
tao558, line 113: Same as above. `assertFalse` is a function.

tao558 (changes requested):

python/kwiver/vital/tests/test_activity_type.py
johnwparent: Yes this was intentional, initially I removed it as I was going to refactor this test as the proof of concept, realized it implemented no nose functionality, and that import was therefor pointless so I left it out.

dstoup:
Looks like the only issue with the Linux build is failing tests.

python/kwiver/vital/tests/test_camera_intrinsics.py
johnwparent: You're right, pytest is able to drive and support most of those other testing frameworks, however pytest itself does not support those features. This test was intended to be an example of  the pure pytest functionality, while the rest of the testing suite served as an example of pytest running other testing frameworks like unittest and nose so we can get a sense which direction to take. Additionally, the tool mentioned above that can be used to automate the transition from nose to pytest performs the refactor in the same way done here, so if we wish to retain the unittest based tests, we'll need to change our approach.

johnwparent:
Right, pytest is found on the system rather than a virtualenv, so some dependencies required to run the tests are not met, causing the failures. The CMake reports that is is a potential scenario, but the failures point out a potential flaw of executing a system version of pytest the way I do, where pytest has the potential to be running from a different version of python than that used to link against the bindings, which is one of the most common pybind11 gotchas. Instead of running pytest from its command line script, it should be run with a call to `${PYTHON_EXECUTABLE} -m pytest` to ensure consistency with python versions. Do we want to continue to provide the option to run the python tests without creating a venv? As even with the aforementioned fix, this linux test will still fail due to unmet dependencies.

dstoup:
> Right, pytest is found on the system rather than a virtualenv, so some dependencies required to run the tests are not met, causing the failures. The CMake reports that is is a potential scenario, but the failures point out a potential flaw of executing a system version of pytest the way I do, where pytest has the potential to be running from a different version of python than that used to link against the bindings, which is one of the most common pybind11 gotchas. Instead of running pytest from its command line script, it should be run with a call to `${PYTHON_EXECUTABLE} -m pytest` to ensure consistency with python versions. Do we want to continue to provide the option to run the python tests without creating a venv? As even with the aforementioned fix, this linux test will still fail due to unmet dependencies.

If I'm understanding correctly, I think we should always call pytest with ${PYTHON_EXECUTABLE} -m pytest regardless of venv or not because that would ensure that the version of python that kwiver links against is always the one used to execute the tests. It also seems that we should always run the find_program(pytest ... that will pick up the venv version if we're running that way and the system version if we're not. We certainly should support running the tests without a virtual env since developers who may be working on C++ code that might impact the python bindings should be able to run tests to ensure their work didn't break anything in that world. Personally I'm not inclined to set up and activate a virtual env just to do that work but would happily pip install pytest on my system so I can forget about it. Let me know if I'm missing some subtlety here.

johnwparent:
> If I'm understanding correctly, I think we should always call pytest with ${PYTHON_EXECUTABLE} -m pytest regardless of venv or not because that would ensure that the version of python that kwiver links against is always the one used to execute the tests. It also seems that we should always run the find_program(pytest ... that will pick up the venv version if we're running that way and the system version if we're not. We certainly should support running the tests without a virtual env since developers who may be working on C++ code that might impact the python bindings should be able to run tests to ensure their work didn't break anything in that world. Personally I'm not inclined to set up and activate a virtual env just to do that work but would happily pip install pytest on my system so I can forget about it. Let me know if I'm missing some subtlety here.

Right, regardless of how we actually find pytest, it needs to be run by whatever python version kwiver is linking the bindings against. The find_program(pytest) command will always need to be run if a venv is unable to be created so that we can at least guarantee the system presence of pytest to be executed via ${PYTHON_EXECUTABLE} -m pytest. The only point that is maybe unclear is that kwiver's cmake does the work of constructing and setting up the venv responsible for managing the python tests and their dependencies, the only burden on users there would be to ensure their system has the ability to create venvs. Activation of the venv is also handled by kwiver. That being said, not forcing that extra overhead would definitely be the best direction to take, you make a very good point. Executing pytest from the system install may cause the python tests to fail unexpectedly (as with the Linux CI) but kwiver's cmake does report that such an outcome is possible/likely, so if the warning is sufficient, then this sounds good!

dstoup:
Jenkins test this please

dstoup:
I installed virtualenv in the docker image, rerunning to see if that's sufficient to have these tests pass.

dstoup:
Jenkins test this please

dstoup:
I installed  python3.4-venv in the docker container that runs the Linux tests. It still fails and I see the message "After installing the python3-venv package, recreate your virtual environment."

1. How do I do that?
2. Maybe the how should be part of the message

johnwparent:
> @johnwparent I installed python3.4-venv in the docker container that runs the Linux tests. It still fails and I see the message "After installing the python3-venv package, recreate your virtual environment."
> 
> 1. How do I do that?
> 2. Maybe the how should be part of the message

@dstoup That message is produced by python, Kwiver only reports that it fails, and forwards the Python error message/stdout. I am unsure as to why that's happening, recreating the virtualenv should be handled by kwiver's CMake, you  shouldn't need to do anything with respect to that. It's perplexing that it's still not working, I'm looking into it now.

dstoup:
> > @johnwparent I installed python3.4-venv in the docker container that runs the Linux tests. It still fails and I see the message "After installing the python3-venv package, recreate your virtual environment."
> > 
> > 1. How do I do that?
> > 2. Maybe the how should be part of the message
> 
> @dstoup That message is produced by python, Kwiver only reports that it fails, and forwards the Python error message/stdout. I am unsure as to why that's happening, recreating the virtualenv should be handled by kwiver's CMake, you shouldn't need to do anything with respect to that. It's perplexing that it's still not working, I'm looking into it now.

Thanks John. I went ahead and installed a system pytest. I will run again to see if that works. If not, I will clean up packages and standby. I just realized that the docker image is from ubuntu 14.04 (we are going to upgrade that once current pressures have relaxed). Wonder if that's causing some issue.

dstoup:
Jenkins test this please

dstoup:
> > > @johnwparent I installed python3.4-venv in the docker container that runs the Linux tests. It still fails and I see the message "After installing the python3-venv package, recreate your virtual environment."
> > > 
> > > 1. How do I do that?
> > > 2. Maybe the how should be part of the message
> > 
> > 
> > @dstoup That message is produced by python, Kwiver only reports that it fails, and forwards the Python error message/stdout. I am unsure as to why that's happening, recreating the virtualenv should be handled by kwiver's CMake, you shouldn't need to do anything with respect to that. It's perplexing that it's still not working, I'm looking into it now.
> 
> Thanks John. I went ahead and installed a system pytest. I will run again to see if that works. If not, I will clean up packages and standby. I just realized that the docker image is from ubuntu 14.04 (we are going to upgrade that once current pressures have relaxed). Wonder if that's causing some issue.

Well, I installed pytest but that doesn't seem to be working. I did see the command that was being run to create the venv, `/usr/bin/python3.4' '-m' 'venv' '/home/kitware/jenkins/workspace/KwiverLinuxPR/build/testing_venv`

Interestingly, running that command does work when run on the command line.

johnwparent:
> > > > @johnwparent I installed python3.4-venv in the docker container that runs the Linux tests. It still fails and I see the message "After installing the python3-venv package, recreate your virtual environment."
> > > > 
> > > > 1. How do I do that?
> > > > 2. Maybe the how should be part of the message
> > > 
> > > 
> > > @dstoup That message is produced by python, Kwiver only reports that it fails, and forwards the Python error message/stdout. I am unsure as to why that's happening, recreating the virtualenv should be handled by kwiver's CMake, you shouldn't need to do anything with respect to that. It's perplexing that it's still not working, I'm looking into it now.
> > 
> > 
> > Thanks John. I went ahead and installed a system pytest. I will run again to see if that works. If not, I will clean up packages and standby. I just realized that the docker image is from ubuntu 14.04 (we are going to upgrade that once current pressures have relaxed). Wonder if that's causing some issue.
> 
> Well, I installed pytest but that doesn't seem to be working. I did see the command that was being run to create the venv, `/usr/bin/python3.4' '-m' 'venv' '/home/kitware/jenkins/workspace/KwiverLinuxPR/build/testing_venv`
> 
> Interestingly, running that command does work when run on the command line.

That's bizarre, particularly because it works on the Travis CI... I think I have the CMake command in execute_process formatted correctly but I could be wrong. This [forum](https://stackoverflow.com/questions/39539110/pyvenv-not-working-because-ensurepip-is-not-available/39539571#39539571) seems to suggest that it's possibly a UTF-8 issue. Installing pytest should have worked however, but it seems as though Python3.4 is unable to find the module when executed by pytest, which would also explain why the camera, descriptor, and image tests were not run (they're the only tests that import pytest) and if pytest isn't accessible to Python3.4 as a module, the dependency checking I added for the python tests will prevent them from being added. However, I also don't see the CMake reporting that they're not being added, and it should be writing a warning. I also need to change how I find pytest, I'd move away from using find_program as I can't see an effective way to prevent it from finding a pytest associated with python2, and instead attempt to import pytest via an exec process to `python3.4 -m pytest` and verifying the return code, that at least, will prevent these tests from being run in situations like this with the Linux machine.

dstoup:
> > > > > @johnwparent I installed python3.4-venv in the docker container that runs the Linux tests. It still fails and I see the message "After installing the python3-venv package, recreate your virtual environment."
> > > > > 
> > > > > 1. How do I do that?
> > > > > 2. Maybe the how should be part of the message
> > > > 
> > > > 
> > > > @dstoup That message is produced by python, Kwiver only reports that it fails, and forwards the Python error message/stdout. I am unsure as to why that's happening, recreating the virtualenv should be handled by kwiver's CMake, you shouldn't need to do anything with respect to that. It's perplexing that it's still not working, I'm looking into it now.
> > > 
> > > 
> > > Thanks John. I went ahead and installed a system pytest. I will run again to see if that works. If not, I will clean up packages and standby. I just realized that the docker image is from ubuntu 14.04 (we are going to upgrade that once current pressures have relaxed). Wonder if that's causing some issue.
> > 
> > 
> > Well, I installed pytest but that doesn't seem to be working. I did see the command that was being run to create the venv, `/usr/bin/python3.4' '-m' 'venv' '/home/kitware/jenkins/workspace/KwiverLinuxPR/build/testing_venv`
> > Interestingly, running that command does work when run on the command line.
> 
> That's bizarre, particularly because it works on the Travis CI... I think I have the CMake command in execute_process formatted correctly but I could be wrong. This [forum](https://stackoverflow.com/questions/39539110/pyvenv-not-working-because-ensurepip-is-not-available/39539571#39539571) seems to suggest that it's possibly a UTF-8 issue. Installing pytest should have worked however, but it seems as though Python3.4 is unable to find the module when executed by pytest, which would also explain why the camera, descriptor, and image tests were not run (they're the only tests that import pytest) and if pytest isn't accessible to Python3.4 as a module, the dependency checking I added for the python tests will prevent them from being added. However, I also don't see the CMake reporting that they're not being added, and it should be writing a warning. I also need to change how I find pytest, I'd move away from using find_program as I can't see an effective, consistent, way to prevent it from finding a pytest associated with python2, and instead attempt to import pytest via an exec process to `python3.4 -m pytest` and verifying the return code, that at least, will prevent these tests from being run in situations like this with the Linux machine.

I was reading that exact post as well ... it also looks like the issue was solved on install of 16.04 which I believe is what travis is running. We are going to update those docker images soon but probably not until early next week. If that's the only issue remaining, I'm pretty confident that the update will fix it. I'm ok with merging if everything else is resolved, on the other hand maybe we don't want those tests running until after the update since it will gum up the dashboards.

johnwparent:
> > > > > > @johnwparent I installed python3.4-venv in the docker container that runs the Linux tests. It still fails and I see the message "After installing the python3-venv package, recreate your virtual environment."
> > > > > > 
> > > > > > 1. How do I do that?
> > > > > > 2. Maybe the how should be part of the message
> > > > > 
> > > > > 
> > > > > @dstoup That message is produced by python, Kwiver only reports that it fails, and forwards the Python error message/stdout. I am unsure as to why that's happening, recreating the virtualenv should be handled by kwiver's CMake, you shouldn't need to do anything with respect to that. It's perplexing that it's still not working, I'm looking into it now.
> > > > 
> > > > 
> > > > Thanks John. I went ahead and installed a system pytest. I will run again to see if that works. If not, I will clean up packages and standby. I just realized that the docker image is from ubuntu 14.04 (we are going to upgrade that once current pressures have relaxed). Wonder if that's causing some issue.
> > > 
> > > 
> > > Well, I installed pytest but that doesn't seem to be working. I did see the command that was being run to create the venv, `/usr/bin/python3.4' '-m' 'venv' '/home/kitware/jenkins/workspace/KwiverLinuxPR/build/testing_venv`
> > > Interestingly, running that command does work when run on the command line.
> > 
> > 
> > That's bizarre, particularly because it works on the Travis CI... I think I have the CMake command in execute_process formatted correctly but I could be wrong. This [forum](https://stackoverflow.com/questions/39539110/pyvenv-not-working-because-ensurepip-is-not-available/39539571#39539571) seems to suggest that it's possibly a UTF-8 issue. Installing pytest should have worked however, but it seems as though Python3.4 is unable to find the module when executed by pytest, which would also explain why the camera, descriptor, and image tests were not run (they're the only tests that import pytest) and if pytest isn't accessible to Python3.4 as a module, the dependency checking I added for the python tests will prevent them from being added. However, I also don't see the CMake reporting that they're not being added, and it should be writing a warning. I also need to change how I find pytest, I'd move away from using find_program as I can't see an effective, consistent, way to prevent it from finding a pytest associated with python2, and instead attempt to import pytest via an exec process to `python3.4 -m pytest` and verifying the return code, that at least, will prevent these tests from being run in situations like this with the Linux machine.
> 
> I was reading that exact post as well ... it also looks like the issue was solved on install of 16.04 which I believe is what travis is running. We are going to update those docker images soon but probably not until early next week. If that's the only issue remaining, I'm pretty confident that the update will fix it. I'm ok with merging if everything else is resolved, on the other hand maybe we don't want those tests running until after the update since it will gum up the dashboards.

I just changed how CMake tries to find pytest if the venv fails, so now the Linux dashboard is passing (because the tests aren't being run), so I don't think merging it would cause any failures on master. That being said, I'm still fairly confused/concerned by the fact that pytest is installed but python can't find the requisite module, so we may want to hold off on the merge until we can be confident this isn't an issue with my implementation.

python/kwiver/vital/tests/test_camera_intrinsics.py
```diff
@@ -109,10 +104,10 @@ def test_as_matrix(self):
...
+        self.assertEqual(ci1, ci2)
+        self.assertEqual(ci1, ci2)
```
tao558, line 108: This is the same call twice. Originally, the tests checked not only that `ci1 == ci21`, but that `ci1 != ci2` returned `False`

python/kwiver/vital/tests/test_camera_intrinsics.py
```diff
@@ -109,10 +104,10 @@ def test_as_matrix(self):
...
 
         ci1 = CameraIntrinsics(2, (10, 10), 3, 1)
         ci2 = CameraIntrinsics(2, (11, 10), 3, 0.1)
-        ntools.assert_false(ci1 == ci2)
-        ntools.assert_true(ci1 != ci2)
+        self.assertNotEqual(ci1, ci2)
+        self.assertNotEqual (ci1, ci2)
```
tao558, line 113: Same as above

python/kwiver/vital/tests/test_camera_intrinsics.py
johnwparent: Well, that's what I get for not reading carefully enough. Will update.

tao558 (approved):

linus-sherrill:
This branch is good to go as soon as the merge conflicts are resolved.

PR #1143, opened 2020-11-24
Fixed incorrect klv tag for target elevation

chetnieter:
Fixed incorrect klv tag for target elevation in klv conversion code.

### DIFF
--- a/doc/release-notes/release.txt
+++ b/doc/release-notes/release.txt
@@ -14,6 +14,8 @@ Config
 Vital
 
  * Added a missing include file in track.cxx required to build on MSVC 2019.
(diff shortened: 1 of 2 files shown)


dstoup (approved):

linus-sherrill (approved):

dstoup:
Two dashboards went down so they are a little backed up, but are running though now and shouldn't be too long for results.

mleotta (changes requested):
Change looks good, but can you make a comment about the fix in the release notes files (release.txt)?

mleotta (approved):

PR #1147, opened 2020-12-03
Update ffmpeg reader to use modern api

chetnieter:
This PR switches the ffmpeg_video_input class off of the deprecated ffmpeg api to the current modern api. This removes several compiler warnings.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_input.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_input.cxx
@@ -66,15 +66,13 @@ public:
     video_path(""),
     filter_desc("yadif=deint=1"),
     metadata(0),
-    frame_advanced(0),
+    frame_advanced(false),
     end_of_video(true),
     number_of_frames(0),
     collected_all_metadata(false),
     estimated_num_frames(false)
   {
-    f_packet.data = nullptr;
-
-    // Allocate data for AVFormatContext
+    // Allocate data
     f_format_context = avformat_alloc_context();
   }
 
@@ -87,7 +85,6 @@ public:
   AVStream* f_video_stream;
   AVFrame* f_frame;
   AVFrame* f_filtered_frame;
-  AVPacket f_packet;
   SwsContext* f_software_context;
   AVFilterGraph* f_filter_graph;
   AVFilterContext *f_filter_sink_context;
@@ -135,7 +132,7 @@ public:
   kwiver::vital::image_container_sptr current_image;
 
   // local state
-  int frame_advanced; // This is a boolean check value really
+  bool frame_advanced;
   bool end_of_video;
   size_t number_of_frames;
   bool collected_all_metadata;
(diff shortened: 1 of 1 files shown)


arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -67,13 +67,14 @@ class ffmpeg_video_input::priv
...
-    frame_advanced(0),
+    frame_advanced(false),
     end_of_video(true),
     number_of_frames(0),
     collected_all_metadata(false),
     estimated_num_frames(false)
   {
-    f_packet.data = nullptr;
+    // Allocate data
+    f_format_context = avformat_alloc_context();
```
mleotta, line 77: Could we just initialize this in the initializer list above?  Previously it was `f_format_context(nullptr),`, but now it could be `f_format_context(avformat_alloc_context()),`.

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -312,51 +310,16 @@ class ffmpeg_video_input::priv
...
-      avcodec_close(this->f_video_encoding);
-      avcodec_free_context(&this->f_video_encoding);
-      this->f_video_encoding = nullptr;
-    }
-    if (this->f_filter_graph)
-    {
-      avfilter_graph_free(&this->f_filter_graph);
-      this->f_filter_graph = nullptr;
-    }
+    av_frame_free(&this->f_frame);
```
mleotta, line 317: Noting that `f_video_stream` and `f_filter_graph` are no longer freed/closed.  Is this intentional?

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -1076,7 +1024,6 @@ ::frame_image( )
     }
 
     AVPixelFormat pix_fmt = static_cast<AVPixelFormat>(frame->format);
-    // If the pixel format is not recognized by then convert the data into RGB_24
```
mleotta, line 1079: This comment is still useful, but is probably not clear or this might not be the best place for it.  This is really about the `else` case down on line 1075.  If we are unable to directly copy the pixels into one of our standard image memory layouts then we ask FFMpeg to convert it from whatever it was into RGB24.  So it's not that the pixel format is not recognized, it's that the pixel format is not a standard format for vital images.

mleotta (commented):
No major concerns, but please review inline comments.

arrows/ffmpeg/ffmpeg_video_input.cxx
chetnieter: Done

arrows/ffmpeg/ffmpeg_video_input.cxx
chetnieter: No - I added back `avfilter_graph_free(&this->f_filter_graph);` and the setting of `f_video-stream` to null. The `avcodec_close` call is deprecated. The codec is now held by `f_video_encoding`. The various `*_free` methods in ffmpeg handle the case when the pointer is null and set the pointer to null after freeing the memory so the protective if statements are not needed.

arrows/ffmpeg/ffmpeg_video_input.cxx
chetnieter: I added the back to the location you suggested.

mleotta (approved):

PR #1148, opened 2020-12-04
Update descriptions for geo_point-type metadata

russelldj:
Fix typo `Target Geodetic Locationq` -> `Target Geodetic Location`

Update third item of metdata short tag from `meters` to `alt` or `elev` based on MISB 0601.

### DIFF
--- a/vital/types/metadata_tags.h
+++ b/vital/types/metadata_tags.h
@@ -127,7 +127,7 @@
         bool,                                                           \
         "" )                                                            \
   CALL( SENSOR_LOCATION,                                                \
-        "Sensor Geodetic Location (lon/lat/meters)",                    \
+        "Sensor Geodetic Location (lon/lat/alt)",                       \
         geo_point,                                                      \
         "Contains the 3D coordinate of the sensor. "                    \
         "The location is ordered lon, lat. "                            \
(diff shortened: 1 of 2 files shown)


vital/types/metadata_tags.h
```diff
@@ -127,7 +127,7 @@
         bool,                                                           \
         "" )                                                            \
   CALL( SENSOR_LOCATION,                                                \
-        "Sensor Geodetic Location (lon/lat/meters)",                    \
+        "Sensor Geodetic Location (lon/lat/alt)",                       \
```
mleotta, line 130: This might be just a GitHub formatting thing, but it looks like this is adding some whitespace to the end of this line.  Can you remove that?

vital/types/metadata_tags.h
```diff
@@ -183,11 +183,11 @@
         double,                                                         \
         "Target Width within sensor field of view." )                   \
   CALL( FRAME_CENTER,                                                   \
-        "Geodetic Frame Center, (lon/lat/meters)",                      \
+        "Geodetic Frame Center, (lon/lat/elev)",                        \
```
mleotta, line 186: remove extra trailing whitespace

vital/types/metadata_tags.h
```diff
@@ -222,11 +222,11 @@
         double,                                                         \
         "Temperature outside aircraft." )                               \
   CALL( TARGET_LOCATION,                                                \
-        "Target Geodetic Locationq (lon/lat/meters)",                   \
+        "Target Geodetic Location (lon/lat/elev)",                      \
```
mleotta, line 225: remove extra trailing whitespace

mleotta (commented):
Looks good other than the apparent addition of some trailing whitespace in a few places.

vital/types/metadata_tags.h
russelldj: I'm not sure what motivated the decision to add the trailing slashes, but each line is space-padded out to there. So the diff correctly reflects the additional required spaces to keep the trailing slashes aligned.

russelldj: I'm not super familiar with pre-processor wizardry, but I think these actually are needed to correctly do a multi-line macro.

linus-sherrill: Hmmm , that is not really at the end of the line.  It just looks like trailing WS because the the old characters were replaced by spaces. The line terminates with a`\` which can not have any following spaces.

mleotta: My mistake, I overlook the trailing `\`.  I forgot this was a macro.

mleotta (approved):
Looks good to me

PR #1153, opened 2020-12-07
Fuse depth applet

mleotta:
This branch was previously reviewed internally at Kitware and is now approved for public release

### DIFF
--- a/arrows/vtk/applets/fuse_depth.h
+++ b/arrows/vtk/applets/fuse_depth.h
@@ -17,13 +17,13 @@ class KWIVER_ALGO_VTK_APPLETS_EXPORT fuse_depth
 {
 public:
   fuse_depth();
-  virtual ~fuse_depth();
+  ~fuse_depth() override;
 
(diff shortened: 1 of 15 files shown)


russelldj (approved):
Looks good.

arrows/vtk/applets/fuse_depth.h
```diff
@@ -0,0 +1,37 @@
...
+{
+public:
+  fuse_depth();
+  virtual ~fuse_depth();
+
+  PLUGIN_INFO( "fuse-depth",
+               "Fuse depth maps from multiple cameras into a single surface");
+
+  virtual int run() override;
+  virtual void add_command_options() override;
```
linus-sherrill, line 26: A minor note - you don't need` virtual` with `override`

linus-sherrill (approved):

arrows/vtk/applets/fuse_depth.h
mleotta: good catch

PR #1155, opened 2020-12-10
WIP: Add TeleSculptor runscript

russelldj:
This adds a script to run `kwiver applets` in sequence to replicate TeleSculptor functionality. It depends on PRs #1152 , #1153 , and #1154 .

Currently, this script isn't being added to the build tree when you build. Perhaps I misunderstand what `install` is supposed to do.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -25,3 +25,6 @@ configure_file(pipelines/images_to_kwa.pipe.in ${EXAMPLE_DIR}/pipelines/images_t
 configure_file(pipelines/images_to_kwa_split.pipe.in ${EXAMPLE_DIR}/pipelines/images_to_kwa_split.pipe @ONLY )
 configure_file(pipelines/stabilize_images.pipe.in ${EXAMPLE_DIR}/pipelines/stabilize_images.pipe @ONLY )
 configure_file(pipelines/track_features.pipe.in ${EXAMPLE_DIR}/pipelines/track_features.pipe @ONLY )
+
+install( DIRECTORY applications
+	     DESTINATION ${EXAMPLE_DIR} )
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -26,6 +26,3 @@ configure_file(pipelines/images_to_kwa.pipe.in ${EXAMPLE_DIR}/pipelines/images_t
 configure_file(pipelines/images_to_kwa_split.pipe.in ${EXAMPLE_DIR}/pipelines/images_to_kwa_split.pipe @ONLY )
 configure_file(pipelines/stabilize_images.pipe.in ${EXAMPLE_DIR}/pipelines/stabilize_images.pipe @ONLY )
 configure_file(pipelines/track_features.pipe.in ${EXAMPLE_DIR}/pipelines/track_features.pipe @ONLY )
-
-install( DIRECTORY applications
-	     DESTINATION ${EXAMPLE_DIR} )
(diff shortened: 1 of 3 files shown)


examples/CMakeLists.txt
```diff
@@ -25,3 +25,6 @@ configure_file(pipelines/images_to_kwa.pipe.in ${EXAMPLE_DIR}/pipelines/images_t
 configure_file(pipelines/images_to_kwa_split.pipe.in ${EXAMPLE_DIR}/pipelines/images_to_kwa_split.pipe @ONLY )
 configure_file(pipelines/stabilize_images.pipe.in ${EXAMPLE_DIR}/pipelines/stabilize_images.pipe @ONLY )
 configure_file(pipelines/track_features.pipe.in ${EXAMPLE_DIR}/pipelines/track_features.pipe @ONLY )
+
+install( DIRECTORY applications
+	     DESTINATION ${EXAMPLE_DIR} )
```
mleotta, line 30: I think the issue you are seeing is that `install` moves files from the build tree to the install location.  These are source files, so they first need to be copied into the build tree and then installed.  I'm not sure what the most "correct" way of doing this is, but one option is what is done for config files, see https://github.com/Kitware/kwiver/blob/master/config/CMakeLists.txt

examples/applications/TeleSculptorCLI.py
```diff
@@ -0,0 +1,178 @@
+import argparse
```
mleotta, line 1: We should probably add a license header similar to what is in the C++ files

examples/applications/TeleSculptorCLI.py
```diff
@@ -0,0 +1,178 @@
...
+            Requires video-file, camera-dir, mesh-file\n
+        """,
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+# Required positiononals
+parser.add_argument(
+    "kwiver", help="Path to the kwiver binary")
+parser.add_argument(
+    "video",
+    help="The source video to run on. Can also be a text file containing a list of newline-seperated image paths")
```
mleotta, line 50: I'm not sure what code style standards we are using for Python in KWIVER (PEP8?) but we should break long lines

mleotta (changes requested):

examples/applications/TeleSculptorCLI.py
russelldj: Yes, it is PEP8. There is no explicit guidance on how to handle long string, and `autopep8` just ignores them. What I did is my preference, but I'm open to other options.

russelldj:
Again, I think I'm probably overlooking something obvious. The recent code adds the script to the specified build directory. However, it also adds `CMakeFiles`, `cmake_install.cmake`, `configure.TeleSculptorCLI.output` and `Makefile`.

russelldj:
Jenkins test this please

dstoup:
FYI, the Windows CI 'failure' is just the (Not Run) tests which we keep as a failure in the PR CI so we don't forget about it. Ghostflow issues need to be resolved though, if they haven't already.

russelldj:
Thanks. I was just using this to test my jenkins privileges, but I would like to get this in at some point.

mleotta:
Jenkins test this please

russelldj:
I can address ghostflow errors tonight.

russelldj:
I made a sloppy mistake on my rebase, and changed history more than I should. But it looks like it's the same except for removing the newlines.

PR #1157, opened 2020-12-11
Add VXL filters for BurnOut integration

russelldj:
This is a continuation of #725 , but I don't have permissions to update that branch. Please let me know if there's a better way to proceed.


russelldj:
I'm not sure what the above error means? It looks like it's trying to grab a commit from the Kitware repository rather than mine? Did I bungle the merge somehow?

russelldj:
I'm not fully convinced I did the merge correctly. However, it built and the same tests failed as on the `master` branch. Interestingly, there were 147 failing tests on both, so I'm not sure if that's a known problem. @mattdawkins is there a way to directly test this functionality?

arrows/vxl/color_commonality_filter.h
```diff
@@ -0,0 +1,79 @@
...
+  virtual ~color_commonality_filter();
+
+  /// Get this algorithm's \link vital::config_block configuration block \endlink
+  virtual vital::config_block_sptr get_configuration() const;
+  /// Set this algorithm's properties via a config block
+  virtual void set_configuration( vital::config_block_sptr config );
+  /// Check that the algorithm's currently configuration is valid
+  virtual bool check_configuration( vital::config_block_sptr config ) const;
+
+  /// Perform frame averaging
```
russelldj, line 67: This should be something about performing the commonality operation, correct?

arrows/vxl/color_commonality_filter.cxx
```diff
@@ -0,0 +1,608 @@
...
+}
+
+// Point an image view to a rectangular region in the image
+template< typename PixType, typename BoxType >
+void point_view_to_region( const vil_image_view< PixType >& src,
+                           const vgl_box_2d< BoxType >& region,
+                           vil_image_view< PixType >& dst )
+{
+  // Early exit case, no crop required
+  if( region.min_x() == 0 && region.min_y() == 0 &&
```
russelldj, line 91: Instead of equality, should this be if the region is equal to or larger than the input image?

arrows/vxl/color_commonality_filter.cxx
```diff
@@ -0,0 +1,608 @@
...
+void check_region_boundaries( vgl_box_2d< int >& bbox,
+  const vil_image_view< PixType >& img )
+{
+  vgl_box_2d<int> boundaries( 0, img.ni(), 0, img.nj() );
+  bbox = vgl_intersection( boundaries, bbox );
+}
+
+// Point an image view to a rectangular region in the image
+template< typename PixType, typename BoxType >
+void point_view_to_region( const vil_image_view< PixType >& src,
```
russelldj, line 86: This seems like an awkward function name.

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,483 @@
...
+    vil_convert_planes_to_grey( src, dst );
+  }
+  else
+  {
+    vil_math_mean_over_planes( src, dst );
+  }
+}
+
+
+template< typename Type >
```
russelldj, line 90: This should probably have a brief description

arrows/vxl/convert_image.h
```diff
@@ -0,0 +1,79 @@
...
+  virtual ~convert_image();
+
+  /// Get this algorithm's \link vital::config_block configuration block \endlink
+  virtual vital::config_block_sptr get_configuration() const;
+  /// Set this algorithm's properties via a config block
+  virtual void set_configuration( vital::config_block_sptr config );
+  /// Check that the algorithm's currently configuration is valid
+  virtual bool check_configuration( vital::config_block_sptr config ) const;
+
+  /// Perform frame averaging
```
russelldj, line 67: This should also talk about the specific operation done.

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,483 @@
...
+      scale_image< ipix_t, opix_t >( input, output, d->scale_factor ); \
+    }                                                                  \
+                                                                       \
+    return std::make_shared< vxl::image_container >( output );         \
+  }                                                                    \
+
+#define HANDLE_INPUT_CASE(T)                                           \
+  case T:                                                              \
+    {                                                                  \
+      typedef vil_pixel_format_type_of<T >::component_type ipix_t;     \
```
russelldj, line 418: formatting

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,722 @@
...
+  {
+    return image_data;
+  }
+
+  // Get input image
+  vil_image_view_base_sptr view =
+    vxl::image_container::vital_to_vxl( image_data->get_image() );
+
+  // Perform different actions based on input type
+#define HANDLE_CASE(T)                                                 \
```
russelldj, line 656: Is this the preferred of dealing with different VXL parameterizations in the underlying image data? I'm implementing similar filters and have not seen a better way, but wanted to make sure I'm not missing something.

mleotta: In general, yes, VXL uses this `DO_CASE` or `HANDLE_CASE` macro pattern to deal with dispatching to different code based on dynamic type.  However, in this case it's not done very cleanly.  Usually the macro is just a couple lines of code to find the type and call a templated function.   That's sort of what we have here except everything after the `vil_convert_cast` does not depend on the type, so it doesn't need to be written out in the macro.  It would be better to call some helper function to do that stuff because 1. it reduces code bloat and 2. it would probably be easier to debug.

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,722 @@
...
+    HANDLE_CASE(VIL_PIXEL_FORMAT_INT_16);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_UINT_32);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_INT_32);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_UINT_64);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_INT_64);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_FLOAT);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_DOUBLE);
+#undef HANDLE_CASE
+
+  default:
```
mleotta, line 696: This is bad practice to just assume the default case is a `vxl_byte` type.  It's possible that we could be missing a type above or someone could add a new type in the future that is not handle above and the default will then incorrectly assume it is a byte.  There is no reason not to use `case VIL_PIXEL_FORMAT_BYTE:` instead.  The `default` case should probably log an error about an unknown type and return the null image.

Also, I'm looking at this out of context, but it seems very odd that the byte image case returns a byte image output if `!d->output_variance` but otherwise (and for all other types) returns a double image.

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,722 @@
+/*ckwg +5
```
mleotta: needs modern license banner

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,722 @@
...
+#include <exception>
+#include <memory>
+
+namespace kwiver {
+namespace arrows {
+namespace vxl {
+
+namespace {
+
+/// Base class for all online frame averagers instances
```
mleotta, line 24: This is one of those cases that doesn't need anything special from VXL.  It's just simple math operations on images which KWIVER can easily do.  No need to revise on this branch.  But at some point I'd like to see a new "image processing" arrow that has operations like this without requiring VXL.

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,722 @@
...
+  unsigned ni = input.ni(), nj = input.nj(), np = input.nplanes();
+  output.set_size( ni, nj, np );
+
+  // s* = source properties, d* = dest
+  std::ptrdiff_t sistep=input.istep(), sjstep=input.jstep(), spstep=input.planestep();
+  std::ptrdiff_t distep=output.istep(), djstep=output.jstep(), dpstep=output.planestep();
+
+  // Iterate through image and perform cast and any required ops. We could
+  // shorten this code using vil_transform and functors (binding arguments),
+  // but just to ensure they get inlined...
```
mleotta, line 225: This is bad practice.  We have transform functions to do exactly these operations and do them efficiently.  It's never a good idea to have multiple copies of the same code repeated multiple time in the hope that "manually inlining" will make the code more efficient.  We should prefer calling functions for maintainability and readability until the code is proven to be a bottleneck and needs optimization.  Even then, we can use template functions to implement that as efficiently as this without keeping multiple copies of the code.

arrows/vxl/average_frames.cxx
russelldj: Thanks, that's really helpful. I think this has been updated appropriately.

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,731 @@
...
+  config->set_value( "type", d->type,
+    "Operating mode of this filter, possible values: cumulative, window, "
+    "or exponential." );
+  config->set_value( "window_size", d->window_size,
+    "The window size if computing a windowed moving average." );
+  config->set_value( "exp_weight", d->exp_weight,
+    "Exponential averaging coefficient if computing an exp average." );
+  config->set_value( "round", d->round,
+    "Should we spend a little extra time rounding when possible?" );
+  config->set_value( "output_variance", d->output_variance,
```
russelldj, line 629: This seems to be in conflict with the actual behavior, which seems to always output variance as a double. However, as @mleotta mentioned, the `average` image can be a byte, which seems odd.

arrows/vxl/color_commonality_filter.cxx
```diff
@@ -0,0 +1,610 @@
...
+    HANDLE_CASE(VIL_PIXEL_FORMAT_BYTE);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_SBYTE);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_UINT_16);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_UINT_32);
+    HANDLE_CASE(VIL_PIXEL_FORMAT_UINT_64);
+#undef HANDLE_CASE
+
+  default:
+    LOG_ERROR( logger(), "Invalid type received" );
+    return image_data;
```
russelldj, line 601: This seems like a dangerous return value. At very least, I think this should be consistent with the final unreachable return and the other analogous `filter` functions. And actually, I think if any functions should return the original image, it should be the `convert_image` or `average`, because then there's at lease some chance of it being correct or meaningful. E.g., the current image is closely related to the average of historical frames.

russelldj:
I was trying to address the `ghostflow` errors by rebasing the history and editing the commits requested by the bot. It looks like I'd have to re-resolve the merge conflicts during the rebase. Is this normal behavior? If so, should I go through with the rebase or is there a better solution? Unfortunately, I didn't enable `git rerere` as Matt suggested, so that might take a bit.

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,483 @@
...
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_BYTE);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_SBYTE);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_UINT_16);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_INT_16);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_UINT_32);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_INT_32);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_UINT_64);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_INT_64);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_FLOAT);
+    HANDLE_INPUT_CASE(VIL_PIXEL_FORMAT_DOUBLE);
```
mleotta, line 467: I'd recommend that we trim down this set of types to the ones we actually plan to support.  This 11x11 set of type expansions is going to create a lot of code bloat.  There will be 121 different copies of many of the functions in this files, most of which will never be used.  We could probably limit this to common types.  We could also refactor the code a bit to limit code bloat by templating only when really necessary.

These changes are not necessary to merge.  I'm just pointing out an area of possible improvement.

arrows/vxl/image_io.cxx
```diff
@@ -138,6 +142,97 @@ convert_image_helper(const vil_image_view<bool>& src,
...
+
+std::string
+plane_filename(std::string filename, unsigned p)
+{
+  std::string parent_directory =
+    kwiversys::SystemTools::GetParentDirectory(filename);
+  std::string file_name_with_ext =
+    kwiversys::SystemTools::GetFilenameName(filename);
+
+  std::size_t last_index = file_name_with_ext.find_last_of( "." );
```
mleotta, line 154: use `GetFilenameLastExtension()` and `GetFilenameWithoutLastExtension()` instead

arrows/vxl/image_io.cxx
```diff
@@ -214,13 +311,21 @@ ::get_configuration() const
...
+
   if( d_->manual_stretch )
   {
     config->set_value("intensity_range", d_->intensity_range.transpose(),
                       "The range of intensity values (min, max) to stretch into "
                       "the byte range.  This is most useful when e.g. 12-bit "
                       "data is encoded in 16-bit pixels");
   }
+
+  config->set_value("split_channels", d_->split_channels,
```
mleotta, line 323: A unit test for this new case would be good.

arrows/vxl/image_io.cxx
```diff
@@ -18,10 +18,14 @@
 #include <arrows/vxl/image_container.h>
 
 #include <vil/vil_convert.h>
+#include <vil/vil_plane.h>
```
mleotta, line 21: I would recommend squashing the changes to this one file into a single commit, add test case for reading and writing split images, and make that a new branch for just that change.  That should be easy to get merged by itself.

vital/util/buffer.h
```diff
@@ -0,0 +1,56 @@
...
+
+#ifndef KWIVER_VITAL_UTIL_BUFFER_H_
+#define KWIVER_VITAL_UTIL_BUFFER_H_
+
+namespace kwiver {
+namespace vital {
+
+
+/// Base interface for simple buffer classe
+template< class Data >
```
mleotta, line 18: I'm curious why we need an abstract base class for buffers of a fixed type?  What are the planed derived classes other than a ring buffer?  Why do we need polymorphism here?  It would seem better to just make an STL style ring buffer class.

vital/util/ring_buffer.h
```diff
@@ -0,0 +1,138 @@
...
+#include <exception>
+
+#include "buffer.h"
+
+namespace kwiver {
+namespace vital {
+
+
+template< class Data >
+class ring_buffer : public buffer< Data >
```
mleotta, line 23: I see very little reason for having this class.  I suggest we just use an `std::deque` instead of a ring buffer and keep the max_capacity in a separate side car variable.

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,740 @@
...
+  /// Reset the internal average.
+  virtual void reset();
+
+  /// Get number of frames used in the current window
+  virtual unsigned frame_count() const;
+
+protected:
+
+  /// Buffer containing pointers to last window_length frames
+  kwiver::vital::ring_buffer< vil_image_view< PixType > > window_buffer_;
```
mleotta, line 146: we don't need a whole new `ring_buffer<>` class for this.  Just use `std::deque<>` and store capacity in a side car variable.  Use `.push_back()` to add values to the buffer, then check if `.size() > capacity` and if so call `.pop_front()`.  This is simple and probably just as efficient.  It might be different if we were actually reusing the memory in the ring buffer, but that's not the way it's implemented here, so there is no benefit to the ring buffer.

mleotta (changes requested):
I've review this any made several suggestions and posted several questions.  I'm mostly looking for @mattdawkins to respond to these to provide clarification or input for @russelldj.

Overall, I would be very helpful to cherry pick out the separate filters in this branch to separate branches and review and merge independently.  Otherwise, this will take much longer to approve.

I know the argument is that rewriting history (i.e. rebasing) code already merged into VIAME makes it harder to merge `master` into VIAME later, but I still don't understand why that is the case.

arrows/vxl/average_frames.cxx
russelldj: I believe that variance is a configurable feature in the [BurnOut feature extractor super process](https://github.com/Kitware/burn-out/blob/1d2afcd36d37127b5a3b6fa3f60bc195408ab3ea/library/object_detectors/pixel_feature_extractor_super_process.txx). In the default config I've seen, it is on. So my understanding is it is being used.

arrows/vxl/average_frames.cxx
russelldj: Addressed in #1184

arrows/vxl/average_frames.cxx
```diff
@@ -0,0 +1,740 @@
...
+    {
+      const inT* srow = splane;
+      outT* drow = dplane;
+      for (unsigned j=0;j<nj;++j,srow += sjstep,drow += djstep)
+      {
+        const inT* spixel = srow;
+        outT* dpixel = drow;
+        for (unsigned i=0;i<ni;++i,spixel+=sistep,dpixel+=distep)
+        {
+          *dpixel = static_cast<outT>(*spixel * scale_factor + 0.5);
```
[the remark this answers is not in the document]
russelldj: There is a prior check in this function which ensures the output is integral in this case. This looks like it's circumventing the fact that `static_cast` truncates rather than rounds. I've replaced this whole thing with `vil_transform_*` things in #1184 .

russelldj:
All the changes in this branch are in the process of landing on #1177 - #1191. Closing.

PR #1159, opened 2020-12-15
Update copyright banner in vital/bindings

mwoehlke-kitware:
Update copyright banners in `vital/bindings/`.

See also #1142 and other PRs referencing that.


mleotta (approved):
I approve, but we probably could have also just removed all of these files.  I think the C bindings are no longer being actively maintained.

tao558 (approved):

PR #1167, opened 2020-12-21
Peg FFmpeg version to 3.3.3. Fletch version 2.6.2 doesn't build anymore.

dstoup:
Recent FFmpeg changes are preventing the arrow from building against the older Fletch version, 2.6.2. I've pushed this PR to peg the find_package version to the newer one from Fletch, 3.3.3. Considering these changes, I'm not sure if there is a need to keep 2.6.2 in Fletch anymore. If not, I will remove that and push a PR there.

Alternatively, if the older version is still desired, I can regenerate the build errors and post them here so they can get fixed.

### DIFF
--- a/CMake/kwiver-depends-ffmpeg.cmake
+++ b/CMake/kwiver-depends-ffmpeg.cmake
@@ -6,5 +6,5 @@ option( KWIVER_ENABLE_FFMPEG
   )
 
 if( KWIVER_ENABLE_FFMPEG )
-  find_package( FFMPEG REQUIRED )
+  find_package( FFMPEG 3.0  REQUIRED )
 endif( KWIVER_ENABLE_FFMPEG )


mleotta:
Maybe the better approach is to find FFMpeg and then check that the version found is at least 3.x  I'm okay with requiring at least version 3.x and removing support for 2.6.2, but I don't think we should require exactly version 3.3.3.  I expect that other version of 3.x will also work, and possibly even 4.x.

mleotta (changes requested):
I would prefer that we check the found version and make 3.x the minimum instead of requiring exactly 3.3.3

dstoup:
> I would prefer that we check the found version and make 3.x the minimum instead of requiring exactly 3.3.3

That makes sense. In this case, 3.3.3 would be the minimum version since that's CMake's behaviour unless 'EXACT' is added. I used 3.3.3 since that is the only version I know works. If you know that earlier versions work, I can lower that to 3.0. Change coming.

dstoup:
Jenkins test this please.

dstoup:
I accidentally killed the agent on Windows which caused the latest build failure.

chetnieter (approved):

dstoup:
Jenkins test this please

dstoup:
@mleotta I force pushed in the interest of a clean history so I can't mark your comments as resolved, but the request has been resolved.

mleotta (approved):

PR #1174, opened 2021-01-07
Avoid global logger handles

mleotta:
When a logger handle is declared globally it causes the logger pluggins
to load and configure before entering the main function of a program.
This means main functions are not given any chance to configure logging
before the loggers are established.

### DIFF
--- a/arrows/vtk/mesh_coloration.cxx
+++ b/arrows/vtk/mesh_coloration.cxx
@@ -56,8 +56,6 @@ static void ComputeMedian(std::vector<T> vector, double& median)
     }
 }
 
-  static kwiver::vital::logger_handle_t main_logger( kwiver::vital::get_logger( "arrows.vtk.mesh_coloration" ) );
-
 }
 
 namespace kwiver {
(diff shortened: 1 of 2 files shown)


mleotta:
This fix addresses my immediate issues, but the use of global logger handles created a subtle bug in TeleSculptor that was difficult to track down.  It might be worth banning this practice of global logger handles altogether, and enforcing it if possible.  It would be great if we could make this a compile failure, but I don't know if that is possible or how to do that.  Instead we might need to require an explicit function call to initialize the global logger manager singleton, which would need to be called from the main function.  Any calls trying to access the logger manager instance before manual initialization would throw an exception.

I'm guilty of this practice of using global logger handles in my older command line tools (which is where I think dan copied this code from), but now I'm against it.  Globals are convenient and mostly "just work" without any need for explicit logger initialization.  However, what we give up is more control over logger initialization.  Executables should be allowed to do some logger configuration (e.g. based on command line arguments) before initializing the logger.

As an example, in TeleSculptor we want to enforce that log4cplus is used as the logger when no environment variable is set otherwise.  So I need to check that the environment variable is empty and, if so, set it to log4cplus *before* the logger is initialized.  Thoughts on this?  It would be a departure from the current design where logger is automatically initialized (and plugin loaded) wherever it is first used to create a handle.

linus-sherrill (approved):

PR #821, opened 2019-07-05
Allow "downsampler" to work on input with missing frames

neal-siekierski:
This PR contains changes to allow `downsampler_process` to work on input that's missing frames by using the provided timestamp to determine the elapsed time. For cases when a timestamp is missing, the code assumes the provided frame rate. It doesn't use it otherwise.

I wasn't sure how to have this interact with the burst mode when there's such a large gap that a down-sampled output would be missed. I chose to have it count the missed output as well. (Normally, at say 5Hz output and a 3/2 cycle, you'd have outputs at 0s, 0.2s, 0.4s, 1s, 1.2s, 1.4s, etc. I've implemented it such that if there is no input around e.g. 1.2s, that output is missed but continues as normal, thus 1s, ~1.2s~, 1.4s, 2s, 2.2s, 2.4s, etc.)

Let me know if this should be against another branch. Once it's finalized, I can update #690 and add this to it.

Here is an extract from a log I made of output time stamps from a test video from @mattdawkins with a nominal ~30Hz frame rate but with many single missing frames and down-sampled to 5Hz:
```
ts(f: 13570, t: 770633333 (Wed Dec 31 19:12:50 1969), d: 0)
ts(f: 13573, t: 770800000 (Wed Dec 31 19:12:50 1969), d: 0)
ts(f: 13576, t: 771000000 (Wed Dec 31 19:12:51 1969), d: 0)
ts(f: 13581, t: 771266666 (Wed Dec 31 19:12:51 1969), d: 0)
ts(f: 13583, t: 771400000 (Wed Dec 31 19:12:51 1969), d: 0)
ts(f: 13587, t: 771633333 (Wed Dec 31 19:12:51 1969), d: 0)
ts(f: 13590, t: 771800000 (Wed Dec 31 19:12:51 1969), d: 0)
ts(f: 13594, t: 772000000 (Wed Dec 31 19:12:52 1969), d: 0)
ts(f: 13599, t: 772266666 (Wed Dec 31 19:12:52 1969), d: 0)
ts(f: 13601, t: 772400000 (Wed Dec 31 19:12:52 1969), d: 0)
ts(f: 13606, t: 772633333 (Wed Dec 31 19:12:52 1969), d: 0)
ts(f: 13609, t: 772800000 (Wed Dec 31 19:12:52 1969), d: 0)
ts(f: 13613, t: 773000000 (Wed Dec 31 19:12:53 1969), d: 0)
[...]

PR #1176, opened 2021-01-12
Cherry pick downsampler fix for missing frames

russelldj:
This PR cherry picks @neal-siekierski 's changes induced in #821 onto `viame/master`. It fixes an issue with a video containing dropped frames but has **not** been tested to assure there are not regressions with currently-working data. I am in the process of doing that.

### DIFF
--- a/sprokit/processes/core/downsample_process.cxx
+++ b/sprokit/processes/core/downsample_process.cxx
@@ -59,7 +59,6 @@ public:
   unsigned burst_frame_break_;
   bool renumber_frames_;
 
-  double ds_factor_;
   double ds_counter_;
   unsigned burst_counter_;
   bool burst_skip_mode_;
@@ -262,12 +261,12 @@ void downsample_process
 bool downsample_process::priv
 ::skip_frame( vital::timestamp const& ts, double frame_rate )
 {
-  ds_factor_ = frame_rate / target_frame_rate_;
+  double ds_factor = frame_rate / target_frame_rate_;
 
   if( is_first_ )
   {
     // Triggers always sending the first frame
-    ds_counter_ = ds_factor_;
+    ds_counter_ = ds_factor;
     is_first_ = false;
   }
   else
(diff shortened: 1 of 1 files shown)


neal-siekierski:
It looks like the changes applied cleanly, which is good, but I'm somewhat concerned about the changed context this now finds itself in. Originally `skip_frame` was called every frame and was passed a notionally valid timestamp and frame rate:

https://github.com/Kitware/kwiver/blob/927c021671972ac39c98b296fd97e4222d9bb700/sprokit/processes/core/downsample_process.cxx#L134-L139

The timestamp didn't need to have a valid time, in which case `skip_frame` was written to estimate it as the previous time plus `1 / frame_rate`.

However, it now looks like a lot of this has changed. The `timestamp` and `frame_rate` input ports are now optional, and associated changes mean:
- `skip_frame` is no longer called every frame, but only when the timestamp has a valid frame or time
- `skip_frame` can be called when the frame rate is not provided or when the frame rate port stops earlier than the timestamp port. In either case, `skip_frame` will be given an invalid value of `-1` for the frame rate.

Those are all potentially sources of correctness issues, and I think the most important realizations in practice are likely:
- If the timestamp port is not connected, no frames will be skipped
- If the timestamp port is receiving frame-only timestamps but the frame rate port is not connected, `skip_frame` will try to use an invalid negative frame rate. In particular, `ds_frame_time_` will tick backwards and every frame except the first will be skipped.

It looks like no-skipping can already happen on `viame/master`, for what that's worth. I'm not sure how `skip_frame` handles a negative frame rate on `viame/master`.

@mattdawkins, what behavior do we want when the timestamp or frame rate ports aren't connected? I also think it would help to document what the intended behavior is.

mattdawkins:
In all viame cases they're connected; gotta get something out with this fix today (and 5 hr lack time to get it into VIAME-web before nightly update) so merging for now, can revisit later and make sure viame downsampler and kwiver master downsampler in sync

PR #1177, opened 2021-01-13
Add high pass filter from VIDTK to support BurnOut

russelldj:
This is the same code as reviewed internally [here](https://kwgitlab.kitware.com/computer-vision/KWIVER/-/merge_requests/260#c339234bff7b638bdc8c8eb20053542b2ced1f81).


russelldj:
I had to rewrite history with respect to the internally-reviewed to satisfy `ghostflow`. I can additionally squash everything if that's useful, given it's all new content.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,478 @@
...
+    const unsigned ni = grey.ni();
+    const unsigned nj = grey.nj();
+
+    output.fill( 0 );
+
+    unsigned offset = ( kernel_width / 2 ) + 1;
+
+    unsigned diff1 = 0, diff2 = 0;
+
+    {
```
mleotta, line 90: What is the purpose of this new scope here?  It shouldn't break anything but it looks extraneous.

arrows/vxl/high_pass_filter.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <vital/algo/image_filter.h>
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace vxl {
+
+/**
```
mleotta, line 18: You might consider switching this to the new standard for KWIVER doxygen:
```
 /// VXL High Pass Filtering Process
 ///
 /// This method contains basic methods for high pass image filtering
 /// on top of input images
 ```

**Review by mleotta — CHANGES_REQUESTED** [review]
I've pointed out two very minor things, but otherwise it looks good.

I'd also recommend a very basic unit test just to make sure that
1. the output of the filter looks reasonable now, and
2. the code continues to produce that same output when changed in the future.

The test can be as simple as loading a test image, apply the filter, load in the result image and compare the values.

File: arrows/vxl/high_pass_filter.cxx [hunk 3]
```diff
@@ -0,0 +1,478 @@
...
+#include <vil/vil_math.h>
+#include <vil/vil_plane.h>
+#include <vil/vil_transpose.h>
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace vxl {
+
```
linus-sherrill, line 19: This is purely optional, so feel free to ignore, but there is support for 1 of n options, such as the "mode" parameter.

Need to include "enum_converter.h"
```suggestion
class enum filter_mode {
    MODE_box,
    MODE_bidir };
    
    ENUM_CONVERTER( mode_converter, filter_mode,
    { "box", MODE_box },
    { "bidir", MODE_bidir } )
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,478 @@
...
+    : p( parent )
+  {
+  }
+
+  ~priv()
+  {
+  }
+
+  // Internal parameters/settings
+  std::string mode = "box";
```
linus-sherrill, line 36: Option impl.
```suggestion
'filter_mode mode = MODE_box;`
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,478 @@
...
+
+// ----------------------------------------------------------------------------
+vital::config_block_sptr
+high_pass_filter
+::get_configuration() const
+{
+  // get base config from base class
+  vital::config_block_sptr config = algorithm::get_configuration();
+
+  config->set_value( "mode", d->mode,
```
linus-sherrill, line 362: optional impl
```suggestion
  config->set_value( "mode",  mode_converter().to_string( d->method )
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,478 @@
...
+::set_configuration( vital::config_block_sptr in_config )
+{
+  // Starting with our generated vital::config_block to ensure that assumed
+  // values are present. An alternative is to check for key presence before
+  // performing a get_value() call.
+  vital::config_block_sptr config = this->get_configuration();
+  config->merge_config( in_config );
+
+  // Settings for filtering
+  d->mode = config->get_value< std::string >( "mode" );
```
linus-sherrill, line 392: optional impl
```suggestion
  d->mode = config->get_enum_value< mode_converter >( "mode" );
```
This will throw an exception listing the allowable values if an unknown value is in the config.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,478 @@
...
+}
+
+// ----------------------------------------------------------------------------
+bool
+high_pass_filter
+::check_configuration( vital::config_block_sptr config ) const
+{
+  std::string mode_string = config->get_value< std::string >( "mode" );
+
+  if( !( mode_string == "box" ||
```
linus-sherrill, line 406: optional impl. No need to validate config item. That is done by the converter/when the config item is converted from string to enum.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,478 @@
...
+    }
+  }
+
+  // Apply a high pass filter to one frame and return the output
+  template < typename PixType >
+  kwiver::vital::image_container_sptr
+  filter( vil_image_view< PixType >& input )
+  {
+    vil_image_view< PixType > output;
+    if( mode == "box" )
```
linus-sherrill, line 316: optional impl. just use a switch statement on the enum. The enum conversion process guarantees that the enum is valid.

linus-sherrill (changes requested):
The only substantial comment is about deprecated algorithm_impl. I had the time and thought I would socialize the enum converter. Its use is not required. If you find it useful, there are other places where it can be used.

russelldj:
> I've pointed out two very minor things, but otherwise it looks good.
> 
> I'd also recommend a very basic unit test just to make sure that
> 
> 1. the output of the filter looks reasonable now, and
> 2. the code continues to produce that same output when changed in the future.
> 
> The test can be as simple as loading a test image, apply the filter, load in the result image and compare the values.

I've implemented the suggested test with four variants, representing the two filtering methods and default and larger filter sizes. The input image and test cases are attached below.

input:
small_grey_logo
box filter:
box
box filter wide:
box_wide
bidirectional filter:
bidir
bidirectional filter wide:
bidir_wide

russelldj:
This testing appeared to show that this code only works for one channel images and throws a `vil` runtime error otherwise. If this is not acceptable behavior, should the input image be converted to greyscale before processing. Or a more descriptive error thrown? Or something else?

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,477 @@
...
+vital::config_block_sptr
+high_pass_filter
+::get_configuration() const
+{
+  // get base config from base class
+  vital::config_block_sptr config = algorithm::get_configuration();
+
+  config->set_value( "mode", mode_converter().to_string(
+                       d->mode ),
+                     "Operating mode of this filter, possible values: box, bidir" );
```
linus-sherrill, line 371: I missed this on the first comment, but there is a way to automatically insert the options into the help string.
```suggestion
                     "Operating mode of this filter, possible values: " 
                     + mode_converter().element_name_string()  );
```

mleotta:
I think it makes sense to convert RGB to grey if the input is three channels.  If the number of channels is not 1 or 3 we should still throw an exception, but it would be better to throw a more meaningful vital exception than to let vil throw a runtime error.

Also, it appears that box and bidirectional filters have the first two channels swapped relative to each other.  That is, in box the first channel response more to vertical features and the second to horizontal.  In bidirectional it is the opposite.  It's probably always been this way, and I'm not sure if this matters with how we are using it.  I think we mostly care about the third combined channel, but it might be worth making them consistent.

russelldj:
I've updated the channel ordering and made fixed the issue with multiple channels. Now the `R` channel represents the response in the horizontal direction. I've attached the updated box filters below. At this point I think it's ready to merge.
box
box_wide

linus-sherrill (approved):

russelldj:
From discussion with @mleotta it sounded like review from Chet and Matt W. was optional. Would someone with permission merge this?

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+    vil_image_view< PixType > filter_x = vil_plane( output, 0 );
+    vil_image_view< PixType > filter_y = vil_plane( output, 1 );
+    vil_image_view< PixType > filter_xy = vil_plane( output, 2 );
+
+    box_average_horizontal( grey_img, filter_x, kernel_width );
+    box_average_vertical( grey_img, filter_y, kernel_height );
+
+    // apply horizontal smoothing to the vertically smoothed image to get a 2D
+    // box filter
+    box_average_horizontal( filter_x, filter_xy, kernel_width );
```
mleotta, line 71: Maybe this got swapped when you reordered the first two channels, but it looks like this is now two horizonal passes on the image instead of one horizontal and one vertical

mleotta: Yes, it looks like that's what happened.  Can you change the first input to `filter_y` here?

russelldj: Thanks for the catch. I updated it and added the updated expected test results.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+    vil_image_view< PixType > output;
+    switch( mode )
+    {
+      case MODE_box:
+        output = box_high_pass_filter( input );
+        break;
+      case MODE_bidir:
+        output = bidirection_box_filter( input );
+        break;
+      default:
```
chetnieter, line 333: Is an error or warning message needed here or is the `default` case effectively unreachable and only here to prevent compiler warnings?

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+namespace vxl {
+
+enum filter_mode
+{
+  MODE_box,
+  MODE_bidir,
+};
+
+ENUM_CONVERTER( mode_converter, filter_mode, { "box", MODE_box }, { "bidir",
+                                                                    MODE_bidir } )
```
mwoehlke-kitware: This is... interesting line breaking, and also over length. Maybe this would make more sense?
```suggestion
ENUM_CONVERTER( mode_converter, filter_mode,
                { "box", MODE_box }, { "bidir", MODE_bidir } )
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+
+// --------------------------------------------------------------------------------------
+/// Private implementation class
+class high_pass_filter::priv
+{
+public:
+
```
mwoehlke-kitware: I'd avoid the blank line here. However, while we don't "typically", given the number and size of member functions with definitions inside the class definition, separator lines might be helpful:
```suggestion
  // --------------------------------------------------------------------------
```
(...and before other methods.)

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+  priv( high_pass_filter* parent )
+    : p( parent )
```
mwoehlke-kitware, line 37:
```suggestion
    : p{ parent }
```

arrows/vxl/high_pass_filter.cxx
chetnieter, line 29: nitpick - maybe make the line break at the first `{` to make it more readable?

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+  size_t kernel_width = 7;
+  size_t kernel_height = 7;
+  bool treat_as_interlaced = false;
+  bool output_net_only = false;
+  high_pass_filter* p;
+
+  // Perform box filtering
+  template < typename PixType >
+  vil_image_view< PixType >
+  box_high_pass_filter( const vil_image_view< PixType >& grey_img )
```
mwoehlke-kitware, line 56:
```suggestion
  box_high_pass_filter( vil_image_view< PixType > const& grey_img )
```
(...and other occurrences, unsurprisingly :slightly_smiling_face:.)

arrows/vxl/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@
 # structures
 
 set(vxl_headers_public
+  high_pass_filter.h
```
chetnieter, line 5: nitpack - list files alphabetically

arrows/vxl/CMakeLists.txt
```diff
@@ -35,6 +36,7 @@ kwiver_install_headers(
   )
 
 set(vxl_sources
+  high_pass_filter.cxx
```
chetnieter, line 39: nitpack - list files alphabetically

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+  {
+    vil_image_view< PixType > output;
+    output.set_size( grey_img.ni(), grey_img.nj(), 3 );
```
mwoehlke-kitware, line 59: Not strictly necessary, but I think these can be rewritten as:
```suggestion
    vil_image_view< PixType > output{ grey_img.ni(), grey_img.nj(), 3 };
```
...? (Unless this way is preferred for some reason?)

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+
+    vil_image_view< PixType > filter_x = vil_plane( output, 0 );
+    vil_image_view< PixType > filter_y = vil_plane( output, 1 );
+    vil_image_view< PixType > filter_xy = vil_plane( output, 2 );
+
+    box_average_horizontal( grey_img, filter_x, kernel_width );
+    box_average_vertical( grey_img, filter_y, kernel_height );
+
+    // apply horizontal smoothing to the vertically smoothed image to get a 2D
```
mwoehlke-kitware, line 68: Pedantic:
```suggestion
    // Apply horizontal smoothing to the vertically smoothed image to get a 2D
```

arrows/vxl/tests/test_high_pass_filter.cxx
```diff
@@ -0,0 +1,152 @@
...
+
+  return RUN_ALL_TESTS();
+}
+
+// ----------------------------------------------------------------------------
+class high_pass_filter : public ::testing::Test
+{
+  TEST_ARG( data_dir );
+};
+
```
chetnieter, line 49: Are there any failure modes that need to be tested in addition to the various filter modes?

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+
+    return output;
+  }
+
+  // Given an input grayscale image, and a smoothed version of this greyscale
+  // image, calculate the bidirectional filter response in the horizontal
+  // direction.
+  template < typename PixType >
+  void
+  horizontal_box_bidirectional_pass( const vil_image_view< PixType >& grey,
```
mwoehlke-kitware, line 86: Totally up to you, but maybe this is a spot to break after the `(` and only indent the parameters by the indent level (two spaces)? The parameter list would then also fit on fewer lines.

Both are acceptable; up to you which you prefer.

chetnieter (approved):
Looks fine to me. I had a few formatting suggestions for you to consider but otherwise it looks good.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+
+    unsigned offset = ( kernel_width / 2 ) + 1;
+
+    unsigned diff1 = 0, diff2 = 0;
+
+    if( ni < 2 * offset + 1 )
+    {
+      return;
+    }
+
```
mwoehlke-kitware, line 104: Being not a huge fan of multiple declarations on one line...
```suggestion
    unsigned offset = ( kernel_width / 2 ) + 1;

    if( ni < 2 * offset + 1 )
    {
      return;
    }

    unsigned diff1 = 0;
    unsigned diff2 = 0;
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,490 @@
...
+enum filter_mode
+{
+  MODE_box,
+  MODE_bidir,
+};
+
+ENUM_CONVERTER( mode_converter, filter_mode, { "box", MODE_box }, { "bidir",
+                                                                    MODE_bidir } )
+
+// --------------------------------------------------------------------------------------
```
mwoehlke-kitware, line 31:
```suggestion
// ----------------------------------------------------------------------------
```
(Seems to be the only one that's the wrong length, or at least the only one I noticed.)

arrows/vxl/high_pass_filter.cxx
russelldj: Yes, it's just for the compiler. Linus suggested using the `ENUM_CONVERTER` approach since it does quality error checking when there's a known set of values.

arrows/vxl/high_pass_filter.cxx
russelldj: Thanks. @mwoehlke-kitware for some reason `uncrustify` keeps changing this line for fairly reasonable formatting styles, including what Chet suggested. I've been trying to avoid commiting anything that uncrusify will change in the future. Is that a reasonable practice?

arrows/vxl/tests/test_high_pass_filter.cxx
russelldj: I couldn't think of anything terribly important but I'd welcome suggestions if you have any. As an aside, unit tests on a number of these related PRs could be simplified but I'd prefer to go back and update them as a batch on a separate PR since I think they're fine now.

mwoehlke-kitware (commented):
This is a partial review because I want to reply to another conversation, and can't while I have a pending review...

arrows/vxl/high_pass_filter.cxx
mwoehlke-kitware: I've been treating uncrustify as a recommendation, and committing things that are clearly correct even if they get changed. In this case, the issue seems to be with `nl_type_brace_init_lst` and is almost certainly a bug. Besides, what it's trying to do has the additional annoying property of being over length.

I haven't tried to minimize the config yet, but just those lines are minimal input for demonstrating the issue.

russelldj: I tried what you suggested and `uncrustify` reverted it to what's shown currently. The best I found, which I added in the last commit, was the following:
```
ENUM_CONVERTER( mode_converter,
                filter_mode, { "box", MODE_box }, { "bidir", MODE_bidir } )
```

arrows/vxl/high_pass_filter.cxx
mwoehlke-kitware: My bad; this should actually be:
```suggestion
    auto const offset = 
      static_cast< decltype( ni ) >( ( kernel_width / 2 ) + 1 );

    if( ni < 2 * offset + 1 )
    {
      return;
    }

    PixType diff1 = 0;
    PixType diff2 = 0;
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+  vertical_box_bidirectional_pass( const vil_image_view< PixType >& grey,
+                                   const vil_image_view< PixType >& smoothed,
+                                   vil_image_view< PixType >& output,
+                                   unsigned kernel_height )
+  {
+    vil_image_view< PixType > grey_t = vil_transpose( grey );
+    vil_image_view< PixType > smoothed_t = vil_transpose( smoothed );
+    vil_image_view< PixType > output_t = vil_transpose( output );
+
+    horizontal_box_bidirectional_pass( grey_t,
```
mwoehlke-kitware, line 135: Probably don't need so many line breaks...

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+    output.set_size( grey_img.ni(), grey_img.nj(), 3 );
+
+    // Create views of each plane
+    vil_image_view< PixType > filter_x = vil_plane( output, 0 );
+    vil_image_view< PixType > filter_y = vil_plane( output, 1 );
+    vil_image_view< PixType > filter_xy = vil_plane( output, 2 );
+
+    // Report the difference between the pixel value, and all of the smoothed
+    // responses. The xy channel is used as a temporary buffer to avoid
+    // additional memory allocation
```
mwoehlke-kitware, line 156:
```suggestion
    // Report the difference between the pixel value, and all of the smoothed
    // responses. The xy channel is used as a temporary buffer to avoid
    // additional memory allocation.
```

...or...

```suggestion
    // Report the difference between the pixel value, and all of the smoothed
    // responses, using the xy channel as a temporary buffer to avoid
    // additional memory allocation
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+  // Function which returns an image view of the even rows of an image
+  template < typename PixType >
+  inline vil_image_view< PixType >
+  even_rows( const vil_image_view< PixType >& im )
+  {
+    return vil_image_view< PixType >( im.memory_chunk(), im.top_left_ptr(),
+                                      im.ni(), ( im.nj() + 1 ) / 2,
+                                      im.nplanes(),
+                                      im.istep(), im.jstep() * 2,
+                                      im.planestep() );
```
mwoehlke-kitware, line 177:
```suggestion
    return vil_image_view< PixType >{
      im.memory_chunk(), im.top_left_ptr(),
      im.ni(), ( im.nj() + 1 ) / 2, im.nplanes(),
      im.istep(), im.jstep() * 2, im.planestep() };
```

As previously noted, up to you where you prefer to put the line breaks; this one is very much a "have you considered?". Using `{}`s would be nice, though :slightly_smiling_face:. (Same thing a few lines below.)

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+      LOG_ERROR( p->logger(), "Kernel width must be odd" );
+    }
+
+    if( kernel_width >= src.ni() )
+    {
+      // Force width to be smaller than image width and odd
+      kernel_width = ( src.ni() == 1 ? 1 : ( src.ni() - 2 ) | 0x01 );
+    }
+
+    unsigned ni = src.ni(), nj = src.nj(), np = src.nplanes();
```
mwoehlke-kitware, line 214: Per prior comment:
```suggestion
    auto const ni = src.ni();
    auto const nj = src.nj();
    auto const np = src.nplanes();
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+    dst.set_size( ni, nj, np );
+
+    unsigned half_width = kernel_width / 2;
+
+    std::ptrdiff_t istepS = src.istep(),  jstepS = src.jstep(),
+                   pstepS = src.planestep();
+    std::ptrdiff_t istepD = dst.istep(),  jstepD = dst.jstep(),
+                   pstepD = dst.planestep();
```
mwoehlke-kitware, line 222: ...and again:
```suggestion
    auto const istepS = src.istep();
    auto const jstepS = src.jstep();
    auto const pstepS = src.planestep();
    auto const istepD = dst.istep();
    auto const jstepD = dst.jstep();
    auto const pstepD = dst.planestep();
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+
+      for( unsigned j = 0; j < nj; ++j, rowS += jstepS, rowD += jstepD )
+      {
+        const PixType* pixelS1 = rowS;
+        const PixType* pixelS2 = rowS;
+        PixType*       pixelD = rowD;
+
+        // fast box filter smoothing by adding one pixel to the sum and
+        // subtracting another pixel at each step
+        unsigned sum = 0;
```
mwoehlke-kitware, line 240: Okay... I'm not immediately sure what to do with this (might want to ask the original author, is that Dawkins?), but this is triggering float-to-int warnings that are concerning, as in "this might be properly broken". However, I also don't think we want to just use `PixType` because I get the impression that could lead to overflow and would be equally broken. I wonder if perhaps this needs some finessing to choose an appropriate intermediate type.

mwoehlke-kitware (commented):
Still a WIP, but please see note on L240; that one looks like it might be an actual issue with the algorithm and not just pedantry.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+  }
+
+  ~priv()
+  {
+  }
+
+  // Internal parameters/settings
+  filter_mode mode = MODE_box;
+  size_t kernel_width = 7;
+  size_t kernel_height = 7;
```
mwoehlke-kitware, line 48:
```suggestion
  unsigned kernel_width = 7;
  unsigned kernel_height = 7;
```
Everywhere that these are actually used seems to use `unsigned`, and they relate to the image dimensions which are also limited to `unsigned`, so there doesn't seem to be much value to using `size_t`. (Besides, I suspect using a kernel size > 2³² is going to be problematic anyway...)

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+  // Starting with our generated vital::config_block to ensure that assumed
+  // values are present. An alternative is to check for key presence before
+  // performing a get_value() call.
+  vital::config_block_sptr config = this->get_configuration();
+  config->merge_config( in_config );
+
+  // Settings for filtering
+  d->mode = config->get_enum_value< mode_converter >( "mode" );
+  d->kernel_width = config->get_value< size_t >( "kernel_width" );
+  d->kernel_height = config->get_value< size_t >( "kernel_height" );
```
mwoehlke-kitware, line 403: Per related, earlier suggestion:
```suggestion
  d->kernel_width = config->get_value< unsigned >( "kernel_width" );
  d->kernel_height = config->get_value< unsigned >( "kernel_height" );
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+    return kwiver::vital::image_container_sptr();
+  }
+
+#define HANDLE_CASE( T )                                                 \
+  case T:                                                                \
+  {                                                                      \
+    typedef vil_pixel_format_type_of< T >::component_type pix_t;         \
+    auto input = static_cast< vil_image_view< pix_t > >( *view );        \
+    kwiver::vital::image_container_sptr output = d->filter( input );     \
+    return output;                                                       \
```
mwoehlke-kitware, line 458:
```suggestion
    return output = d->filter( input ); \
```
...and fix up the `\` alignment if you prefer keeping them that way.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+  }                                                                      \
```
mwoehlke-kitware, line 459:
```suggestion
  }
```

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,489 @@
...
+  {
+    LOG_ERROR( logger(), "Expected 1 or 3 channels but recieved "
+               << view->nplanes() );
+    return kwiver::vital::image_container_sptr();
+  }
+
+#define HANDLE_CASE( T )                                                 \
+  case T:                                                                \
+  {                                                                      \
+    typedef vil_pixel_format_type_of< T >::component_type pix_t;         \
```
mwoehlke-kitware, line 455:
```suggestion
    using pix_t = vil_pixel_format_type_of< T >::component_type;         \
```

mwoehlke-kitware (commented):
(Still WIP)

arrows/vxl/high_pass_filter.cxx
russelldj: This looked quite messy to me so I pulled the definitions outside of the class definition.

arrows/vxl/high_pass_filter.cxx
russelldj: Yes, I believe @mattdawkins wrote the majority of this. I assume using a `double` is just inefficient and could have accuracy issues for integral types? I'm going to look at VXL to see if there's either a drop-in replacement or something that can be used as a reference. 

Also, this might be a bit of a cop-out but at least for burn-out this should only take `VXL_BYTE` images as input.

mwoehlke-kitware: I was thinking of using `double` for any floating point and `int64` otherwise. We might want to remove support for 64-bit (integer) input, but the others should be passably okay with such an approach.

mattdawkins: Burn-Out (and specifically this adaboost approach) is heavily optimized for and has really only ever had 8-bit RGB inputs (though greyscale was supported as well in old as well). You might just want to add an error early on checking for this as opposed to over-optimizing for cases we don't care about, for filters which require it. There might be one other.

mwoehlke-kitware:
@russelldj, I pushed a fix for the accumulator type; PTAL.

mwoehlke-kitware (approved):

mwoehlke-kitware:
@mleotta, your objection is buried in the comment history somewhere; has it been addressed?

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,564 @@
...
+  MODE_bidir,
+};
+
+ENUM_CONVERTER( mode_converter, filter_mode,
+                { "box", MODE_box }, { "bidir", MODE_bidir } )
+
+// ----------------------------------------------------------------------------
+template <typename T> struct accumulator;
+
+template <> struct accumulator<char>
```
russelldj, line 36: Pedantic, but my understanding is the whitespace between opening and closing <>.

Also, I find particular line is a bit confusing. My understanding is char can be signed or unsigned depending on implementation but int is always signed. So maybe if my understanding is correct, being consistent across this and the following line would be ever so slightly clearer.

arrows/vxl/high_pass_filter.cxx
```diff
@@ -0,0 +1,564 @@
...
+{ using type = int; };
```
russelldj, line 37: Pedantic, but my understanding is the whitespace between opening and closing <>.

Also, I find particular line is a bit confusing. My understanding is char can be signed or unsigned depending on implementation but int is always signed. So maybe if my understanding is correct, being consistent across this and the following line would be ever so slightly clearer.

mleotta:
Jenkins test this please

mleotta (approved):

arrows/vxl/high_pass_filter.cxx
mwoehlke-kitware: Derp, and there should be a line break after `template <>`.

I was actually thinking that "signed char" → "signed int" and "char" → "int" *was* consistent :slightly_smiling_face:. I suppose more consistent would be to make `accumulator<char>::type` be signed or unsigned to match `char`, but that complicated to at best gain us an extra bit (since `char` can always promote to `int`).

russelldj: Yeah, no worries with how it is now.

mleotta:
Jenkins test this please

arrows/vxl/high_pass_filter.cxx
```diff
@@ -188,9 +188,9 @@ ::horizontal_box_bidirectional_pass( vil_image_view< PixType > const& grey,
   PixType diff1 = 0;
   PixType diff2 = 0;
 
-  for( decltype( +nj ) j{ 0 }; j < nj; ++j )
+  for( unsigned j = 0; j < nj; ++j )
```
mwoehlke-kitware, line 191: ...as long as `nj` is also `unsigned`, I am okay with this; I mainly don't want to have `size_t` vs. `unsigned` mismatches.

russelldj:
There's no indication that Jenkins is queued. Is this the issue @dstoup mentioned?

mleotta:
Yeah, Jenkins is ignoring my requests on this branch for some reason. @dstoup @Cookt2 any idea why?

mleotta:
Jenkins test this please

PR #1178, opened 2021-01-13
Fix init landmarks bug

mleotta:


mleotta:
This is a simple (one line) fix, but it's important and I'd like to merge it soon.

In summary, data to optimize was passed in by reference, contents are unpacked into local variables, local variables are optimized, but we forgot to package up the result into the original variable for return by reference.

chetnieter (approved):

PR #1179, opened 2021-01-14
Added option to ignore metadata

mleotta:
This new option is added to ignore KLV metadata in files where it is known to be wrong.  It also helps with large files that are known to not contain any metadata where you would otherwise need to spend a long time scanning the whole file to find nothing.


russelldj:
Looks good. Just as an aside, the style guide suggests adding whitespace inside of all matching braces, but it looks like the surrounding code mostly doesn't do that.

russelldj (approved):

mleotta:
Thanks for pointing out the style issues.  Since most of the code is in violation, I think it's best that we get our formatting tool working an then fix this file (and others) in batch.

PR #1180, opened 2021-01-14
Avoid unnecessary computation

mleotta:
This branch removes some significant wasted computation by detecting and terminating futile attempts earlier.


chetnieter (approved):
Looks good to me

PR #1182, opened 2021-01-14
Use OpenMP to compute uncertainty in parallel

mleotta:
This new uncertainty function has become a bottleneck when running on multicore systems.  This change applies OpenMP to compute the uncertainty image in parallel the same way we do for several other computations in this file.

### DIFF
--- a/arrows/super3d/compute_depth.cxx
+++ b/arrows/super3d/compute_depth.cxx
@@ -200,7 +200,8 @@ compute_uncertainty(vil_image_view<double> const& height_map,
   // This scale is 1/(2*sigma) converted from [0,255] to [0,1]
   const double cost_scale = 255.0 / (2.0 * 5.0);
 
-  for (unsigned int j = 0; j < cost_volume.nj(); j++)
+#pragma omp parallel for
+  for (int64_t j = 0; j < cost_volume.nj(); j++)
   {
     for (unsigned int i = 0; i < cost_volume.ni(); i++)
     {


arrows/super3d/compute_depth.cxx
```diff
@@ -200,7 +200,8 @@ compute_uncertainty(vil_image_view<double> const& height_map,
   // This scale is 1/(2*sigma) converted from [0,255] to [0,1]
   const double cost_scale = 255.0 / (2.0 * 5.0);
 
-  for (unsigned int j = 0; j < cost_volume.nj(); j++)
+#pragma omp parallel for
+  for (int64_t j = 0; j < cost_volume.nj(); j++)
   {
     for (unsigned int i = 0; i < cost_volume.ni(); i++)
```
russelldj, line 206: Should this be updated to `int64_t` as well? Also, I'd assume we can get sufficient parallelism across the `j` channel and don't need to parallelize across `i` as well?

russelldj (approved):

arrows/super3d/compute_depth.cxx
mleotta: Ideally both should be `unsigned int` or maybe `size_t`, but for some reason OpenMP requires parallel loop variables to be `int64_t`, which is why this one is changed here.

The `j` channel is enough as long as your image has more rows than your processor has cores, which is always the case in practice.  On my desktop with 20 cores (40 logical processors) I use 100% of the CPU after this change and only about 5% before it.

PR #1186, opened 2021-01-14
Add convert image vxl filter

russelldj:
Part of #1157 which is being broken up. This adds a variety of per-pixel filtering operations.


arrows/vxl/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(vxl_headers_public
   camera.h
   camera_map.h
   close_loops_homography_guided.h
+  convert_image.h
```
mwoehlke-kitware, line 10: pedantic: 'n' ('convert') > 'm' ('compute')

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+          *dpixel = max_output_value;
+        }
+      }
+    }
+  }
+}
+
+template < typename Type >
+void
+combine_channels( const vil_image_view< Type >& src,
```
mwoehlke-kitware, line 79: pedantic: please use `T const` instead of `const T`

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+  {
+    dst = src;
+  }
+}
+
+// Calculate the values of our image percentiles from x sampling points
+template < typename PixType >
+void
+sample_and_sort_image( const vil_image_view< PixType >& src,
+                       std::vector< PixType >& dst,
```
mwoehlke-kitware, line 122: Ugh, why is this not a return value?

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+      }
+    }
+  }
+}
+
+template < typename PixType >
+void
+get_image_percentiles( const vil_image_view< PixType >& src,
+                       const std::vector< double >& percentiles,
+                       std::vector< PixType >& dst,
```
mwoehlke-kitware, line 189: Same here, why not a return value?

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+                               0.5 );
+    dst[ i ] = sorted_samples[ ind ];
+  }
+}
+
+template < typename PixType >
+void
+percentile_threshold_above( const vil_image_view< PixType >& src,
+                            const std::vector< double >& percentiles,
+                            vil_image_view< bool >& dst,
```
mwoehlke-kitware, line 214: ...and here too

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+               p ) =
+            static_cast< OutputType >( ( *pixel - lower_bound ) *
+                                       scale );
+        }
+      }
+    }
+  }
+}
+
+} // end anonoymous namespace
```
mwoehlke-kitware, line 300: pedantic: `} // namespace <anonymous>`

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+
+// ----------------------------------------------------------------------------
+/// Private implementation class
+class convert_image::priv
+{
+public:
+
```
mwoehlke-kitware, line 307: pedantic: no blank line here

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+  priv()
+    : format( "byte" )
+      , single_channel( false )
+      , scale_factor( 0.0 )
+      , random_greyscale( 0.0 )
+      , percentile_norm( -1.0 )
+  {
+  }
+
+  ~priv()
```
mwoehlke-kitware, line 317: pedantic: an explicitly empty dtor in a class defined in a `.cxx` is rather pointless

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,508 @@
...
+  config->set_value( "single_channel", d->single_channel,
+                     "Convert input (presumably multi-channel) to contain a single channel, using "
+                     "either standard RGB to grayscale conversion weights, or averaging." );
+  config->set_value( "scale_factor", d->scale_factor,
+                     "Optional input value scaling factor" );
+  config->set_value( "random_greyscale", d->random_greyscale,
+                     "Convert input image to a 3-channel greyscale image randomly with this percentage "
+                     "between 0.0 and 1.0. This is used for machine learning augmentation." );
+  config->set_value( "percentile_norm", d->percentile_norm,
+                     "If set, perform percentile normalization such that the output image's min and max "
```
mwoehlke-kitware, line 359: I don't think this is worded well. It sounds like this *reduces* the image contrast?

mwoehlke-kitware (commented):
Disclaimer: this wasn't a very thorough review

arrows/vxl/convert_image.cxx
russelldj: Could this be tool-enforced?

russelldj: I'm guessing based on your [open uncrustify feature request](https://github.com/uncrustify/uncrustify/issues/1952), that it can't at the moment.

arrows/vxl/convert_image.cxx
russelldj: This function took me longer than I'd have liked to interpret. It basically uses this provided `fraction`, which is between [`0.0`,` 0.5`), so right of the bat the name is missleading. Then it finds the quintiles at `fraction` and `1 - fraction`. Then the output image is scaled and clamped so that the lower quintile is scaled to the lower bound of the output and the upper quintile is scaled to the upper bound.

 I'm not sure if my updated description is effective at describing the behavior so I'd welcome other suggestions.

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,498 @@
...
+  double scale_factor;
+  double random_greyscale;
+  double percentile_norm;
+};
+
+// ----------------------------------------------------------------------------
+template < typename ipix_t >
+vil_image_view< ipix_t >
+convert_image::priv
+::convert( vil_image_view_base_sptr& view )
```
mleotta, line 298: Is this called from anywhere?

mleotta: If not, then the functions (`random_grey_conversion`, and `combine_channels`) and config options (`single_channel`) used below are also not used anywhere else that I can see.

mleotta (changes requested):
Looks okay, but I don't see any thing calling `convert()`.  Should this code path be removed, or does it need to be called from somewhere?

arrows/vxl/convert_image.cxx
russelldj: Yes, in line 454 in this file. I've updated it to be named in a more informative and distinct way.

mleotta (approved):

mleotta (changes requested):

mleotta:
Jenkins test this please

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+  {
+    for( unsigned i = 1; i < dst.size(); ++i )
+    {
+      if( dst[ i ] != 0 )
+      {
+        dst.erase( dst.begin(), dst.begin() + i );
+        break;
+      }
+    }
+  }
```
mwoehlke-kitware, line 170:
```suggestion
  if( remove_extremes )
  {
    constexpr auto low = PixType{ 0 };
    while( !dst.empty() && dst.front() == low )
    {
      dst.erase(dst.begin());
    }

    constexpr auto high = std::numeric_limits< PixType >::max();
    while( !dst.empty() && dst.back() == high )
    {
      dst.pop_back();
    }
  }
```

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+  unsigned sampling_points, bool remove_extremes )
+{
+  std::vector< PixType > sorted_samples =
+    sample_and_sort_image( src, sampling_points, remove_extremes );
+
+  std::vector< PixType > dst( percentiles.size() );
+  double sampling_points_minus1 =
+    static_cast< double >( sorted_samples.size() - 1 );
+
+  for( unsigned i = 0; i < percentiles.size(); ++i )
```
mwoehlke-kitware, line 190: All the `unsigned` here should probably be `size_t`. (Also, this sort of thing is why I love `kwiver::vital::range::iota`; it makes it much easier to use `auto`. :slightly_smiling_face:)

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+
+  InputType lower_bound = percentile_values[ 0 ];
+  InputType upper_bound = percentile_values[ 1 ];
+
+  double scale;
+
+  if( percentile_values[ 1 ] - percentile_values[ 0 ] > 0 )
+  {
+    scale = ( static_cast< double >( max_val ) + 0.5 ) /
+            ( percentile_values[ 1 ] - percentile_values[ 0 ] );
```
mwoehlke-kitware, line 230:
```suggestion
    scale = ( static_cast< double >( max_val ) + 0.5 ) /
            static_cast< double >( percentile_values[ 1 ] -
                                   percentile_values[ 0 ] );
```

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+  }
+  else
+  {
+    scale = static_cast< double >( max_val ) /
+            std::numeric_limits< InputType >::max();
```
mwoehlke-kitware, line 235:
```suggestion
    scale = static_cast< double >( max_val ) /
            static_cast< double >( std::numeric_limits< InputType >::max() );
```

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+                   {
+                     return static_cast< OutputType >( std::numeric_limits< OutputType >
+                                                       ::max() );
+                   }
+                   else
+                   {
+                     return static_cast< OutputType >( ( pixel -
+                                                         lower_bound ) * scale );
+                   }
+                 } );
```
mwoehlke-kitware, line 260:
```suggestion
  vil_transform(
    src, dst,
    [ lower_bound, upper_bound, scale ]( InputType pixel ){
      if( pixel < lower_bound )
      {
        return static_cast< OutputType >( 0 );
      }
      else if( pixel > upper_bound )
      {
        return static_cast< OutputType >(
          std::numeric_limits< OutputType >::max() );
      }
      else
      {
        return static_cast< OutputType >(
          static_cast< double >( pixel - lower_bound ) * scale );
      }
    } );
```

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+    : format{ "byte" }
+      , single_channel{ false }
+      , scale_factor{ 0.0 }
+      , random_greyscale{ 0.0 }
+      , percentile_norm{ -1.0 }
+  {
+  }
+
+  // Apply appropriate transforms
+  template < typename ipix_t > vil_image_view< ipix_t >
```
mwoehlke-kitware, line 280: The `template` spec should probably not be on the same line as the return type. (In both cases, the return plus name and argument list fit on the same line, but breaking after the return type is also okay; feel free to pick whichever you prefer.)

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+// ----------------------------------------------------------------------------
+template < typename ipix_t >
+vil_image_view< ipix_t >
+convert_image::priv
+::apply_transforms( vil_image_view_base_sptr& view )
+{
+  vil_image_view< ipix_t > output;
+  if( random_greyscale > 0.0 )
+  {
+    vil_image_view< ipix_t > tmp{ view };
```
mwoehlke-kitware, line 303: BTW, do you know why this line is needed?

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+};
+
+// ----------------------------------------------------------------------------
+template < typename ipix_t >
+vil_image_view< ipix_t >
+convert_image::priv
+::apply_transforms( vil_image_view_base_sptr& view )
+{
+  vil_image_view< ipix_t > output;
+  if( random_greyscale > 0.0 )
```
mwoehlke-kitware, line 301: It seems a little odd that `random_greyscale` would take precedence over `single_channel`...

(Also, as something of an aside, "gray[scale]" seems to be the preferred en_US spelling.)

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+vil_image_view< OutType >
+scale_image( vil_image_view< InType > const& src,
+             double const& dp_scale )
+{
+  auto const ni = src.ni();
+  auto const nj = src.nj();
+  auto const np = src.nplanes();
+  vil_image_view< OutType > dst{ ni, nj, np };
+
+  OutType const max_output_value = std::numeric_limits< OutType >::max();
```
mwoehlke-kitware, line 40:
```suggestion
  constexpr OutType max_output_value = std::numeric_limits< OutType >::max();
```

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+  if( d->percentile_norm >= 0.5 )
+  {
+    d->percentile_norm /= 100.;
+  }
+}
+
+// ----------------------------------------------------------------------------
+bool
+convert_image
+::check_configuration( vital::config_block_sptr config ) const
```
mwoehlke-kitware, line 417:
```suggestion
::check_configuration( vital::config_block_sptr VTK_UNUSED config ) const
```

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+  if( sampling_points == 0 )
+  {
+    return dst;
+  }
+
+  auto const scanning_area = src.size();
+  auto const ni = src.ni();
+  auto const nj = src.nj();
+  auto const np = src.nplanes();
+  auto const pixel_step = scanning_area / sampling_points;
```
mwoehlke-kitware, line 127:
```suggestion
  auto const pixel_step =
    static_cast< unsigned >( scanning_area / sampling_points );
```

arrows/vxl/convert_image.h
```diff
@@ -0,0 +1,56 @@
...
+  : public vital::algo::image_filter
+{
+public:
+  PLUGIN_INFO( "vxl_convert_image",
+               "Convert image between different formats or scales." )
+
+  convert_image();
+  virtual ~convert_image();
+
+  /// Get this algorithm's \link vital::config_block configuration block
```
mwoehlke-kitware, line 32: Missing periods.

arrows/vxl/convert_image.h
```diff
@@ -0,0 +1,56 @@
...
+  virtual bool check_configuration( vital::config_block_sptr config ) const;
+
+  /// Convert to the right type and optionally transform
+  virtual kwiver::vital::image_container_sptr filter(
+    kwiver::vital::image_container_sptr image_data );
+
+private:
+  class priv;
+
+  const std::unique_ptr< priv > d;
```
mwoehlke-kitware, line 47:
```suggestion
  std::unique_ptr< priv > const d;
```

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+  case T:                                                                     \
+  {                                                                           \
+    using ipix_t = vil_pixel_format_type_of< T >::component_type;             \
+    if( d->format == "disable" )                                              \
+    {                                                                         \
+      return image_data;                                                      \
+    }                                                                         \
+    vil_image_view< ipix_t > input = d->apply_transforms< ipix_t >( view );   \
+                                                                              \
+    HANDLE_OUTPUT_CASE( "bool", VIL_PIXEL_FORMAT_BOOL );                      \
```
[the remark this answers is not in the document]
mwoehlke-kitware: Another option would be to specialize the scaling functions as no-ops for `bool`.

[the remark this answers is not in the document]
russelldj: I'd say no. I think the list of initially-supported types for many of these filters is unreasonably expansive.

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,499 @@
...
+  unsigned position = 0;
+
+  for( unsigned p = 0; p < np; ++p )
+  {
+    for( unsigned s = 0;
+         s < sampling_points; ++s, position += pixel_step )
+    {
+      unsigned i = position % ni;
+      unsigned j = ( position / ni ) % nj;
+      dst[ np * s + p ] = src( i, j, p );
```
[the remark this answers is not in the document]
russelldj: I updated this. As a probably pedantic aside, it seems worrying to sample in a regular pattern such as this. I'd assume random would be better at preventing weird artifacts like aliasing.

arrows/vxl/convert_image.cxx
russelldj: Agreed, addressed

arrows/vxl/convert_image.cxx
russelldj: It looks like this is typically placed before the type. And it's `VITAL_UNUSED`.

arrows/vxl/convert_image.cxx
russelldj: Replaced by a static cast.

arrows/vxl/convert_image.cxx
mwoehlke-kitware: Oops, sorry, muscle memory fail :flushed:. (`VTK_UNUSED` is IIRC `Type VTK_UNUSED(name)` which is why I also got the placement wrong; hopefully you caught that also on the other PR? Pretty sure I at least got the right macro name in the other one...)

arrows/vxl/convert_image.cxx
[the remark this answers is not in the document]
mwoehlke-kitware: Yeah, I've had the same thought. I know @mattdawkins implemented it this way...

arrows/vxl/convert_image.cxx
mwoehlke-kitware: Aaaaaah... I see; my bad, I was thinking `view` was also `vil_image_view`; I missed it is `vil_image_view_base_sptr`. This was fine.

arrows/vxl/convert_image.cxx
russelldj: Both should be correctly placed.

mleotta:
Jenkins test this please

arrows/vxl/convert_image.cxx
```diff
@@ -74,13 +77,18 @@ combine_channels( vil_image_view< Type > const& src,
...
-random_grey_conversion( vil_image_view< Type > const& src,
-                        double const random_factor )
+random_gray_conversion( vil_image_view< Type > const& src,
+                        double const random_fraction )
 {
-  if( static_cast< double >( rand() ) / RAND_MAX < random_factor )
+  std::random_device random_device;
+  std::mt19937 engine{ random_device() };
+  std::uniform_real_distribution< double > dist( 0.0, 1.0 );
+  auto random_value = dist( engine );
```
mwoehlke-kitware, line 89: Thanks! However, you might want to consider kicking this up to be a member of the algorithm so it is not necessary to recreate the engine and distribution for every image.

mwoehlke-kitware (approved):

arrows/vxl/convert_image.cxx
mwoehlke-kitware: (Ping?)

mwoehlke-kitware:
Looks okay; there is a small style nit that got overlooked, and the random thing could be optimized, but neither is critical.

arrows/vxl/convert_image.cxx
russelldj: I left this style because uncrustify regresses to it.

arrows/vxl/convert_image.cxx
russelldj: Address, I should have refactored that into a method of the implementation class earlier.

arrows/vxl/convert_image.cxx
russelldj: I did update it in in the most recent commit. However, I'm still a bit wary of going against `uncrustify` for low-value changes. I'd hate to see someone stop running `uncrustify` because they know they're going to have to manually address a bunch of regressions like this.

mwoehlke-kitware: Sigh... that's my bad; on further consideration, I don't think `nl_template_func_decl` should be `remove`. Here:
```c++
  template < typename OutputType, typename InputType >
  void
  nonmax_suppression( vil_image_view< InputType > const& grad_i,
                      vil_image_view< InputType > const& grad_j,
                      vil_image_view< OutputType >& output_i,
                      vil_image_view< OutputType >& output_j );
```
...you must be ignoring uncrustify? (Because `nl_template_func_decl = remove` should be removing the newline there also, and seems to do so.)

mwoehlke-kitware: Nota bene: While it's true that uncrustify has bugs, sometimes the bugs are instead in the configuration :slightly_smiling_face:. (When you have a few hundred knobs to twiddle, it's bound to happen occasionally.)

russelldj: Yes, it looks like you're right about that. I agree the shown (non-uncrustify) style is better, so I guess I've been rather sporadic in which way I've gone. You'll notice the following function, `filter`, is styled as uncrustify suggests...so I'm just a mess.

arrows/vxl/convert_image.cxx
```diff
@@ -275,21 +245,62 @@ class convert_image::priv
...
+// ----------------------------------------------------------------------------
+template < typename Type >
+vil_image_view< Type >
+convert_image::priv
+::random_gray_conversion( vil_image_view< Type > const& src,
+                          double const random_fraction )
+{
+  auto random_value = dist( engine );
+
+  if( random_value < random_fraction )
```
mwoehlke-kitware, line 284: ...not necessary, but at this point you almost might as well write:
```suggestion
  if( dist( engine ) < random_fraction )
```

(On that note, however, you might also consider renaming `dist` → `random_dist`, and `engine` → `random_engine`, or maybe `rng`.)

arrows/vxl/convert_image.cxx
```diff
@@ -320,7 +331,7 @@ ::apply_transforms( vil_image_view_base_sptr& view )
 template < typename opix_t, typename ipix_t >
 vil_image_view< opix_t >
 convert_image::priv
-::scale_and_convert( vil_image_view< ipix_t > input )
+::scale_and_convert( vil_image_view< ipix_t >& input )
```
mwoehlke-kitware, line 334: Uh... did you intend to make this a mutable reference? Maybe you meant `const&`?

arrows/vxl/convert_image.cxx
```diff
@@ -0,0 +1,510 @@
...
+  }
+}
+
+// ----------------------------------------------------------------------------
+template < typename ipix_t >
+vil_image_view< ipix_t >
+convert_image::priv
+::apply_transforms( vil_image_view_base_sptr& view )
+{
+  vil_image_view< ipix_t > output;
```
mwoehlke-kitware, line 310: BTW, sorry for not noticing earlier, but should this maybe first check that the input isn't already one-channel?

If yes, it might also make sense to move `output` to only the `if( single_channel )` block and make everything else just `return`.

mleotta:
Jenkins test this please

arrows/vxl/convert_image.cxx
```diff
@@ -154,12 +122,14 @@ sample_and_sort_image( vil_image_view< PixType > const& src,
   if( remove_extremes )
   {
     constexpr auto low = PixType{ 0 };
+
```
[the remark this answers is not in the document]
russelldj: I have `nl_var_def_blk_end = 0`. This newline gets added.

mleotta:
Jenkins test this please

mleotta (approved):

PR #1191, opened 2021-01-25
Add axis-aligned edge filter ported from VIDTK

russelldj:
This is similar to other recent PRs (e.g. #1177, #1184, #1185) which port image filters from VIDTK into KWIVER. This computes the horizontal and vertical responses using a 3x3 sobel filter and nonmax suppression.

There is currently an issue with the VXL templated sobel filter only having parameterizations for a limited number of types, and I am currently looking into this.


russelldj:
> There is currently an issue with the VXL templated sobel filter only having parameterizations for a limited number of types, and I am currently looking into this.

It looks like the filter is [only instantiated](https://github.com/vxl/vxl/search?q=VIL_SOBEL_3X3_INSTANTIATE) for `byte`,  `uint16` and `float`, when producing a `float` output so I'm considering this resolved.

russelldj:
I think this is complete. The only thing I was wondering is if the `calculate_alinged_edges` and `nonmax_suppression` functions are general enough to merit porting to a separate file? I'd probably lean toward getting this in, then refactoring later.

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,307 @@
...
+  // Calculate sobel approx in x/y directions
+  vil_sobel_3x3( input, grad_i, grad_j );
+
+  // Take absolute value of gradients
+  vil_transform< GradientType, GradientType( GradientType ) >( grad_i,
+                                                               std::abs );
+  vil_transform< GradientType, GradientType( GradientType ) >( grad_j,
+                                                               std::abs );
+
+  // Perform NMS in vert/hori directions and threshold magnitude
```
dstoup, line 120: Two of these long lines below have rather arbitrary line breaks. Try breaking after the equal to see if we can make the code more readable. Breaking early should allow less line breaks.

dstoup (changes requested):
The only CI failure appears to be the test associated with this change.

arrows/vxl/aligned_edge_detection.cxx
russelldj: Thanks for the catch. This formatting was induced by `uncrustify` so if I get a chance I'll see if there's a config option to get it to break earlier. However, the current config did not regress the formatting when I manually broke at the equals.

russelldj:
> The only CI failure appears to be the test associated with this change.

I'm confused why this failed because when running `ctest` locally on Ubuntu 18, all tests, including this one, passed. Are there any gotchas for replicating the testing process locally? Also, is there any way to get a more detailed error message from Travis, such as that provided by `--output-on-failure`?

arrows/vxl/aligned_edge_detection.cxx
mwoehlke-kitware: I believe when breaks are added, they are added as late as possible. The preferred solution is to manually add breaks at more sensible locations. (In my experience, it's effectively impossible to write tools that are as clever as a human. As I've noted elsewhere, *I* view the goal of tool assisted formatting not as "make it perfect", but "accept correct formatting, and let you know if it's wrong". By contrast, clang-format might *enforce* the poor formatting in a case like this...)

dstoup:
Jenkins test this please

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+  return output;
+}
+
+template < typename pix_t >
+vil_image_view< pix_t >
+aligned_edge_detection::priv
+::filter( const vil_image_view< pix_t >& input_image )
+{
+  size_t source_ni = input_image.ni();
+  size_t source_nj = input_image.nj();
```
mwoehlke-kitware, line 132:
```suggestion
  auto const source_ni = input_image.ni();
  auto const source_nj = input_image.nj();
```

I am mildly concerned that this is `size_t` while above was `unsigned`... Appropriate use of `auto` allows using the correct type without having to explicitly name said type.

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+
+  vil_image_view< pix_t > combined_edges( source_ni, source_nj );
+
+  if( produce_joint_output )
+  {
+    combined_edges.fill( 0 );
```
mwoehlke-kitware, line 138: Minor nit: it seems like this could move down a few lines so there is one fewer `if( produce_joint_output )`? In fact, do we even need `combined_edges` outside that `if`?

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+    unsigned min_dim = std::min( joint_nms_edges.ni(), joint_nms_edges.nj() );
+
+    if( 2 * half_step + 1 >= min_dim )
+    {
+      half_step = ( min_dim - 1 ) / 2;
+    }
+
+    if( half_step != 0 )
+    {
+      vil_gauss_filter_2d( joint_nms_edges,
```
mwoehlke-kitware, line 167: This doesn't seem to need so many line breaks?

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+
+template < typename pix_t >
+vil_image_view< pix_t >
+aligned_edge_detection::priv
+::filter( const vil_image_view< pix_t >& input_image )
+{
+  size_t source_ni = input_image.ni();
+  size_t source_nj = input_image.nj();
+
+  vil_image_view< pix_t > combined_edges( source_ni, source_nj );
```
mwoehlke-kitware, line 134:
```suggestion
  vil_image_view< pix_t > combined_edges{ source_ni, source_nj };
```

...and similar for other locals.

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+    vil_plane( all_channels, 2 ) = joint_nms_edges;
+    return all_channels;
+  }
+  return aligned_edges;
+}
+
+// ----------------------------------------------------------------------------
+aligned_edge_detection
+::aligned_edge_detection()
+  : d( new priv( this ) )
```
mwoehlke-kitware, line 189:
```suggestion
  : d{ new priv{ this } }
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+                     d->produce_joint_output,
+                     "Set to false if we do not want to spend time computing "
+                     "joint edge images comprised of both horizontal and vertical "
+                     "information." );
+  config->set_value( "smoothing_sigma",
+                     d->smoothing_sigma,
+                     "Smoothing sigma for the output NMS edge density map." );
+  config->set_value( "smoothing_half_step",
+                     d->smoothing_half_step,
+                     "Smoothing half step for the output NMS edge density map." );
```
mwoehlke-kitware, line 220:
```suggestion
  config->set_value(
    "smoothing_half_step",
    d->smoothing_half_step,
    "Smoothing half step for the output NMS edge density map." );
```

...and maybe similar for others, for consistency. (Alternatively, break the help string to avoid going over length; no particular preference as to which.)

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+aligned_edge_detection
+::~aligned_edge_detection()
+{
+}
+
+vital::config_block_sptr
+aligned_edge_detection
+::get_configuration() const
+{
+  // get base config from base class
```
mwoehlke-kitware, line 203:
```suggestion
  // Get base config from base class
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+aligned_edge_detection
+::check_configuration( vital::config_block_sptr config ) const
+{
+  return true;
+}
+
+// ----------------------------------------------------------------------------
+kwiver::vital::image_container_sptr
+aligned_edge_detection
+::filter( kwiver::vital::image_container_sptr image_data )
```
mwoehlke-kitware, line 256: I wonder if this should take a `const&`?

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+{
+  // Get input image
+  vil_image_view_base_sptr source_image =
+    vxl::image_container::vital_to_vxl( image_data->get_image() );
+
+  // Perform Basic Validation
+  if( !image_data )
+  {
+    return kwiver::vital::image_container_sptr();
```
mwoehlke-kitware, line 265:
```suggestion
    return {};
```

...or possibly `return nullptr;`. (Multiple instances.)

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+  if( !source_image || source_image->nplanes() != 1 )
+  {
+    LOG_ERROR( logger(), "Input must be a grayscale image!" );
+    return kwiver::vital::image_container_sptr();
+  }
+
+#define HANDLE_CASE( T )                                          \
+  case T:                                                         \
+  {                                                               \
+    typedef vil_pixel_format_type_of< T >::component_type ipix_t; \
```
mwoehlke-kitware, line 278:
```suggestion
    using ipix_t = vil_pixel_format_type_of< T >::component_type; \
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+    auto filtered = d->filter< ipix_t >( source_image );          \
+    auto container = vxl::image_container( filtered );            \
```
mwoehlke-kitware, line 280:
```suggestion
    auto container = vxl::image_container{ filtered };            \
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+    return std::make_shared< vxl::image_container >( container ); \
+  }                                                               \
+
+
```
mwoehlke-kitware, line 284:
```suggestion

```

arrows/vxl/aligned_edge_detection.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace vxl {
+
+/**
+ * @brief Extract axis-aligned edges
+ */
```
mwoehlke-kitware, line 20: Please follow [the style guide](https://github.com/Kitware/kwiver/blob/master/doc/code-style.rst#api-documentation-doc):
```suggestion
/// Extract axis-aligned edges.
```

The rest of the doc looks mostly okay besides missing periods.

arrows/vxl/aligned_edge_detection.h
```diff
@@ -0,0 +1,55 @@
...
+
+  aligned_edge_detection();
+  virtual ~aligned_edge_detection();
+
+  /// Get this algorithm's \link vital::config_block configuration block
+  /// \endlink
+  virtual vital::config_block_sptr get_configuration() const;
+  /// Set this algorithm's properties via a config block
+  virtual void set_configuration( vital::config_block_sptr config );
+  /// Check that the algorithm's currently configuration is valid
```
mwoehlke-kitware, line 36:
```suggestion
  /// Check that the algorithm's current configuration is valid.
```

arrows/vxl/aligned_edge_detection.h
```diff
@@ -0,0 +1,55 @@
...
+  virtual bool check_configuration( vital::config_block_sptr config ) const;
+
+  /// Convert to the right type and optionally transform
+  virtual kwiver::vital::image_container_sptr filter(
+    kwiver::vital::image_container_sptr image_data );
+
+private:
+  class priv;
+
+  const std::unique_ptr< priv > d;
```
mwoehlke-kitware, line 46:
```suggestion
  std::unique_ptr< priv > const d;
```

arrows/vxl/tests/test_aligned_edge_detection.cxx
```diff
@@ -0,0 +1,94 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/**
```
mwoehlke-kitware, line 5: I wonder if we really want to document source files for tests. I seem to recall the usual reason for a `\file` is that documentation therein may not show up otherwise (although I also want to say that's only the case for e.g. macros and maybe free functions).

arrows/vxl/tests/test_aligned_edge_detection.cxx
```diff
@@ -0,0 +1,94 @@
...
+  void SetUp();
+
+  TEST_ARG( data_dir );
+
+  ka::vxl::image_io io;
+  ka::vxl::aligned_edge_detection filter;
+  kv::image_container_sptr input_image;
+};
+
+void
```
mwoehlke-kitware, line 61:
```suggestion
// ----------------------------------------------------------------------------
void
```

...and for remainder of file.

mwoehlke-kitware (changes requested):

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+#include <arrows/vxl/image_container.h>
+
+#include <vil/algo/vil_gauss_filter.h>
+#include <vil/algo/vil_sobel_3x3.h>
+#include <vil/vil_convert.h>
+#include <vil/vil_image_view.h>
+#include <vil/vil_math.h>
+#include <vil/vil_plane.h>
+
+#include <cstdlib>
```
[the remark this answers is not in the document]
mwoehlke-kitware: (Read: "consider this optional", in case that wasn't obvious...)

dstoup:
> > The only CI failure appears to be the test associated with this change.
> 
> I'm confused why this failed because when running `ctest` locally on Ubuntu 18, all tests, including this one, passed. Are there any gotchas for replicating the testing process locally? Also, is there any way to get a more detailed error message from Travis, such as that provided by `--output-on-failure`?

I don't know if there is a way to get that output in Travis. I was hoping it would finish quicker so we could look. Here is what I just pulled from the Centos log. Most of the output is junk until the bolded section at the bottom.


319: Test command: /home/kitware/jenkins/workspace/KwiverCentOSPR/build/tests/bin/test-vxl-aligned_edge_detection "--gtest_filter=aligned_edge_detection.combined" "--gtest_also_run_disabled_tests" "/home/kitware/jenkins/workspace/KwiverCentOSPR/arrows/vxl/tests/data"
319: Test timeout computed to be: 30
319: WARNING: Could not load logger factory "/home/kitware/jenkins/workspace/KwiverCentOSPR/build/lib/kwiver/plugins/logger/vital_log4cplus_logger.so" as specified in environment variable "VITAL_LOGGER_FACTORY"
319: Defaulting to built-in logger.
319: Unable to load logger factory plug-in: /home/kitware/jenkins/workspace/KwiverCentOSPR/build/lib/kwiver/plugins/logger/vital_log4cplus_logger.so: cannot open shared object file: No such file or directory
319: 2021-01-26 13:12:16.774 WARNING kwiver.vital.modules.loaders(155): Could not import: kwiver.vital.tests.alg.simple_bundle_adjust, Reason: No module named 'multimethod'
319: 2021-01-26 13:12:16.791 WARNING kwiver.vital.modules.loaders(155): Could not import: kwiver.vital.tests.alg.simple_compute_track_descriptors, Reason: cannot import name 'ComputeTrackDescriptors'
319: 2021-01-26 13:12:16.798 WARNING kwiver.vital.modules.module_loader(53): Python module "<module 'kwiver.sprokit.processes.kwiver_process' from '/home/kitware/jenkins/workspace/KwiverCentOSPR/build/lib/python3.6/site-packages/kwiver/sprokit/processes/kwiver_process.py'>" does not have registrar method
319: Note: Google Test filter = aligned_edge_detection.combined
319: [==========] Running 1 test from 1 test case.
319: [----------] Global test environment set-up.
319: [----------] 1 test from aligned_edge_detection
319: [ RUN      ] aligned_edge_detection.combined
**319: /home/kitware/jenkins/workspace/KwiverCentOSPR/arrows/vxl/tests/test_aligned_edge_detection.cxx:93: Failure
319: Value of: equal_content( filtered_image_ptr->get_image(), expected_image_ptr->get_image() )
319:   Actual: false
319: Expected: true**

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+                           vil_image_view< GradientType >& grad_j )
+{
+  // Calculate sobel approx in x/y directions
+  vil_sobel_3x3( input, grad_i, grad_j );
+
+  // Take absolute value of gradients
+  vil_transform< GradientType, GradientType( GradientType ) >( grad_i,
+                                                               std::abs );
+  vil_transform< GradientType, GradientType( GradientType ) >( grad_j,
+                                                               std::abs );
```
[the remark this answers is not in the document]
russelldj: This fails to compile with `g++ 7.5.0` which is shipped with ubuntu 18.

arrows/vxl/aligned_edge_detection.cxx
russelldj: That seems reasonable, but this change would hit a bunch of filters since this method is overriden from 'vital/algo/image_filter`.

arrows/vxl/aligned_edge_detection.cxx
russelldj: This does compile under `g++ 7.5.0`. Is there a `c++` year flag that needs to be included for proper compilation of this?

arrows/vxl/tests/test_aligned_edge_detection.cxx
russelldj: I agree this doesn't seem to add much.

arrows/vxl/aligned_edge_detection.cxx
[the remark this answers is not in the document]
russelldj: This isn't an initialization it's a function call.

arrows/vxl/aligned_edge_detection.cxx
russelldj: I'm open to suggestions between the two. I'd hope a clear error message mitigates any confusion, so I'm not sure it matters much.

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+    }
+    else
+    {
+      vil_copy_reformat( joint_nms_edges, combined_edges );
+    }
+    vil_image_view< pix_t > all_channels( joint_nms_edges.ni(),
+                                          joint_nms_edges.nj(), 3 );
+    vil_plane( all_channels, 0 ).deep_copy( vil_plane( aligned_edges, 0 ) );
+    vil_plane( all_channels, 1 ).deep_copy( vil_plane( aligned_edges, 1 ) );
+    vil_plane( all_channels, 2 ).deep_copy( joint_nms_edges );
```
russelldj, line 178: The source of the prior bug was I was using assignment (shallow copy) rather than deep copy. However, I'm not thrilled about this current approach since it seems to waste memory. I was curious if there's any better way to deal with this?

russelldj: More generally, is there a cheap channel-wise concatenation operation in `vil` that can be used in similar situations?

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  const unsigned ni = grad_i.ni();
+  const unsigned nj = grad_i.nj();
+
+  vil_image_view< OutputType > output( ni, nj, 2 );
+  output.fill( 0 );
+
+  // Perform non-maximum suppression
+  for( unsigned j = 1; j < nj - 1; j++ )
```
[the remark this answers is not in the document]
mwoehlke-kitware: Ah... well, there's a much easier fix for that: `decltype(+nj)`. Note the `+` :slightly_smiling_face:.

arrows/vxl/aligned_edge_detection.cxx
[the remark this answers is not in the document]
mwoehlke-kitware: Ah, never mind then

arrows/vxl/aligned_edge_detection.cxx
mwoehlke-kitware: Does, or does not? "Uniform" initialization is a C++11 feature. It's unclear if `image_container` is a type; it _looks_ like a type, but even in this PR there are places that look like types that are actually functions.

arrows/vxl/aligned_edge_detection.cxx
mwoehlke-kitware: There is no effective difference; all return a null shared pointer (and at least at -O2 or higher should produce the same assembly). It's just a question of whether it's worthwhile to spell out the full type and whether `nullptr` is clearer than a default-constructed `shared_ptr`.

arrows/vxl/aligned_edge_detection.cxx
russelldj: This actually works fine, I'm not sure why I made this mistake. Sorry for the confusion.

arrows/vxl/aligned_edge_detection.cxx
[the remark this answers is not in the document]
russelldj: Addressed, thanks.

arrows/vxl/aligned_edge_detection.cxx
russelldj: I changed everything to `nullptr`.

arrows/vxl/aligned_edge_detection.cxx
mwoehlke-kitware: I don't think so? If I am understanding this correctly, you are combining planes from separate images into one image. There is not a way to do that that is generally usable (maybe in very specific instances sticking entirely to VXL, but not generally) that doesn't involve a deep copy, because the resulting image needs to have all the data in a single memory block, and the input data is in separate memory blocks. Plus, even if you could, the resulting image would need to refcount the inputs.

What you *might* be able to do, but would at least require some refactoring, would be to allocate the result image up front and do the other operations in place on "slices" of that image.

BTW, what happens to `combined_edges`? Is it me, or does it just get discarded?

arrows/vxl/aligned_edge_detection.cxx
mwoehlke-kitware: Ah, well... that's just unfortunate :slightly_smiling_face:. I wonder if it would be worthwhile to change that API... should be done as a separate PR, though, if so.

arrows/vxl/aligned_edge_detection.cxx
russelldj: The commit seems to have been messed up by the force push. The following code sets it by reference.

`
if( half_step != 0 )
{
    vil_gauss_filter_2d( joint_nms_edges, combined_edges, smoothing_sigma,
                         half_step );
}
else
{
    vil_copy_reformat( joint_nms_edges, combined_edges );
}
`

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,315 @@
...
+  // Internal parameters/settings
+  float threshold = 10;
+  bool produce_joint_output = true;
+  double smoothing_sigma = 1.3;
+  unsigned smoothing_half_step = 2;
+};
+
+// ----------------------------------------------------------------------------
+// Perform NMS on the input gradient images in horizontal and vert directions
+// only
```
mwoehlke-kitware, line 58:
```suggestion
// Perform NMS on the input gradient images in horizontal and vertical
// directions only
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,315 @@
...
+  auto const ni = grad_i.ni();
+  auto const nj = grad_i.nj();
+
+  vil_image_view< OutputType > output( ni, nj, 2 );
+  output.fill( 0 );
+
+  // Perform non-maximum suppression
+  for( decltype(+nj) j = 1; j < nj - 1; j++ )
+  {
+    for( decltype(+ni) i = 1; i < ni - 1; i++ )
```
mwoehlke-kitware, line 81:
```suggestion
  for( decltype(+nj) j = 1; j < nj - 1; ++j )
  {
    for( decltype(+ni) i = 1; i < ni - 1; ++i )
```

(Aside: I usually try to write these using `range::iota`, but there isn't a version that starts at non-zero.)

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,315 @@
...
+                           vil_image_view< GradientType >& grad_j )
+{
+  // Calculate sobel approx in x/y directions
+  vil_sobel_3x3( input, grad_i, grad_j );
+
+  // Take absolute value of gradients
+  vil_transform< GradientType, GradientType( GradientType ) >(
+    grad_i, std::abs );
+  vil_transform< GradientType, GradientType( GradientType ) >(
+    grad_j, std::abs );
```
mwoehlke-kitware, line 121: Okay... mostly because trying to read this is making me go "WTH?"...
```suggestion
  vil_transform( grad_i, std::abs< GradientType > );
  vil_transform( grad_j, std::abs< GradientType > );
```

Alternatively:
```suggestion
  vil_transform< GradientType, GradientType (*)( GradientType ) >(
    grad_i, std::abs );
  vil_transform< GradientType, GradientType (*)( GradientType ) >(
    grad_j, std::abs );
```
...but the former is more concise. (The problem is the second template argument looks like a nonsense object construction rather than a function type.)

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,315 @@
...
+
+  // Perform NMS in vertical/horizonal directions and threshold magnitude
+  vil_image_view< PixType > output =
+    nonmax_suppression< GradientType, PixType >( grad_i, grad_j );
+  return output;
```
mwoehlke-kitware, line 126: **Optional**:
```suggestion
  return nonmax_suppression< GradientType, PixType >( grad_i, grad_j );
```

If you're feeling frisky, swap the template parameters of that function and:
```suggestion
  return nonmax_suppression< PixType >( grad_i, grad_j );
```

arrows/vxl/aligned_edge_detection.cxx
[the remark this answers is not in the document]
mwoehlke-kitware: ...Also, my bad on the original suggestion, I usually *actually* write it like `auto j = decltype(nj){ 1 }` which doesn't have the issue :flushed:. (Though it's shorter in its current form, so meh...)

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,315 @@
...
+  {
+    LOG_ERROR(
+      p->logger(),
+      "Input gradient image dimensions must be equivalent" );
+  }
+
+  auto const ni = grad_i.ni();
+  auto const nj = grad_i.nj();
+
+  vil_image_view< OutputType > output( ni, nj, 2 );
```
mwoehlke-kitware, line 75:
```suggestion
  vil_image_view< OutputType > output{ ni, nj, 2 };
```

Heh, both of us missed one... (I've actually read past this *several* times without noticing...)

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,315 @@
...
+  return config;
+}
+
+// ----------------------------------------------------------------------------
+void
+aligned_edge_detection
+::set_configuration( vital::config_block_sptr in_config )
+{
+  // Starting with our generated vital::config_block to ensure that assumed
+  // values are present. An alternative is to check for key presence before
```
mwoehlke-kitware, line 232: Because my inner grammarian is offended :wink::
```suggestion
  // Start with our generated vital::config_block to ensure that assumed values
  // are present. An alternative would be to check for key presence before
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+#define HANDLE_CASE( T )                                          \
+  case T:                                                         \
+  {                                                               \
+    using ipix_t = vil_pixel_format_type_of< T >::component_type; \
+    auto filtered = d->filter< ipix_t >( source_image );          \
+    auto container = vxl::image_container{ filtered };            \
+    return std::make_shared< vxl::image_container >( container ); \
+  }                                                               \
```
mwoehlke-kitware, line 294: You should be able to avoid a copy here:
```suggestion
    return std::make_shared< vxl::image_container >( filtered );  \
  }
```

(I also recommend omitting the `\` from the last line of a `#define`.)

arrows/vxl/aligned_edge_detection.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <vital/algo/image_filter.h>
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace vxl {
+
+
```
mwoehlke-kitware, line 18: Please avoid consecutive blank lines.
```suggestion

```

arrows/vxl/aligned_edge_detection.cxx
russelldj: I did the latter. I think I got carried away saving intermediate results for debugging...

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  // Perform Basic Validation
+  if( source_image->nplanes() == 3 )
+  {
+    source_image = vil_convert_to_grey_using_average( source_image );
+  }
+  else if( source_image->nplanes() != 1 )
+  {
+    LOG_ERROR( logger(), "Input must be a grayscale image!" );
```
mleotta, line 279: Better error message might be `"Input must have either 1 or 3 channels but has " << source_image->nplanes()`

arrows/vxl/aligned_edge_detection.cxx
mleotta: This is probably not important to do unless it proves to be a bottleneck, but you can avoid the deep copy with refactoring as @mwoehlke-kitware suggested.  This is why many of the `vil` functions are written to return the output image in a parameter passed in by reference.  This allows the caller to decide which memory the results are placed in.

arrows/vxl/tests/test_aligned_edge_detection.cxx
```diff
@@ -0,0 +1,92 @@
...
+  auto const expected_filename = data_dir + "/" + expected_seperate_edges;
+
+  auto config = kv::config_block::empty_config();
+  config->set_value( "produce_joint_output", false );
+  filter.set_configuration( config );
+
+  auto const filtered_image_ptr = filter.filter( input_image );
+  auto const expected_image_ptr = io.load( expected_filename );
+
+  EXPECT_TRUE( equal_content( filtered_image_ptr->get_image(),
```
mleotta, line 78: Why is the "separate" image mostly white but the "combined" image is most black?  I thought these were supposed to be the same output but the combined adds one additional channel the combines the first two.

mleotta (commented):
Can you explain why the two test output images look so different?

arrows/vxl/tests/test_aligned_edge_detection.cxx
russelldj: I'm looking into it, but I'm fairly certain both are correct that the two channel "separate" image is simply confusing your image viewer. I believe it's being treated as a grey+alpha.

russelldj: The first two planes of the images are the same between the two outputs. So I believe this is working correctly and it's simply a visualization error.

arrows/vxl/aligned_edge_detection.cxx
russelldj: Despite Matt W. pointing it out, I still made a mistake and the images were never smoothed. In updating this to be more efficient, I fixed that issue.

arrows/vxl/tests/test_aligned_edge_detection.cxx
mleotta: makes sense to me.

mleotta (approved):

mleotta:
Jenkins test this please

mleotta (changes requested):

mleotta:
Jenkins test this please

mleotta (approved):

mleotta:
I'm ready to merge this unless any of the other reviewers want to revisit this.  @mwoehlke-kitware?  The CI builds are clean with the exception of the Linux build that is perpetually broken.

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,320 @@
...
+  // Perform NMS on the input gradient images in horizontal and vertical
+  // directions only
+  template < typename OutputType, typename InputType >
+  void
+  nonmax_suppression( vil_image_view< InputType > const& grad_i,
+                      vil_image_view< InputType > const& grad_j,
+                      vil_image_view< OutputType >& output_i,
+                      vil_image_view< OutputType >& output_j );
+  // Compute axis-aligned edges and perform NMS on them
+  template < typename pix_t > vil_image_view< pix_t >
```
mwoehlke-kitware, line 49: The `template` spec should probably not be on the same line as the return type. I guess the return type should be on its own line to match the above, but on the same line as the rest of the prototype is also okay and still within length; feel free to pick whichever you prefer.

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,320 @@
...
+  output_j.fill( 0 );
+
+  // Perform non-maximum suppression
+  for( unsigned j = 1; j < nj - 1; ++j )
+
+  {
+    for( unsigned i = 1; i < ni - 1; ++i )
+    {
+      const InputType val_i = grad_i( i, j );
+      const InputType val_j = grad_j( i, j );
```
mwoehlke-kitware, line 89:
```suggestion
      InputType const val_i = grad_i( i, j );
      InputType const val_j = grad_j( i, j );
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,320 @@
...
+  {
+    auto combined_response = vil_plane( aligned_edges, 2 );
+
+    // Add vertical and horizontal edge planes together and smooth
+    vil_math_image_sum( i_response,
+                        j_response,
+                        combined_response );
+
+    unsigned half_step = smoothing_half_step;
+    unsigned min_dim = std::min( source_ni, source_nj );
```
mwoehlke-kitware, line 170: (Weak suggestion, mostly for the `const`...)
```suggestion
    auto const half_step = smoothing_half_step;
    auto const min_dim = std::min( source_ni, source_nj );
```

arrows/vxl/aligned_edge_detection.cxx
```diff
@@ -0,0 +1,320 @@
...
+    config->get_value< bool >( "produce_joint_output" );
+  d->smoothing_sigma = config->get_value< double >( "smoothing_sigma" );
+  d->smoothing_half_step =
+    config->get_value< unsigned >( "smoothing_half_step" );
+}
+
+// ----------------------------------------------------------------------------
+bool
+aligned_edge_detection
+::check_configuration( vital::config_block_sptr config ) const
```
mwoehlke-kitware, line 254:
```suggestion
::check_configuration( vital::config_block_sptr VITAL_UNUSED config ) const
```

mwoehlke-kitware (commented):
LGTM aside from a very small number :slightly_smiling_face: of minor style nits and one missing warning suppression.

mleotta:
The recent changes broke the CI

russelldj:
> The recent changes broke the CI

Ack, sorry I missed that. Addressed.

arrows/vxl/aligned_edge_detection.cxx
mwoehlke-kitware: (Ping?)

mleotta:
Jenkins test this please

PR #1192, opened 2021-01-27
Dev/applet command tree

linus-sherrill:
This branch adds hierarchical commands to the kwiver applet runner. The command name is still taken from the applet plugin name, but the string is then processed to extract the command levels. The command levels are separated by a colon (':'). The "kwiver run pipe" command would be encoded as "run:pipe". It is also possible to have the same applet appear at multiple points in the command tree. The alternate command locations are separated by a semi-colon (';').

### DIFF
--- a/tools/kwiver_tool_runner.cxx
+++ b/tools/kwiver_tool_runner.cxx
@@ -19,6 +19,7 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <set>
 #include <utility>
 
 using applet_context_t = std::shared_ptr< kwiver::tools::applet_context >;
@@ -33,8 +34,7 @@ class command_node
 {
 public:
   command_node(std::string const& name)
-    : m_name( name ),
-      m_is_command( false )
+    : m_name( name )
   { }
 
   virtual ~command_node() = default;
@@ -43,11 +43,10 @@ public:
   add_factory( kv::plugin_factory_handle_t fact )
   {
     m_factory = fact;
-    m_is_command = true;
   }
 
   bool
-  is_command() const { return m_is_command; }
+  is_command() const { return ( m_factory != nullptr ); }
 
   command_node_sptr
   find( std::string const& name )
@@ -65,10 +64,7 @@ public:
   void
   add( command_node_sptr n )
   {
-    m_nodes.push_back(n);
-
-    // sort vector
-    std::sort( m_nodes.begin(), m_nodes.end(), compare_node );
+    m_nodes.insert(n);
     return;
   }
 
@@ -87,6 +83,7 @@ public:
     }
   }
 
+  // This is just used for debugging tree construction.
   void
   dump_tree(std::string const& pfx = "") const
   {
@@ -97,17 +94,18 @@ public:
     }
   }
 
+  struct compare_node
+  {
+    bool operator()( command_node_sptr const& n1, command_node_sptr const& n2 )
+    {
+      return (n1->m_name < n2->m_name);
+    }
+  };
+
   std::string m_name;
-  bool m_is_command { false };
-  std::vector<command_node_sptr> m_nodes;
+  std::set< command_node_sptr, compare_node > m_nodes;
   kv::plugin_factory_handle_t m_factory;
 };
-
-// --------------------------------------------------------------------
-bool compare_node( command_node_sptr const& n1, command_node_sptr const& n2 )
-{
-  return (n1->m_name < n2->m_name);
-}
 
 // ============================================================================
 /**
@@ -171,8 +169,6 @@ public:
 
   // ----------------------
   // tool runner arguments.
-  std::string m_output_file; // empty is no file specified
-
   std::vector<std::string> m_runner_args;
   std::vector<std::string> m_applet_args;
 
@@ -232,8 +228,7 @@ build_command_tree( kwiver::vital::plugin_manager& vpm )
 
           }
           std::cerr << "\" already exists. Defined by ";
-          auto nn = current_node->find(c);
-          auto f = nn->m_factory;
+          auto f = n->m_factory;
           std::string plugin_file;
           f->get_attribute( kv::plugin_factory::PLUGIN_FILE_NAME, plugin_file );
           std::cerr << plugin_file << std::endl;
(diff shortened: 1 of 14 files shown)


linus-sherrill:
Jenkins, test this please

linus-sherrill:
Jenkins test this please

linus-sherrill:
After we decide on the individual commands, I will update the places where it needs to be changed in the examples, etc.

tools/kwiver_tool_runner.cxx
```diff
@@ -22,8 +21,93 @@
...
+    } // end for
+    return nullptr;
+  }
+
+  void
+  add( command_node_sptr n )
+  {
+    m_nodes.push_back(n);
+
+    // sort vector
```
mleotta, line 70: this comment is unnecessary as is, but you could explain why this needs to be sorted.  Alternatively you could just use a data structure like `std::set` that keeps them sorted.

tools/kwiver_tool_runner.cxx
```diff
@@ -22,8 +21,93 @@
...
+
+    if( this->is_command() )
+    {
+      std::cout << pfx << " " << this->m_name << " " << std::endl;
+      return;
+    }
+  }
+
+  void
+  dump_tree(std::string const& pfx = "") const
```
mleotta, line 91: is this just for debugging?  It doesn't look to be used and is similar to `print_subtree_help`

tools/kwiver_tool_runner.cxx
```diff
@@ -134,43 +313,11 @@ void tool_runner_usage( VITAL_UNUSED applet_context_t ctxt,
...
-  applet->add_command_options();
-
-  // display help text
-  std::cout << applet->m_cmd_options->help();
-}
-
 // ============================================================================
 int main(int argc, char *argv[])
 {
+command_node_sptr cmd_root;
```
mleotta, line 319: indentation looks odd here.  Also, could you just declare this later when it is initialized?

tools/kwiver_tool_runner.cxx
```diff
@@ -186,21 +333,59 @@ int main(int argc, char *argv[])
...
+      if ( ! n )
+      {
+        break;
+      }
+      cn = n;
+    } // end for
+
+    // If there are no matches or we run out of command words
+    // before we reach a terminal command node.
+    if ( ci == 0 && !cn->is_command() )
```
mleotta, line 364: Is there ever a case where `ci == 0` and `cn` is a command?  Doesn't `ci == 0` imply that `cn == cmd_root`?  I suppose it can't hurt to check.

tools/kwiver_tool_runner.cxx
```diff
@@ -186,21 +333,59 @@ int main(int argc, char *argv[])
...
+      std::cerr << "Command not found. Related commands are as follows:\n";
+      cn->print_subtree_help();
+      exit(-1);
+    }
+
+    // Insert name words[ci+1] [size-1] at start of options.m_applet_args
+    for( size_t ai = ci; ai < options.m_applet_name.size(); ai++)
+    {
+      options.m_applet_args.insert(
+        options.m_applet_args.begin()+1, // skip first "kwiver" element
```
mleotta, line 383: What is this first "kwiver" element?  Does this mean the arguments to the function are {"kwiver", "arg1", "arg2",...}?  Is "kwiver" in this list but the name of the applet not in this list?

mleotta (changes requested):
This looks pretty good, but I have some minor feedback/questions.

vital/applets/applet_context.h
```diff
@@ -28,8 +28,8 @@ class applet_context
   // Used to wrap large text blocks
   kwiver::vital::wrap_text_block m_wtb;
 
-  // name of the applet. as in kwiver <applet> <args..>
-  std::string m_applet_name;
+  // name of the applet. as in kwiver <ap pl et> <args..>
```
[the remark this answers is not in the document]
linus-sherrill: Done

tools/kwiver_tool_runner.cxx
```diff
@@ -22,8 +21,93 @@
...
+using command_node_sptr = std::shared_ptr<command_node>;
+bool compare_node( command_node_sptr const& n1, command_node_sptr const& n2 );
+
+// -------------------------------------------------------------------
+class command_node
+{
+public:
+  command_node(std::string const& name)
+    : m_name( name ),
+      m_is_command( false )
```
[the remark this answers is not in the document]
linus-sherrill: Done

tools/kwiver_tool_runner.cxx
linus-sherrill: Done

tools/kwiver_tool_runner.cxx
linus-sherrill: It is just used for debugging the tree construction.

tools/kwiver_tool_runner.cxx
linus-sherrill: Done

tools/kwiver_tool_runner.cxx
linus-sherrill: Commands can be present at the root if it is a one word command. This is the case when supporting legacy plugins.

mleotta: Understood, but isn't the root always the node named "Root" that contains everything else and is not a command?  So those top-level commands would be at `ci == 1`, right?

linus-sherrill: This can happen if the command line is malformed and there are no elements in the `m_applet_name` vector.

tools/kwiver_tool_runner.cxx
```diff
@@ -77,9 +175,82 @@ class command_line_parser
...
+            std::cerr << ew;
+            if ( ew == c )
+            {
+              break;
+            }
+            std::cerr << " ";
+
+          }
+          std::cerr << "\" already exists. Defined by ";
+          auto nn = current_node->find(c);
```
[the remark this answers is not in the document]
linus-sherrill: Done

tools/kwiver_tool_runner.cxx
linus-sherrill: The "kwiver" element is inserted to comply with standard argv semantics (the name of "program") so that command line parsing has something there. If I put an applet name there which has a multi-word name, the command line parser will break each word int a separate argument. The applet name is available to the applet via the applet_name() call.

mleotta (approved):
I haven't tried running it, but the code changes look good to me.

linus-sherrill:
I don't think this change should go into the minot release since it is a major change to the commands. Other opinions welcome. That being said, there was some discussion on the grammar of the commands (noun-verb vs verb-noun) that should be discussed. Once the new naming convention is established, all the other references to the commands need to be modified (e.g in examples).

PR #1198, opened 2021-02-10
Super process for Burnout local pixel-wise filters

russelldj:
This depends on multiple image filters (#1177, #1184, #1185, #1186, #1191) which are still in review. It simply serves as a convenient way to run all filters on the same image.


russelldj:
All the above branches have been merged, so this could be added now. As Leotta, Dawkins, and Woehlke have all mentioned, this method should almost certainly be replaced with direct use of a pipeline to run filters individually. But for the time being, I think it's good enough and should be merged, despite likely having some code quality issues.

arrows/vxl/pixel_feature_extractor.cxx
```diff
@@ -0,0 +1,321 @@
...
+  {
+    auto high_pass_bidir = high_pass_bidir_filter.filter( input_image );
+    // 2 channels
+    filtered_images.push_back(
+      vxl::image_container::vital_to_vxl( high_pass_bidir->get_image() ) );
+  }
+  if( enable_convert )
+  {
+    // 3 channels
+    // This should be removed
```
mleotta, line 155: If it should be removed, why is it here for now?

arrows/vxl/pixel_feature_extractor.cxx
```diff
@@ -0,0 +1,321 @@
...
+    auto converted = convert_image_filter.filter( input_image );
+    filtered_images.push_back(
+        vxl::image_container::vital_to_vxl( converted->get_image() ) );
+  }
+  // TODO this should be the variance instead
```
mleotta, line 160: You mean this should include the variance in the future, but does not now?

arrows/vxl/pixel_feature_extractor.cxx
```diff
@@ -0,0 +1,321 @@
...
+  if( enable_average )
+  {
+    // This should
```
mleotta, line 163: This should?

arrows/vxl/pixel_feature_extractor.cxx
```diff
@@ -0,0 +1,321 @@
...
+
+// ----------------------------------------------------------------------------
+template < typename pix_t >
+vil_image_view< pix_t >
+pixel_feature_extractor::priv
+::filter( kwiver::vital::image_container_sptr input_image )
+{
+  std::vector< vil_image_view< vxl_byte > > filtered_images;
+
+  if( enable_color || enable_gray )
```
mleotta, line 114: I actually don't think there is much value in building a sprokit pipeline that runs each of these filters in separate pipeline processes.  These are all independent filters that run in parallel, so it would probably be just as efficient, or more efficient, to make this function multi-threaded (later, not now) to run this processing in parallel.   The only advantage of the sprokit pipeline would be better run-time configurability if you wanted to reorder these or add a new one without recompiling.

mleotta (approved):
This looks good to me, but it might help to make the TODO comments more clear.

If it would help to merge this as is you can address that on a new branch.

arrows/vxl/pixel_feature_extractor.cxx
russelldj: For some reason the `convert_color` filter was in the original big PR even though it's not used in the default configs. I just removed it, but I could see the argument for leaving it in, in case we might use it in the future.

arrows/vxl/pixel_feature_extractor.cxx
russelldj: The updated comment is more clear. I was just considering renaming these options to `variance` since the variance output is used primarily or exclusively, rather than the average.

arrows/vxl/pixel_feature_extractor.cxx
russelldj: This makes sense to me.

mleotta:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

russelldj:
From my read of the timestamp, it looks like the failing build was triggered a few minutes before the fix in #1208 landed. Am I correct that Jenkins triggers a merge before testing? Or do I need to rebase this branch onto master to pull in that fix?

russelldj:
Jenkins test this please

dstoup:
> ting? Or do I need to rebase this

I always assumed that it did merge behind the scenes but I have seen cases where is clearly did not. I believe it just gets the pull id from github. Best to pull in master and repush. Rebase is ok but it will change the hash, in case that matters.

mleotta:
Jenkins test this please

arrows/vxl/pixel_feature_extractor.cxx
```diff
@@ -0,0 +1,301 @@
...
+    config->get_value< bool >( "enable_high_pass_bidir" );
+
+  if( !( enable_color || enable_gray || enable_aligned_edge ||
+         enable_average || enable_color_commonality || enable_high_pass_box ||
+         enable_high_pass_bidir) )
+  {
+    LOG_ERROR( logger(), "At least one filter must be enabled" );
+    return false;
+  }
+  return true;
```
mleotta, line 275: @russelldj I'm going to merge this as is to keep things moving, but the other thing I just noticed that was missing is that you should also call `check_configuration` on all of the sub algorithm that are enabled and return the `&&` of those.  You can do this on a new branch.

PR #1204, opened 2021-02-15
Switch to modern CMake CUDA support

mleotta:


dstoup:
LGTM but I'm not really a pro with CUDA so can't comment on a deep level

PR #1213, opened 2021-02-19
Execute `check_configuration` on pixel_feature_extractor sub-algorithms

russelldj:
There's quite a bit of noise from formatting changes and updating sub-algorithms to pointers, but the only functional change should be ensuring that all sub-algorithms pass their config checks.


mleotta (approved):

mleotta:
Jenkins test this please

PR #1215, opened 2021-02-24
Update versioning in relevant files for upcoming patch release v1.5.1

johnwparent:


mleotta (approved):

russelldj:
The Windows CI appears to be hung here as well. Can someone rerun Jenkins?

mleotta:
This change is still good, but I want to hold the release up until we can address one more critical issue with the last release that was recently reported by a customer.  There is an initial fix in #1217, but it still needs to be rebased to release, and this fix is more of a workaround.  At a minimum I want to land a workaround for the issue as in #1217, but it might be better to understand the problem correct the underlying issue.

johnwparent:
Alright, the releases for Kwiver and Fletch will have no problem holding off for a while longer, this also gives us time to get the outstanding Fletch branch marked for patch release in. Given that there is a minor release coming on the heels of the patch, depending on timing, we may want to split the workaround and the underlying fix between the two releases.

mleotta:
#1219 has landed (which supersedes #1217 and #1218) to address the bug I was concerted about.  If we resolve the conflicts we can merge this branch now and make the patch release.

mleotta (approved):

PR #1217, opened 2021-02-26
dev/eb

borovik135:
address core dump crash on new project, compute/track features under ubuntu 20.04


mleotta:
@borovik135 thank you for tracking this issue down.  Looking at this fix we need to understand if there is a bigger underlying problem here.  This crash is the result of descriptors switching over to a new `vital::set` container about a year ago.  This set contains a `std::vector`.  You are showing that extracting the underlying `std::vector` and iterating over that works, but iterating over the `vital::set` container leads to a crash.  I'm also seeing this crash on older versions of Ubuntu, but not Windows.  It would help to understand the cause of this crash better.  The better solution might be to fix the `vital::set` iterators if they are broken. 

Do you have any insight into why the code was crashing?  E.g. which line number or function call it was crashing on?

mleotta:
Whether we use this fix or a different one, here is some guidance on our git workflow.

For small fixes like this it should probably be just a single commit.  It will keep the history cleaner if we cut out all those extra merge commits.  Also in this case, this is a bug fix and the bug was introduced before the last release of KWIVER, so we would ideally apply this fix to the `release` branch to include in the next patch release.  The `release` branch is also merged into `master` so this change will land in master as well.  So what I would do is make a new branch off of the `release` branch and cherry-pick this commit (1c1a6e3) and then push that branch.  When you make the pull request, select `release` as the merge target instead of `master`.

For bug fixes targeting `release`, it's also a good idea to make a brief comment in the `doc/release-notes/release.txt` describing the bug that was fixed. 

Lastly, the preferred naming for development branches is to use something more descriptive about the change itself.  So something like `dev/fix-track-feature-crash` would be more appropriate here.

mleotta (changes requested):
This is a fairly important fix for the imminent patch release.  At a minimum we need to retarget this fix for inclusion in the release branch.

This workaround is probably sufficient if we want to get the patch release out soon, but it seems like there is a deeper issue here that we need to understand.

borovik135:
> ... which line number or function call it was crashing on?

In the current Kwiver master, the app is crashing during [descriptor->clone](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/vital/types/feature_track_set.h#L77)() due to the underlying descriptor object being incorrectly initialized (only showing a reference  to VMT) and nothing else, i.e. no members
==
descriptor	kwiver::vital::descriptor_sptr	{...}	
> get()	std::__shared_ptr<kwiver::vital::descriptor, (__gnu_cxx::_Lock_policy)2>::element_type *	0x7fff64005f30	
>> _vptr.descriptor	int (**)(void)	0x7000000030000	
==

The tail of the log showed that there was an attempt to compute and save descriptor(s) for the first frame
==
...
INFO - Reduced 4711 features to 1038 features with non-max radius 5.33943
DEBUG - Computing new descriptors on frame 1
DEBUG - Saved features on frame 1 to results/features/09172008flight1tape3_2-00001.kwfd
==

The [code computing new descriptors](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/arrows/core/track_features_core.cxx#L310) is found track_features_core.cxx, but curr_desc was computed OK, producing a proper cv:Mat-like object with descriptors:
==
curr_desc	kwiver::vital::descriptor_set_sptr	{...}	
> get()	kwiver::arrows::ocv::descriptor_set *	0x7fffb8f88090	
>> kwiver::vital::descriptor_set	kwiver::vital::descriptor_set	{...}	
>> data_	cv::Mat	{...}	
==

Problems surfaced in the [special case for the first frame](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/arrows/core/track_features_core.cxx#L351) with fts->descriptor showing incorrectly initialized descriptor after the assignment:
==
fts->descriptor	kwiver::vital::descriptor_sptr	{...}	
> get()	std::__shared_ptr<kwiver::vital::descriptor, (__gnu_cxx::_Lock_policy)2>::element_type *	0x7fff64005f30	
>> _vptr.descriptor	int (**)(void)	0x7000000030000	
==
which looks very similar to our first descriptor trace causing the core [dump crash](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/vital/types/feature_track_set.h#L77).

The source of the problem was thus *dit:
==
dit	kwiver::arrows::core::track_features_core::desc_itr	{...}	
> kwiver::vital::base_iterator<std::shared_ptr<kwiver::vital::descriptor>, std::shared_ptr<kwiver::vital::descriptor> const>	kwiver::vital::base_iterator<std::shared_ptr<kwiver::vital::descriptor>, std::shared_ptr<kwiver::vital::descriptor> const>	{...}	
>> m_cur_val_ptr	kwiver::vital::base_iterator<std::shared_ptr<kwiver::vital::descriptor>, std::shared_ptr<kwiver::vital::descriptor> const>::pointer	0x7fffb800b158	
>>> std::__shared_ptr_access<kwiver::vital::descriptor, (__gnu_cxx::_Lock_policy)2, false, false>	std::__shared_ptr_access<kwiver::vital::descriptor, (__gnu_cxx::_Lock_policy)2, false, false>	{...}	
>>>> _M_ptr	std::__shared_ptr<kwiver::vital::descriptor, (__gnu_cxx::_Lock_policy)2>::element_type *	0x7fffb8003860	
>>>>> _vptr.descriptor	int (**)(void)	0x7000000030000
==
which was [initialized](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/arrows/core/track_features_core.cxx#L345) as desc_itr dit = curr_desc->begin();

Notice that our feature iterator is [initialized differently](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/arrows/core/track_features_core.cxx#L344): feat_itr fit = vf.begin(); using [std::vector of features](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/arrows/core/track_features_core.cxx#L335) as the container.

The introduced change simply made iterator initialization for both features and descriptors analogous. It is only done for the first frame, so the overall overhead may be negligible.

mleotta:
@borovik135 I think we should rebase or cherry-pick your change onto the release branch (as described above) and get that merged to address the immediate problem quickly.  However, it seems that there is something deeper wrong with `vital::iterator` or `vital::iterable` or `vital::set` and maybe @linus-sherrill or @Purg who have had more involvement with those features of kwiver can help us track that down so that it doesn't cause problems elsewhere.

mleotta:
I've done a little more digging and I have a hunch.  I think this is a bug with the iterable functions in the ocv arrow descriptor set.  This is a special case in which the iterator is not returning a reference to data contained in the set, but it converting the data on the fly into a different format and returning that.  I'm thinking we are converting to a temporary object to that we return are reference to and then it goes out of scope.  See

https://github.com/Kitware/kwiver/blame/master/arrows/ocv/descriptor_set.cxx#L225
https://github.com/Kitware/kwiver/blob/master/arrows/ocv/descriptor_set.cxx#L258

I'm wondering if we change `reference` to `value_type` on those line if that would resolve this issue.  We should probably also update the unit test to actually do something with the data in the loop (like clone) to confirm that it is valid.

https://github.com/Kitware/kwiver/blob/master/arrows/ocv/tests/test_descriptor_set.cxx#L101

borovik135:
> @borovik135 I think we should rebase or cherry-pick your change onto the release branch (as described above) and get that merged to address the immediate problem quickly.

@mleotta : I have attempted to check-out release, branch out, but that version of kwiver would not build on my Ubuntu-20.04 with CUDA-11 probably for the same reason the master branch would not build a couple of weeks ago, needing a few changes accomodating newer versions of OpenCV, CUDA and VTK, e.g.

/work/kwiver/src/arrows/ocv/extract_descriptors_DAISY.cxx:96:44: error: invalid conversion from ‘int’ to ‘cv::xfeatures2d::DAISY::NormalizationType’ [-fpermissive]
   96 |                                            norm, cv::noArray(), interpolation,
      |                                            ^~~~
      |                                            int

borovik135:
@mleotta per one of your suggestions, I have cherry-picked the change in (1c1a6e3), merged it into a release branch, pushed it to the repo and created a corresponding pull request, summarizing this discussion.

mleotta:
replaced by #1218

PR #1218, opened 2021-02-26
fix the crash on new project feature/descriptor extraction

borovik135:
Address core dump crash on new project, compute track features under ubuntu 20.04. Fix attained by getting a set of descriptors in a std:vector (as opposed to vital::set based) container and iterating over them for the first frame case. The vital set container for the descriptors apparently needs a closer look in case of descriptors that are stored in a cv::Mat, as current implementation attempts to simulate iteration over matrix rows, but does it in an unreliable manner, causing this crash at the attempt to [clone](https://github.com/Kitware/kwiver/blob/888502a3e30fdb74c1882b9c936ed6e2ed264f1e/vital/types/feature_track_set.h#L77) track's descriptors.

### DIFF
--- a/arrows/core/track_features_core.cxx
+++ b/arrows/core/track_features_core.cxx
@@ -365,19 +365,18 @@ track_features_core
                          << " to " << kwfd_file );
   }
 
-  std::vector<feature_sptr> vf = curr_feat->features();
+  auto vf = curr_feat->features();
+  auto vd = curr_desc->descriptors();
 
   track_id_t next_track_id = 0;
 
   // special case for the first frame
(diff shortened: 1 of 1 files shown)


mleotta:
Jenkins test this please

mleotta:
So it turns out that I have an even simpler fix.  There is an issue with casting between const and non-const iterators.  `begin()` returns a non-const iterator but we assign it to a `const_iterator`.  If we replace `begin()` with `cbegin()` and `end()` with `cend()` that should fix it.  I'll push up an alternate solution that does that.

mleotta:
superseded by #1219

Issue #995, opened 2020-05-18
Undefined behavior when dereferencing iterator for kwiver::arrows::ocv::descriptor_set after copying


neal-siekierski:
When running [`stabilize_images.pipe`](https://github.com/Kitware/kwiver/blob/97c76e182eb991d41f6d6c9fd60ad046d26910aa/examples/pipelines/stabilize_images.pipe.in), I reliably get a segfault on the second input frame. I investigated and narrowed the problem down to here:

https://github.com/Kitware/kwiver/blob/97c76e182eb991d41f6d6c9fd60ad046d26910aa/arrows/core/track_features_core.cxx#L376-L378

Changing the last line to `auto dit = curr_desc->begin();` removes the issue. This appears to be because the (unelidable) copy in the existing code puts the iterator in a bad state because the `std::function` returned by `kwiver::arrows::ocv::descriptor_set::get_iter_next_func` returns an internal reference (see [here](https://github.com/Kitware/kwiver/blob/97c76e182eb991d41f6d6c9fd60ad046d26910aa/arrows/ocv/descriptor_set.cxx#L276)), which is stored in the `m_cur_val_ptr` of the iterator but invalidated by the destruction of the original. The first subsequent use of `*dit` returns a bogus value (the undefined behavior), ultimately crashing somewhere completely different on the next frame when trying to interpret said value to perform a dynamic cast:

https://github.com/Kitware/kwiver/blob/97c76e182eb991d41f6d6c9fd60ad046d26910aa/arrows/ocv/descriptor_set.cxx#L90

This was a debug build of recent `master` (2fca7dacd) with GCC 9.3.1. That said, by my reading of <https://en.cppreference.com/w/cpp/utility/functional/function/function> this is expected behavior for `std::function`:

> 3-4) \[The copy constructor c\]opies \[...\] the *target* of `other` to the *target* of `*this`

As I understand it, this is contrary to the documentation of `kwiver::vital::iterator`, which says that "the next-value function's state is not fully copied" (see [here](https://github.com/Kitware/kwiver/blob/97c76e182eb991d41f6d6c9fd60ad046d26910aa/vital/iterator.h#L254)).

I suppose the fix would be to wrap `m_next_value_func` in a `shared_ptr`, which should give the documented shallow-copy semantics at the cost of another pointer indirection. @Purg, thoughts? If that sounds like the right way forward I can work on a PR.

PR #1219, opened 2021-02-26
Fix iterator conversion crash

mleotta:


mleotta:
@borovik135 here is a simple 2-character fix for the same issue addressed in #1218.  I've tested this on Linux and it fixes the issue for me.   The issue is with converting iterators to const iterators.  Separately, I will try to make a unit test that currently fails as a result of this bug, so we can try to fix it.  But that is not needed for the patch release.

You might want to try merging this into your local master to see if it resolves the issue for you.  I expect it will.

Purg (approved):
These changes LGTM, types are now manually aligned. I will have to, as a separate issue, refresh myself with the underlying iterator components to see what can be done to improve/fix compatibility or type safety.

borovik135 (approved):

borovik135:
I think those make sense for the time being. Tnx.

Purg:
I found an alternative "fix". I put quotes around that because I'm not 100% there isn't something to improve/fix in the base iterable/iterator stuff.

So I read a few community posts [1,2] that detail complications around that capturing shared pointers lambdas, probably alluding to that as bad practice and can lead to weird behavior, usually circular reference leading to memory leaks (we're seeing the opposite, though?). Revisiting what's happening in the OCV `get_[const_]iter_next_func` methods, and why the capture was happening, a in-lambda static reference would do logically the same job, but transferring prime ownership of shared pointer to the function instance, at least to my loose understanding... Empirically anyway, applying the following diff causes the reference to no longer be lost and the test is passing:
```diff
diff --git a/arrows/ocv/descriptor_set.cxx b/arrows/ocv/descriptor_set.cxx
index 9947a4c74..cf76a8d8b 100644
--- a/arrows/ocv/descriptor_set.cxx
+++ b/arrows/ocv/descriptor_set.cxx
@@ -252,8 +252,8 @@ ::get_iter_next_func()
   size_t row_counter = 0;
   // Variable for copy into the lambda instance to hold the current row
   // descriptor reference.
-  vital::descriptor_sptr d_sptr;
-  return [row_counter,d_sptr,this] () mutable ->iterator::reference {
+  return [row_counter,this] () mutable ->iterator::reference {
+    static vital::descriptor_sptr d_sptr;
     if( row_counter >= size() )
     {
       VITAL_THROW( vital::stop_iteration_exception, "descriptor_set" );
@@ -285,8 +285,8 @@ ::get_const_iter_next_func() const
   size_t row_counter = 0;
   // Variable for copy into the lambda instance to hold the current row
   // descriptor reference.
-  vital::descriptor_sptr d_sptr;
-  return [row_counter,d_sptr,this] () mutable ->const_iterator::reference {
+  return [row_counter,this] () mutable ->const_iterator::reference {
+    static vital::descriptor_sptr d_sptr;
     if( row_counter >= size() )
     {
       VITAL_THROW( vital::stop_iteration_exception, "descriptor_set" );
```

In an earlier offline discussion with @mleotta it was brought up that there "should" be a way to better protect a user from making this kind of sneaky mistake. I'm not sure what that mechanism would be here, given this most recent surface re-investigation. There might be, but there might also not be without rethinking how this iterator/iterable construct works as a whole. I've only heard over time that this creation of mine is unduly complicated and has caused audible pain (sorry...).

[1] https://floating.io/2017/07/lambda-shared_ptr-memory-leak/
[2] https://eklitzke.org/notes-on-std-shared-ptr-and-std-weak-ptr

mleotta:
@Purg This fix seems like it could be problematic.  According to my understanding of how `static` variables and lambdas work together (see [1]) if you create multiple iterators they would each share the same static `d_sptr`.  A common use case of iterators is to have two nested loops that iterate through a set making pairwise comparisons.  The `static` variable would not allow you to have two different iterators pointing to two different items in the set.

[1] https://stackoverflow.com/questions/8391058/how-do-static-variables-in-lambda-function-objects-work

mleotta:
I'm going to merge this so that we have a work code base for the upcoming patch release, but we should continue to investigate ways to fix the code to be more robust for future use.

Purg:
> According to my understanding of how `static` variables and lambdas work together (see [1]) if you create multiple iterators they would each share the same static `d_sptr`.

Oops, apologies. I think my python is leaking.

mleotta:
I haven't tested this, but do we need that `d_sptr` to be captured or global at all?  It's only the `row_counter` that we need to keep track of.  So it seems like `d_sptr` could just be a regular local variable inside the scope of the lambda.  I haven't tested this to see if that changes how the failing test behaves.

Purg:
I'm in the code adding a test around the nested iterator use you mentioned to enforce my own understanding. I can try that while I'm in here.

mleotta:
Okay, I see where my suggestion breaks down.  This function needs to return a reference so it would return a reference to a temporary.  The main problem is that iterators are really designed to give access to underlying data, rather than construct intermediate representations and give access to those.  In fact, I would argue that the non-const iterator on the ocv `descriptor_set` should not be allowed because we cannot iterate through and change the values.  At least not without implementing some sort of proxy descriptor class that can reference the row of a `cv::Mat`.

borovik135:
Interesting discussion. I have taken another look at the code and found that ocv_to_vital_descriptor copies a matrix row into a vital::descriptor_array_of<T> instance, which basically amounts to the legacy code of copying all descriptors into a collection such as std::vector<vital::descriptor_sptr> with subsequent iteration over those. Thus, I fail to see what the new vital iterator actually gives us performance-wise. To implement some sort of in-place iteration over cv::Mat rows, our vital::descriptor::iterator dit should a) keep track of the current row index instead of or with the the current pointer, and b) *dit should return cv::Mat that it gets via data_.row(ndx) to achieve O(1) performance row-wise. I am not exactly sure how to implement that in the existing framework, though.

mleotta:
It's not really a performance issue.  It's a design issue.  Vital is stuck between two different and incompatible design paradigms right now.  Some classes, like the descriptor set, are trying to adhere to both families of capabilities and have ended up broken as a result.  We should discuss this in the next monthly KWIVER meeting.

neal-siekierski:
For what it's worth, issue #995 exists for the underlying issue. As I mentioned there, you could remove the undefined behavior and give the iterators their documented shallow-copy semantics by storing the iterator's function in a `shared_ptr`, though that doesn't address the other issues mentioned here.

PR #1222, opened 2021-03-03
Fix VTK depth loading

mleotta:
The code for loading depth images from a VTK file was storing the data both
flipped horizontally and transposed in the vital image.  The only user of
this data, CUDA depth fusion, was then transposing again during copy to
CUDA memory and flipping during pixel access to compensate.  This worked,
but any other algorithm using the same images would see the same garbled
memory and need would need to do the same operations to make sense of it.

This change removes the flip and transpose in both locations so that the
image passed in between is valid and can be used for other purposes.  This change will not change the outputs current algorithms, but it fixes the intermediate values so that reusing the depth loading code will not cause problems for other future uses.

### DIFF
--- a/arrows/cuda/integrate_depth_maps.cxx
+++ b/arrows/cuda/integrate_depth_maps.cxx
@@ -186,11 +186,11 @@ copy_img_to_gpu(kwiver::vital::image_container_sptr h_img)
 
   //copy to cuda format
   kwiver::vital::image img = h_img->get_image();
-  for (unsigned int i = 0; i < h_img->width(); i++)
+  for (unsigned int j = 0; j < h_img->height(); j++)
   {
-    for (unsigned int j = 0; j < h_img->height(); j++)
+    for (unsigned int i = 0; i < h_img->width(); i++)
     {
-      temp[i*h_img->height() + j] = img.at<double>(i, j);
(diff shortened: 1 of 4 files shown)


arrows/cuda/integrate_depth_maps.cxx
```diff
@@ -186,11 +186,11 @@ copy_img_to_gpu(kwiver::vital::image_container_sptr h_img)
 
   //copy to cuda format
   kwiver::vital::image img = h_img->get_image();
-  for (unsigned int i = 0; i < h_img->width(); i++)
+  for (unsigned int j = 0; j < h_img->height(); j++)
```
russelldj, line 189: This is a very weak suggestion but I found the use of `i` and `j` relatively confusing because they don't agree with the matrix meaning. Perhaps `x` and `y` as in the next section of edited code would be better.

russelldj (approved):

arrows/cuda/integrate_depth_maps.cxx
mleotta: This notation comes from VXL where it is used heavily.  I'm trying to be consistent with that.  The accessors for width and height on a VXL image are `ni()` and `nj()`.  This is the old row-major vs column major debate.

mleotta:
@smithe4 might be interested to know that this change is happening.

PR #1226, opened 2021-03-03
CPU depth fusion

mleotta:
This branch provides the long missing CPU implementation of depth map fusion, which was previously only available as a CUDA algorithm.  This CPU alternative will enable TeleSculptor to run end-to-end without a CUDA enabled GPU.

### DIFF
--- a/arrows/mvg/algo/integrate_depth_maps.cxx
+++ b/arrows/mvg/algo/integrate_depth_maps.cxx
@@ -142,8 +142,8 @@ integrate_depth_maps::priv
 {
   int const u = static_cast<int>(image_pt.x() + 0.5);
   int const v = static_cast<int>(image_pt.y() + 0.5);
-  if (u < 0 || u >= depth.width() ||
-      v < 0 || v >= depth.height())
+  if (u < 0 || u >= static_cast<int>(depth.width()) ||
+      v < 0 || v >= static_cast<int>(depth.height()))
   {
     return 0.0;
   }
@@ -167,8 +167,8 @@ integrate_depth_maps::priv
 {
   int const u = static_cast<int>(image_pt.x() + 0.5);
   int const v = static_cast<int>(image_pt.y() + 0.5);
-  if (u < 0 || u >= depth.width() ||
-      v < 0 || v >= depth.height())
+  if (u < 0 || u >= static_cast<int>(depth.width()) ||
+      v < 0 || v >= static_cast<int>(depth.height()))
   {
     return 0.0;
   }
--- a/arrows/mvg/algo/integrate_depth_maps.cxx
+++ b/arrows/mvg/algo/integrate_depth_maps.cxx
@@ -200,7 +200,7 @@ integrate_depth_maps::priv
     // extrinsic paramters into a single 3x4 projection for faster iteration
     if (weight.size() == 0)
     {
-      auto func = [&](vector_3d const& hpt)
+      auto func = [this, &depth](vector_3d const& hpt)
       {
         vector_2d image_pt{ hpt[0] / hpt[2], hpt[1] / hpt[2] };
         return ray_potiential_at_point(image_pt, hpt[2], depth);
@@ -210,7 +210,7 @@ integrate_depth_maps::priv
     }
     else
     {
-      auto func = [&](vector_3d const& hpt)
+      auto func = [this, &depth, &weight](vector_3d const& hpt)
       {
         vector_2d image_pt{ hpt[0] / hpt[2], hpt[1] / hpt[2] };
         return ray_potiential_at_point(image_pt, hpt[2], depth, weight);
@@ -224,7 +224,7 @@ integrate_depth_maps::priv
     auto const K = camera.intrinsics();
     if (weight.size() == 0)
     {
-      auto func = [&](vector_3d const& hpt)
+      auto func = [this, K, &depth](vector_3d const& hpt)
       {
         vector_2d image_pt = K->map(hpt);
         return ray_potiential_at_point(image_pt, hpt[2], depth);
@@ -234,7 +234,7 @@ integrate_depth_maps::priv
     }
     else
     {
-      auto func = [&](vector_3d const& hpt)
+      auto func = [this, K, &depth, &weight](vector_3d const& hpt)
       {
         vector_2d image_pt = K->map(hpt);
         return ray_potiential_at_point(image_pt, hpt[2], depth, weight);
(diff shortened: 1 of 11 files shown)


arrows/mvg/algo/integrate_depth_maps.cxx
```diff
@@ -0,0 +1,336 @@
...
+  {
+    double const z = origin[2] + (k + 0.5) * spacing[2];
+    for (size_t j = 0; j < nj; ++j)
+    {
+      double const y = origin[1] + (j + 0.5) * spacing[1];
+      for (size_t i = 0; i < ni; ++i)
+      {
+        double const x = origin[0] + (i + 0.5) * spacing[0];
+        vector_3d world_pt = vector_3d{ x, y, z };
+        vector_2d image_pt = camera.project(world_pt);
```
mleotta, line 107: This could be optimized in the future to break apart the common parts of the projection and reuse some computation across voxels and also avoid duplication of computation with `depth()` in L123 below.  But we can optimize in a follow-up branch.

arrows/mvg/algo/integrate_depth_maps.cxx
```diff
@@ -0,0 +1,336 @@
...
+// ----------------------------------------------------------------------------
+
+/// Get this algorithm's \link vital::config_block configuration block \endlink
+vital::config_block_sptr
+integrate_depth_maps::get_configuration() const
+{
+  // get base config from base class
+  auto config = vital::algo::integrate_depth_maps::get_configuration();
+
+  config->set_value("ray_potential_thickness", d_->ray_potential_thickness,
```
mleotta, line 178: These config values are all shared with the CUDA version.  It may make sense to make both of these classes share a common base class that handles common configuration.

borovik135 (approved):
makes sense to use a math header for common constants such as PI

borovik135 (approved):
OK, a note on the expected performance penalty may be nice

arrows/mvg/algo/integrate_depth_maps.cxx
borovik135: Speeding things up a bit one could increment x (y and z) by the respective amount, e.g. x+=dx, to avoid multiplication and array access costs, but I suspect this is not the slowest part in the loop. I would also avoid the extra copy construction of vector_3d, but define and use a regular constructor, e.g. vector_3d world_pt(x, y, z).

arrows/mvg/algo/integrate_depth_maps.cxx
borovik135: this sounds reasonable to me as well

borovik135:
Matt, let me know if you'd like me to dive in the depth fusion to potentially provide more algorithmic improvements.

arrows/mvg/algo/integrate_depth_maps.cxx
mleotta: I have several changes locally to optimize this.  I've got it running about 10x faster in the special case where there is no lens distortion.  I think it make sense to get this version working and merged before addressing optimizations.

mleotta:
I've changed my mind.  It seems several test are failing due to timeout (30 seconds).  My optimized code should run much faster, so I will push that, along with updated tests, now.

mleotta:
@borovik135 I've pushed my optimized code.  On Windows, where I wrote and tested it, it runs very fast.  The unit tests all run the depth map integration in less than a second.  However, on the Linux dashboards they are still taking more than 10 seconds and one test is timing out after 30 seconds.  So something is very different between how the compilers are optimizing this code.  If you could try this on linux and see what you get, that might be helpful.

borovik135:
> @borovik135 I've pushed my optimized code. On Windows, where I wrote and tested it, it runs very fast. The unit tests all run the depth map integration in less than a second. However, on the Linux dashboards they are still taking more than 10 seconds and one test is timing out after 30 seconds. So something is very different between how the compilers are optimizing this code. If you could try this on linux and see what you get, that might be helpful.

Matt, I am taking a look at the issue, checked out the pull request. I assume in CMake I need to enable tests. Do I also need to add test targets? what test do I run to address the issue?

mleotta:
I've done a few more experiments on Linux and have a somewhat better understanding of this speed discrepancy.  This issue is that it appears OpenMP puts all threads on the same CPU core by default on Linux unless you set the environment variable `OMP_PROC_BIND=spread`.  When all threads are on the same CPU core there is little benefit to multi-threading.  I haven't found a way to change this default behavior without the environment variable.  Secondly, it seems that there is more overhead in setting up the threads on Linux than Windows.  If I comment out the OpenMP pragma, the Linux tests run twice as fast (on a single core).  There may still be other factors at play.  Windows still seems considerably faster than Linux when OpenMP calls are removed, but I'm not comparing on the same hardware, so it's not a fair comparison.

borovik135:
> OMP_PROC_BIND=spread

I rebuilt my kwiver with the noted pull request setting CMake build type explicitly to Release.
Now I am getting the following timing figures on the test. No environment manipulation has been done.
Looks pretty quick to me. I think, explicit requiring Release build type sets the necessary optimization options for the compiler, while omitting the build type leaves some defaults, which may not be optimal for speed.
==
[       OK ] integrate_depth_maps.create (1233 ms)
[ RUN      ] integrate_depth_maps.integrate
integration time: 1.62207
[       OK ] integrate_depth_maps.integrate (83 ms)
[ RUN      ] integrate_depth_maps.integrate_weighted
integration time: 1.46927
[       OK ] integrate_depth_maps.integrate_weighted (68 ms)
[ RUN      ] integrate_depth_maps.integrate_distorted
integration time: 1.57457
[       OK ] integrate_depth_maps.integrate_distorted (67 ms)
[----------] 4 tests from integrate_depth_maps (1451 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (1451 ms total)
[  PASSED  ] 4 tests.

mleotta:
@borovik135 if you run `top` while running this test, can you see if it is hitting all of your cores?  Maybe there is just something weird about my system that doesn't distribute OpenMP threads by default.

I've cut the number of images used in the test in half.  Hopefully that will make it quick enough for the slower CI machines so they don't time out.

borovik135:
> @borovik135 if you run `top` while running this test, can you see if it is hitting all of your cores? Maybe there is just something weird about my system that doesn't distribute OpenMP threads by default.
> 
> I've cut the number of images used in the test in half. Hopefully that will make it quick enough for the slower CI machines so they don't time out.

my system monitor shows consistent spikes on all of the cores when I run the test, so I believe the threads get distributed well

PR #1231, opened 2021-03-04
Fixes to metadata extraction for ffmpeg reader

chetnieter:
This fixes two issues with the metadata extraction from video sources using ffmpeg. The first fix allows the video reader to pull metadata from multiple video streams. The second fix associates the metadata with the correct video frame when it arrives early in the data stream.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_input.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_input.cxx
@@ -25,7 +25,6 @@
 #include <deque>
 #include <mutex>
 #include <memory>
-#include <unordered_set>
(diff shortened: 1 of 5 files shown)


mleotta:
@chetnieter it looks like we need to add `VIDEO_DATA_STREAM_INDEX` to a python test that checks the metadata tags.

https://github.com/Kitware/kwiver/blob/master/python/kwiver/vital/tests/test_metadata_tags.py

mleotta:
A release note is a good idea.  We haven't been as strict about requiring release notes as we should.

mleotta (approved):

PR #1232, opened 2021-03-05
Update burnout pipeline to add averaging, closing, and dilation

russelldj:
This makes the results significantly better. It seems like the original BurnOut used the following thresholding paramters for lines 55 and 56, but they seem to significantly undersegment representative data.
```
filter:vxl_threshold:type             = absolute
filter:vxl_threshold:threshold        = -0.037
```


mleotta (approved):

russelldj:
This relies on #1227 and #1229 .

russelldj:
For reference, this no longer relies on #1229.

Also, is it worth me breaking this up into an include-able pipeline and then include that in an example with endpoints?

mleotta:
I will merge this as is, but yes we should break this out into an includable pipeline and move that includable part into `config/pipelines`.  Please coordinate with @mwoehlke-kitware.  I think he was going to help make `config/pipelines` files installed.

PR #1243, opened 2021-03-11
Add json serialization of metadata maps

russelldj:
This is code which was internally reviewed [here](https://kwgitlab.kitware.com/computer-vision/KWIVER/-/merge_requests/248). It adds the ability to read and write video metadata as a json. Also, it updates `dump_klv` to have the option to write out the metadata.

I then rebased the internal changes onto `master`, cleaned up the history a bit, and autoformatted the two four new files, `arrows/serialize/json/serialize_metadata.{cxx, h}` and `vital/algo/serialize_metadata.{cxx, h}`.


arrows/serialize/json/serialize_metadata.h
```diff
@@ -0,0 +1,79 @@
...
+  /// Destructor
+  ~serialize_metadata();
+
+  /// Implementation specific load functionality.
+
+  /**
+   * \param filename the path to the file the load
+   * \returns an image container refering to the loaded image
+   */
+  virtual kwiver::vital::metadata_map_sptr load_( std::string const& filename )
```
mleotta, line 53: minor formatting thing:   It looks odd that the `const` is wrapped to the next line.  Usually when the function declaration is too long we break after the return type.  E.g.
```c++
virtual kwiver::vital::metadata_map_sptr
load_( std::string const& filename ) const;
```

arrows/serialize/json/tests/test_load_save.cxx
```diff
@@ -503,6 +504,58 @@ TEST( load_save, metadata_vector )
   }
 }
 
+// ----------------------------------------------------------------------------
+TEST( load_save, metadata_map )
+{
+  kwiver::vital::metadata_sptr meta = std::make_shared<kwiver::vital::metadata>( create_meta_collection() );
```
mleotta, line 510: this line is quite long.  Maybe wrap, or use `auto` or use namespace alias.

vital/algo/serialize_metadata.h
```diff
@@ -0,0 +1,134 @@
...
+
+namespace kwiver {
+
+namespace vital {
+
+namespace algo {
+
+/// An abstract base class for reading and writing meatadata maps
+
+/**
```
mleotta, line 32: We might take this opportunity to switch to the new standard for KWIVER doxygen where we use only `///` for each line and not `/**`.

arrows/serialize/json/serialize_metadata.cxx
```diff
@@ -0,0 +1,122 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/**
+ * \file
+ * \brief Implementation of load/save functionality.
```
mleotta, line 7: Implementation of **metadata** load/save functionality

arrows/serialize/json/metadata.h
```diff
@@ -27,8 +29,12 @@ class KWIVER_SERIALIZE_JSON_EXPORT metadata
 
   metadata();
   virtual ~metadata();
-
+  std::shared_ptr< std::string > serialize_meta( const vital::metadata_vector& elements );
```
mleotta, line 32: consider wrapping long lines

arrows/serialize/json/load_save_metadata.cxx
```diff
@@ -67,32 +66,50 @@ struct meta_item
...
+      type = "geo-polygon";
     }
     else
     {
       //+ throw something
     }
+
+    // These two items are included to increase readability of the
+    // serialized form and are not used when deserializing.
+    archive(  ::cereal::make_nvp( "name", trait.name() ) );
```
mleotta, line 111: looks like extra whitespace before `::cereal`

arrows/serialize/json/load_save_metadata.cxx
```diff
@@ -67,32 +66,50 @@ struct meta_item
...
     }
     else if ( trait.tag_type() == typeid( kwiver::vital::geo_point ) )
     {
       const kwiver::vital::geo_point value = kwiver::vital::any_cast<kwiver::vital::geo_point  > ( this->item_value );
       archive( CEREAL_NVP( value ) );
+      type = "geo-point";
     }
     else if ( trait.tag_type() == typeid( kwiver::vital::geo_polygon ) )
     {
       const kwiver::vital::geo_polygon value = kwiver::vital::any_cast<kwiver::vital::geo_polygon  > ( this->item_value );
```
mleotta, line 100: I know this is not really part of your change, but might be useful to break long lines here or use `auto` and also remove extra whitespace in `geo_polygon  >`

arrows/serialize/json/load_save.h
```diff
@@ -63,6 +64,11 @@ void save( ::cereal::JSONOutputArchive& archive, const ::kwiver::vital::image_co
 KWIVER_SERIALIZE_JSON_EXPORT
 void load( ::cereal::JSONInputArchive& archive, ::kwiver::vital::image_container_sptr& obj );
 
+KWIVER_SERIALIZE_JSON_EXPORT
+void save( ::cereal::JSONOutputArchive& archive, const ::kwiver::vital::metadata_map::map_metadata_t& meta );
```
mleotta, line 68: break long lines

mleotta (commented):
Generally looks good.  I've added some comments about formatting.  We don't necessarily need to fix all the formatting, especially not for code that wasn't changed, but we may want to break some of the long lines.

arrows/serialize/json/tests/test_load_save.cxx
```diff
@@ -503,6 +504,58 @@ TEST( load_save, metadata_vector )
...
+  kwiver::vital::metadata_map::map_metadata_t obj_dser;
+  {
+    cereal::JSONInputArchive ar( msg );
+    cereal::load( ar, obj_dser );
+  }
+
+  EXPECT_EQ( meta_map.size(), obj_dser.size() );
+
+  // Check to make sure they are the same
+  for ( auto const& item : meta_map)
```
mwoehlke-kitware, line 549:
```suggestion
  for( auto const& item : meta_map )
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+INSTANTIATE_ALGORITHM_DEF( kwiver::vital::algo::serialize_metadata );
+/// \endcond
+
+namespace kwiver {
+
+namespace vital {
+
+namespace algo {
+
+serialize_metadata
```
mwoehlke-kitware, line 29:
```suggestion
// ----------------------------------------------------------------------------
serialize_metadata
```

...and likewise throughout.

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+
+metadata_map_sptr
+serialize_metadata
+::load( std::string const& filename ) const
+{
+  // Make sure that the given file path exists and is a file.
+  if( !kwiversys::SystemTools::FileExists( filename ) )
+  {
+    VITAL_THROW( path_not_exists,
+                 filename );
```
mwoehlke-kitware, line 43:
```suggestion
    VITAL_THROW( path_not_exists, filename );
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+::serialize_metadata()
+{
+  attach_logger( "algo.serialize_metadata" );
+}
+
+metadata_map_sptr
+serialize_metadata
+::load( std::string const& filename ) const
+{
+  // Make sure that the given file path exists and is a file.
```
mwoehlke-kitware, line 39:
```suggestion
  // Make sure that the given file path exists and is a file
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+  if( !kwiversys::SystemTools::FileExists( filename ) )
+  {
+    VITAL_THROW( path_not_exists,
+                 filename );
+  }
+  else if( kwiversys::SystemTools::FileIsDirectory( filename ) )
+  {
+    VITAL_THROW( path_not_a_file,
+                 filename );
```
mwoehlke-kitware, line 48:
```suggestion
    VITAL_THROW( path_not_a_file, filename );
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  return this->load_( filename );
+}
+
+void
+serialize_metadata
+::save( std::string const& filename,
+        metadata_map_sptr data ) const
```
mwoehlke-kitware, line 57:
```suggestion
::save( std::string const& filename, metadata_map_sptr data ) const
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+{
+  // Make sure that the given file path's containing directory exists and is
+  // actually a directory.
```
mwoehlke-kitware, line 60:
```suggestion
  // actually a directory
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+  std::string containing_dir =
+    kwiversys::SystemTools::GetFilenamePath(
+      kwiversys::SystemTools::CollapseFullPath( filename ) );
+
+  if( !kwiversys::SystemTools::FileExists( containing_dir ) )
+  {
+    VITAL_THROW( path_not_exists,
+                 containing_dir );
```
mwoehlke-kitware, line 68:
```suggestion
    VITAL_THROW( path_not_exists, containing_dir );
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+  }
+  else if( !kwiversys::SystemTools::FileIsDirectory( containing_dir ) )
+  {
+    VITAL_THROW( path_not_a_directory,
+                 containing_dir );
```
mwoehlke-kitware, line 73:
```suggestion
    VITAL_THROW( path_not_a_directory, containing_dir );
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  this->save_( filename,
+               data );
```
mwoehlke-kitware, line 77:
```suggestion
  this->save_( filename, data );
```

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+}
+
+const vital::algorithm_capabilities&
```
mwoehlke-kitware, line 80:
```suggestion
vital::algorithm_capabilities
```

Note the deliberate removal of `const&` (will need to change in header also).

vital/algo/serialize_metadata.cxx
```diff
@@ -0,0 +1,114 @@
...
+  return this->m_capabilities;
+}
+
+void
+serialize_metadata
+::set_capability( algorithm_capabilities::capability_name_t const& name,
+                  bool val )
+{
+  this->m_capabilities.set_capability( name,
+                                       val );
```
mwoehlke-kitware, line 93:
```suggestion
  this->m_capabilities.set_capability( name, val );
```

vital/algo/serialize_metadata.h
```diff
@@ -0,0 +1,134 @@
...
+#define VITAL_ALGO_SERIALIZE_METADATA_H_
+
+#include <vital/vital_config.h>
+
+#include <string>
+
+#include <vital/algo/algorithm.h>
+#include <vital/algorithm_capabilities.h>
+#include <vital/types/metadata.h>
+#include <vital/types/metadata_map.h>
```
mwoehlke-kitware, line 22:
```suggestion
#include <vital/algo/algorithm.h>

#include <vital/algorithm_capabilities.h>

#include <vital/types/metadata.h>
#include <vital/types/metadata_map.h>

#include <vital/vital_config.h>

#include <string>
```

vital/algo/serialize_metadata.h
```diff
@@ -0,0 +1,134 @@
...
+
+namespace kwiver {
+
+namespace vital {
+
+namespace algo {
+
+/// An abstract base class for reading and writing meatadata maps
```
mwoehlke-kitware, line 30: Again, following the updated style recommendations would be nice...

vital/algo/serialize_metadata.h
```diff
@@ -0,0 +1,134 @@
...
+   * \param data the metadata_map_sptr pointing to the metadata
+   */
+  virtual void save_( std::string const& filename,
+                      kwiver::vital::metadata_map_sptr data ) const = 0;
+
+  algorithm_capabilities m_capabilities;
+};
+
+/// Shared pointer type for generic serialize_metadata definition type.
+typedef std::shared_ptr< serialize_metadata > serialize_metadata_sptr;
```
mwoehlke-kitware, line 126:
```suggestion
using serialize_metadata_sptr = std::shared_ptr< serialize_metadata >;
```

arrows/serialize/json/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ set( headers_public
   metadata.h
   object_track_set.h
   object_track_state.h
+  serialize_metadata.h
```
linus-sherrill, line 17: This serializer used by dump_klv is only built if JSON serialization is selected in CMake, so there is a subtle dependency on this CMake option if you want to dump metadata to a file.

vital/types/metadata_traits.h
```diff
@@ -56,6 +57,7 @@ struct vital_meta_trait_base
     static std::string  description() { return std::string(LD); }       \
     static std::type_info const& tag_type() { return typeid(T); }       \
     static bool is_integral() { return std::is_integral<T>::value; }    \
+    static bool is_signed() { return std::is_signed<T>::value; }    \
```
mwoehlke-kitware, line 60:
```suggestion
    static bool is_signed() { return std::is_signed<T>::value; }        \
```

arrows/serialize/json/load_save_metadata.cxx
mwoehlke-kitware: :+1:, balancing the whitespace in `<>` would be nice also. (It looks like it is probably okay to follow the style guide's prescription to include space inside `<>`, as the surrounding code seems to generally use space inside open/close symbols.)

vital/algo/serialize_metadata.h
mwoehlke-kitware: Yup, noted also in my review.

arrows/core/applets/dump_klv.cxx
```diff
@@ -44,7 +50,9 @@ add_command_options()
     ( "h,help", "Display applet usage" )
     ( "c,config", "Configuration file for tool", cxxopts::value<std::string>() )
     ( "o,output", "Dump configuration to file and exit", cxxopts::value<std::string>() )
+    ( "l,log", "Log metadata to a .json file.", cxxopts::value<std::string>() )
```
linus-sherrill, line 53: This specifically indicates that the output will be JSON, but it is possible to specify a different type of output using the config file. If you are only supporting JSON, you could programmatically force that setting into the config. this would also get around a possible error loading the serializer if it is not selected on the command line and not available in the build (as described below).

Also, the JSON output is in a specific format that is not exactly the same as the stdout dump. Maybe all that is required is to explain the specific metadata map format. If somebody else wants to use JSON format, it should be clear what they are getting.

arrows/serialize/json/serialize_metadata.h
mwoehlke-kitware: :+1: I suggested breaking after the `(`, but this would be okay also.

Also, `override` is missing, and I believe we decided to prefer to omit `virtual`.

arrows/serialize/json/register_algorithms.cxx
```diff
@@ -66,6 +67,9 @@ register_factories( kwiver::vital::plugin_loader& vpm )
   reg.register_algorithm< string >( "kwiver:video_name" );
 
   reg.mark_module_as_loaded();
+
+  // register the algo which writes metadata maps to disk
+  vpm.ADD_ALGORITHM("json", serialize_metadata);
```
linus-sherrill, line 72: It seems a little odd that an algorithm is being added this way, although it is supported. Maybe move this above the `mark_module_as_loaded()` call.

linus-sherrill (changes requested):

arrows/core/applets/dump_klv.cxx
```diff
@@ -110,86 +122,130 @@ run()
...
+    }
+
+    kv::metadata_vector metadata = video_reader->frame_metadata();
 
-    kwiver::vital::metadata_vector metadata = video_reader->frame_metadata();
-    for( auto meta : metadata )
+    if ( log )
     {
-      std::cout << "\n\n---------------- Metadata from: " << meta->timestamp() << std::endl;
+      int timestamp = ts.get_frame();
```
mwoehlke-kitware, line 194: FYI, from the "reasons I like `auto`" front... `int` is the wrong type here. (That said, the temporary really isn't needed at all...)

arrows/serialize/json/register_algorithms.cxx
mwoehlke-kitware: Additionally, this differs from uses elsewhere in that it does not set attributes on the algorithm.

mwoehlke-kitware:
I pushed a commit addressing most of the style issues; PTAL, thanks!

Things I did *not* address:
- Most of Linus' comments
- The `int` vs. `int64_t` question

On the latter, it wasn't clear to me how values get into the `any`. At minimum, I think there is a change missing later in that file where it seems numbers are (still) only handled as `uint64_t`. However, it isn't obvious how we know that the `any` won't contain other types that we might need to handle. (AFAICT, `any_cast` does not support conversions, so e.g. an `any` that contains a `short` will throw if you try to cast to `int`.)

arrows/serialize/json/CMakeLists.txt
russelldj: There is a `core` implementation of csv metadata writing which was supposed to address this issue. It looks like it is still [solely internal](https://kwgitlab.kitware.com/computer-vision/KWIVER/-/blob/nga-klv/master/arrows/core/serialize_metadata_csv.cxx) though. I plan to open a WIP PR for that for bookkeeping purposes, but I think this is a much higher priority.

russelldj:
@mwoehlke-kitware Thank you for all this cleanup. I reviewed your changes and they all look good to me. With respect to both sets of the comments I'll likely need to reacquaint myself with this for a bit before addressing them.

arrows/serialize/json/serialize_metadata.cxx
```diff
@@ -0,0 +1,122 @@
...
+  else
+  {
+    VITAL_THROW( kwiver::vital::file_not_read_exception, filename,
+                 "Insufficient permissions or moved file" );
+  }
+
+  auto metadata_map_ptr =
+    std::make_shared< kwiver::vital::simple_metadata_map >(
+      kwiver::vital::simple_metadata_map( metadata_map ) );
+  return metadata_map_ptr;
```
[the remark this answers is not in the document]
russelldj: I think I was initially overzealous w.r.t. creating temporary variables.

vital/algo/serialize_metadata.cxx
russelldj: Is this a danger for const-qualified methods?

mwoehlke-kitware: If you return a reference, the caller might accept the value as a reference. If the owning object then goes away, you have a problem. We've had issues with this in the past.

I ended up not changing this because, although this isn't actually a `virtual` method, the same API does exist elsewhere, so for the moment I elected to keep it consistent. We should, however, consider changing it, but ideally we would change all of them, so we may want to revisit this as a follow-up.

arrows/serialize/json/load_save_metadata.cxx
```diff
@@ -67,32 +66,50 @@ struct meta_item
       {
         const bool value = kwiver::vital::any_cast< bool > ( this->item_value );
         archive( CEREAL_NVP( value ) );
+        type = "boolean";
+      }
+      // We don't want negative ints to be serialized as large postive unsigned
+      // values since this would be confusing for external applications
+      else if ( trait.is_signed() )
+      {
+        const int value = kwiver::vital::any_cast< int > ( this->item_value );
```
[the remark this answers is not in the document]
russelldj: There's only a limited set of metadata tags and the only time a signed integer comes up it is an `int`. However, I I think it makes sense to update this as you suggested to avoid ugly mistakes in the future. Though I understand this does induce a slight performance penalty.

vital/algo/serialize_metadata.h
```diff
@@ -0,0 +1,134 @@
...
+namespace algo {
+
+/// An abstract base class for reading and writing meatadata maps
+
+/**
+ * This class represents an abstract interface for reading and writing
+ * video metadata.
+ */
+class VITAL_ALGO_EXPORT serialize_metadata
+  : public kwiver::vital::algorithm_def< serialize_metadata >
```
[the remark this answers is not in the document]
russelldj: @mleotta might remember better than I do, but I think it was because `vital/algo/data_serializer`, which I believe you're referring to, goes from data to a string representation and vice versa. The goal of this metadata serialization is to explicitly read and write to disk.

Perhaps this indirectly addresses the other two questions. I was initially just going to update this to `serialize_metadata_map` but maybe something like `metadata_map_io` would make more sense. I think it should be possible to switch to taking streams as input but there's a bit of logic for checking that filenames are valid which I think would have to be moved to the calling code if this switched to taking streams. That makes me a little hesitant to switch to streams but if you feel it's important to switch I can.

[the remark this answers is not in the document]
mwoehlke-kitware: Another option might be to add *overloads* taking streams, and have the overloads that take files do the checking, open the file (in order to get a stream), then hand off to the stream overloads.

Alternatively, maybe we should just have a utility function that tries to return a stream (may need to return it as a `unique_ptr`) from a file.

arrows/core/applets/dump_klv.cxx
russelldj: I added a rather lengthy explanation here. I'm happy to pare it down if this is too much.

arrows/serialize/json/register_algorithms.cxx
russelldj: I moved this up as suggested. I tried to do this the same way as the other algorithms but this resulted in the implementation type changing to `serialize-json`, which seems unhelpful. See the output from `plugin_explorer` below:
```
Info on algorithm type "serialize-json" implementation "json_serialize_metadata_map"
  Plugin name: json_serialize_metadata_map
      Perform IO on video metadata using JSON.
```

I'm happy to look further into this if that's useful, but I know very little about the registration process.

vital/algo/serialize_metadata.h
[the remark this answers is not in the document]
russelldj: I feel like the overload makes sense in this case. But I do also think the utility function you describe could be broadly useful.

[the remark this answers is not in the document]
linus-sherrill: The data serializers have very specific requirements along with naming conventions which are not really what you want here. I like the comment about making this `metadata_map_io` to separate it from the serializers and overloading to handle file names and streams seems like a good approach. These type of algorithms/classes get used in unexpected ways if they turn out to be useful.

arrows/serialize/json/register_algorithms.cxx
linus-sherrill: The naming convention for serializers is formalized to support the serialization and deserialization processes which select the appropriate algorithm based on the sprokit semantic type name and the type of serialization to be done.

arrows/serialize/json/load_save_metadata.cxx
```diff
@@ -48,51 +51,69 @@ struct meta_item
...
     archive( CEREAL_NVP( tag ) );
 
-    // These two items are included to increase readability of the
-    // serialized form and are not used when deserializing.
-    archive(  ::cereal::make_nvp( "type", item_value.type_name() ) );
-    archive(  ::cereal::make_nvp( "name", trait.name() ) );
-
+    std::string type;
     // This is a switch on the item data type
     if ( trait.is_floating_point() )
```
mwoehlke-kitware, line 56: So... to update everyone else, @russelldj and I had a long offline discussion on this code. Summary: I am concerned that we need to recover the exact type, or users are likely to be confused because their `any_cast`s will unexpectedly fail. To that end, the `is_integral`, etc. methods are not useful, and in fact may be actively harmful. I believe this whole lot should be rewritten to handle many more data types (at minimum, `float`, `double` and every distinct integer type, plus e.g. strings and geo-points). However, I also want to get this PR landed, so we're thinking to address this as a follow-up.

Meanwhile, contrary to my previous comment, we *do* want to handle `int`, as that is the type used by tagged metadata items.

arrows/serialize/json/register_algorithms.cxx
```diff
@@ -37,43 +39,45 @@ KWIVER_SERIALIZE_JSON_PLUGIN_EXPORT
...
 register_factories( kwiver::vital::plugin_loader& vpm )
 {
-  kwiver::vital::serializer_registrar reg( vpm, "arrows.serialize.json",
-                                           "json" );
+  auto const module_name = std::string{ "arrows.serialize.json" };
+  kwiver::vital::serializer_registrar sreg( vpm, module_name, "json" );
+  kwiver::vital::algorithm_registrar areg( vpm, module_name );
 
-  if (reg.is_module_loaded())
+  if( sreg.is_module_loaded() )
```
mwoehlke-kitware, line 46: FYI for other reviewers; yes, it's fine that we only use `sreg` to deal with is-loaded / mark-loaded. These just pass the module name to relevant methods of the plugin manager, and the module name is the same for both `sreg` and `areg`, so we don't need to check/use both, and `sreg`/`areg` may be used interchangeably for these operations.

arrows/serialize/json/load_save_metadata.cxx
linus-sherrill: How does `std::any` handle this. We should be compatible with that so it can just drop in when the time comes.

linus-sherrill: Our `kwiver::vital::any` is designed to be compatible with the soon to be available `std::any` and if we extend our implementation, it should be compatible with the `std` version. Ideally, when we get to the point where we can use C++17 features, all we have to do is replace the type  `kwiver::vital::any` with `std::any`.

russelldj: If I understand what you're asking, this is a hack because the only floating point type we have in [metadata_tags](https://github.com/Kitware/kwiver/blob/master/vital/types/metadata_tags.h) is a double. However, it seems like from what @mwoehlke-kitware has described, if someone added a floating point type which was not a double, `any_cast` would break.

russelldj: Ahh sorry I completely missed the `std` in your comment. I can't really comment on the behavior of `std::any` but I can look into it if you'd like. Also, I'm on the same page as Matthew that this should land before we try to address any of these subtleties since there's other work depending on this code.

mwoehlke-kitware: @linus-sherrill, AFAIK, a) `kwiver::any` and `std::any` have the same behavior, which is that b) both `kwiver::any_cast` and `std::any_cast` perform *no* lexical conversions. That is, you have to ask for *exactly* what you put in. Even (64-bit) `long` and `long long` are not interchangeable.

As far whether `kwiver::any` and `std::any` are "interchangeable" in the behavioral sense, TTBOMK they are.

arrows/serialize/json/metadata_map_io.cxx
```diff
@@ -0,0 +1,96 @@
...
+    buffer << fin.rdbuf();
+
+    auto const& input_string = buffer.str();
+
+    metadata_map = d_->serializer.deserialize_map( input_string );
+  }
+  else
+  {
+    VITAL_THROW( vital::file_not_read_exception, filename,
+                 "Coult not read from stream" );
```
mwoehlke-kitware, line 64:
```suggestion
                 "Could not read from stream" );
```

vital/algo/metadata_map_io.cxx
```diff
@@ -0,0 +1,137 @@
...
+  else if( !kwiversys::SystemTools::FileIsDirectory( containing_dir ) )
+  {
+    VITAL_THROW( path_not_a_directory, containing_dir );
+  }
+
+  /// \brief Default plugin algorithm registration interface implementation.
+
+  std::ofstream fout{ filename };
+
+  save( fout, data, "" );
```
mwoehlke-kitware, line 90: I am guessing you meant:
```suggestion
  save( fout, data, filename );
```

vital/algo/metadata_map_io.cxx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// ----------------------------------------------------------------------------
+void
+metadata_map_io
+::save( std::ostream& fout, metadata_map_sptr data, std::string const& filename ) const
```
mwoehlke-kitware, line 96: Too long :slightly_smiling_face::
```suggestion
::save( std::ostream& fout, metadata_map_sptr data,
        std::string const& filename ) const
```

vital/algo/metadata_map_io.h
```diff
@@ -0,0 +1,139 @@
...
+  ///   a directory).
+  ///
+  /// \param filename The path to the file to load.
+  /// \returns A metadata_map_sptr pointing to the data.
+  kwiver::vital::metadata_map_sptr load( std::string const& filename ) const;
+
+  /// Load metadata from the file.
+  ///
+  /// \param fin The stream to read from.
+  /// \param filename The path to the file to load.
```
mwoehlke-kitware, line 60: I'd suggest something closer to the `save` documentation? "Filename that generated the stream." You might also say (in both places) "May be used for logging, or in case of other operations for which the filesystem path to the input is relevant."

Maybe make similar changes for `load_` and `save_`?

arrows/serialize/json/load_save_metadata.cxx
linus-sherrill: There is an `any_converter.h` in vital/util which, I think, does lexical conversions. I don't know if that is of any use here.

arrows/core/applets/dump_klv.cxx
```diff
@@ -42,31 +51,38 @@ add_command_options()
...
+      "of metadata packets associated with that frame. Each packet is an "
+      "array of metadata fields. Alternatively, the configuration file, "
+      "dump_klv.conf, can be updated to use CSV instead.",
+      cxxopts::value< std::string >() )
     ( "d,detail", "Display a detailed description of the metadata" )
+    ( "q,quiet", "Do not show metadata. Overrides -d/--detail." )
 
     // positional parameters
-    ( "video-file", "Video input file", cxxopts::value<std::string>())
+    ( "video-file", "Video input file", cxxopts::value< std::string >())
```
mwoehlke-kitware, line 66:
```suggestion
    ( "video-file", "Video input file", cxxopts::value< std::string >() )
```

arrows/serialize/json/register_algorithms.cxx
```diff
@@ -16,18 +15,22 @@
 #include "detected_object.h"
 #include "detected_object_set.h"
 #include "detected_object_type.h"
-#include "timestamp.h"
 #include "image.h"
-#include "string.h"
-#include "track_state.h"
+#include "object_track_set.h"
 #include "object_track_state.h"
+#include "metadata_map_io.h"
```
mwoehlke-kitware, line 21: Pedantic: 'm' < 'o' (re-sort includes)

mwoehlke-kitware (changes requested):
Please fix the typo :slightly_smiling_face:

arrows/serialize/json/load_save_metadata.cxx
mwoehlke-kitware: I don't think it's desired here; I think we rather want to preserve exactly the original C++ type. Conversion makes more sense at the point someone goes to try to extract a usable value from the metadata.

mleotta: There is a convention that vital metadata traits should use a limited subset of data types.  The types here should be the only ones in use.

arrows/core/applets/dump_klv.cxx
```diff
@@ -11,18 +11,27 @@
 #include <iostream>
 #include <fstream>
 
+
```
mwoehlke-kitware, line 14: Ack, how/when did this sneak in and not get spotted before now?
```suggestion
```

arrows/serialize/json/load_save_metadata.cxx
mwoehlke-kitware: Is that documented somewhere? Anyway, if that's the case, I would suggest instead of `is_floating_point`, etc., having an enumeration of "allowed" types. That should allow us to avoid string comparisons (outside of `any_cast` itself, anyway) and also surprising failures if anyone violates the set of expected types.

I still submit this code ought to work (if it can, anyway; we can't do anything about someone using a custom type) if someone tries to use it on custom metadata items that don't follow the set of "allowed" types.

mwoehlke-kitware (approved):

mwoehlke-kitware:
Ack, I missed that all the CMake stuff also needed resorting; good catch, thanks!

@linus-sherrill, are your concerns all addressed?

arrows/serialize/json/load_save_metadata.cxx
mwoehlke-kitware: Is that documented somewhere? Anyway, if that's the case, I would suggest instead of `is_floating_point`, etc., having an enumeration of "allowed" types. That should allow us to avoid string comparisons (outside of `any_cast` itself, anyway) and also surprising failures if anyone violates the set of expected types.

I still submit this code ought to work (if it can, anyway; we can't do anything about someone using a custom type) if someone tries to use it on custom metadata items that don't follow the set of "allowed" types.

arrows/core/applets/dump_klv.cxx
russelldj: No idea, I went back a handful of commits and didn't see it introduced. Addressed.

arrows/serialize/json/load_save_metadata.cxx
mleotta: I'm not sure that it is documented, but it should be.  I would not object to formalizing this convention and possibly using an enum in a later PR, but that should not hold up this PR.  The point of the change in this PR is to add the "signed" case because were previously assuming all `int`s were unsigned, which was a problem when we added a field that needed negative integer numbers.

For the purposes of JSON we don't need or want to preserve C++ types.  The `any_converter` might be just what we need as a safety net here.  We want to just encode this data in string form.   When we read it back in we will put it back into the appropriate C++ types based on the tag, but the JSON should not encode C++ types.  In fact, it probably doesn't even need to encode "float" vs "integer" type names because I think the JSON type is supposed to be inferred from the string formatting.

arrows/serialize/json/register_algorithms.cxx
mleotta: I was not aware that we had a separate registration for serialization functions.  Independent of registering the algorithm `metadata_map_io` should we also register the serialization functions for metadata types with `sreg`?  I haven't looked enough at how this serialization stuff works, but it seems like we might want something like
```c++
sreg.register_algorithm< metadata >();
sreg.register_algorithm< metadata_map >();
```
to allow people to serialize those types outside of using the `metadata_map_io` algorithm.  Is that right, or am I misunderstanding?

mleotta:
Jenkins test this please.

arrows/serialize/json/load_save_metadata.cxx
mwoehlke-kitware: I refer you back to the first comment in this thread:

> I also want to get this PR landed, so we're thinking to address this as a follow-up.

:slightly_smiling_face:

I believe the only thing (possibly) holding this PR at present is hoping to hear back from @linus-sherrill that his concerns are adequately addressed.

> [we] were previously assuming all ints were unsigned, which was a problem when we added a field that needed negative integer numbers.

I expect it also ended badly when the `metadata` user expected the item to hold an `int` and was unable to extract the value *at all* (because it was actually a `uint64_t`). Whether or not the value was actually negative.

----

Okay, on closer inspection, it looks like we may not be able to handle custom items, period. I'll still argue (though I note it sounds like we may agree, here), that `is_signed`, etc. is perhaps not the best way to encode that an item should be one of a specific set of types. It may *work*, but it's rather obtuse.

>  The `any_converter` might be just what we need as a safety net here.

I'm not sure we need this; not in the [de]serialization, anyway. The JSON code has other ways of converting from JSON to the appropriate C++ type, and we find the appropriate type based on the tag. We just have a rather obtuse way (per previous paragraph in this comment) of extracting the appropriate type from the tag.

> the JSON should not encode C++ types

Actually, IIUC we are effectively only doing this as a convenience to the reader. Note: "These two items are included to increase readability of the serialized form and are not used when deserializing."

mleotta (approved):
I think we are ready to merge, but @russelldj I'd like your thoughts first on whether we should also be registering the `metadata_map` serializers directly (see inline comments).  Even if we want to do this, it could be added in a new PR, so its probably not worth holding this up.

mleotta:
Jenkins test this please

mleotta:
The CI is all passing.  I'm going to merge this and we can address any other needed improvements in future PRs.

arrows/serialize/json/register_algorithms.cxx
russelldj: I think we want to register `metadata` but `metadata_map` is not actually a class. The functionality to serialize a metadata map is contained in `metadata`. It's possible to register `metadata` and have this build with passing tests, but I'm not sure that means that it's actually working as expected. I'm a bit surprised that it wasn't registered before, because that class preceded this PR. My only concern is that `metadata` provides additional methods, (`serialize_meta`, `serialize_map`, and `deserialize_map`) on top of the `serialize` and `deserialize` provided by the other classes. My guess is this won't break anything but I think it's likely that this will only be able to serialize a vector of metadata items, e.g. the information from one metadata packet, since this is what `serialize`/`deserialize` handles.

mleotta: okay, it's not important for this PR (which I've already merged).  I was just wondering if that was something we overlooked for general JSON serialization purposed.  I don't know of enough about what registering serialization algorithms actually does in KWIVER.  It may make sense to register `metadata` if that's just a one line addition and doesn't break any tests.  Or we could just wait until someone finds a need for this.

PR #1244, opened 2021-03-12
Reverted change to MACOSX_RPATH

mleotta:
This was breaking the packaging of TeleSculptor.

I'm not sure what `${lib_subdir_path_to_root}${lib_dir_path_to_root}` was supposed to be doing here, but in my build it evaluated to the empty string.  The results was `@executable_path//lib` when it needs to be `@executable_path/../lib`.  Basically this is telling the linker where to find the libraries relative to the executable, which we assume is in `bin`.

Is is possible that we want `@executable_path/../${lib_subdir_path_to_root}${lib_dir_path_to_root}/${KWIVER_DEFAULT_LIBRARY_DIR}` or some other combination?  I don't understand what the use cases are when `KWIVER_DEFAULT_LIBRARY_DIR` is not just `lib`.

I'm not sure how to handle this fix in terms of releases.  I need to make a patch release of TeleSculptor against the patch release of KWIVER to pick up bug fixes in KWIVER, but I can't build against v1.5.1 with this bug.  I suppose we should have made a release candidate of KWIVER and tested packaging of downstream packages against it before release.  Unfortunately, I rarely run the packaging on MacOS.  I suppose what we really need to do is fix TeleSculptor CI, which has been broken for years, and then we might catch these things earlier.

### DIFF
--- a/CMake/utils/kwiver-utils-targets.cmake
+++ b/CMake/utils/kwiver-utils-targets.cmake
@@ -238,7 +238,7 @@ function(kwiver_add_library     name)
   if ( APPLE )
     set( props
       MACOSX_RPATH         TRUE
-      INSTALL_NAME_DIR     "@executable_path/${lib_subdir_path_to_root}${lib_dir_path_to_root}/${KWIVER_DEFAULT_LIBRARY_DIR}"
+      INSTALL_NAME_DIR     "@executable_path/../${KWIVER_DEFAULT_LIBRARY_DIR}"
       )
   else()
     if ( NOT no_version ) # optional versioning


mleotta:
I suppose what I can do is make TeleSculptor build against this commit of KWIVER, even if it isn't an official release, it's pretty close.

as6520:
`${lib_subdir_path_to_root}${lib_dir_path_to_root}` use the values defined in `library_dir` and `library_subdir` and determine how many times `../` needs to be applied to go back to the lib directory. In the original function, we were specifying rpaths with a single `../` from origin. However this did not work when plugins had their own sub folders. Thus I had a way to compute relative location of `lib` from `$origin` so that rpaths are set correctly. I don't think we are using `KWIVER_DEFAULT_LIBRARY_DIR`. I had added it so that it was variable that we could set rather than hard coded string. Not sure why the resolution is not working properly in the telescupltor. Could you paste a link to the call so that I can see where the problem might be? Since kwiver wheel does not support mac os this does not affect the build. So approving it.

as6520 (approved):

johnwparent:
@as6520 Will this be relevant down the line for a MacOS wheel build? If so, these changes are fine, but this should be tracked so it can be handled for MacOS wheels at a later point. Would the second solution suggested maybe be a promising alternative going forward?

@mleotta I agree, producing RCs for our Kwiver patch releases is a better policy, and future patch releases will receive a candidate, I'll add this to the release documentation I'm putting together.

johnwparent (approved):

mleotta:
This use case is not about path from the library to the root.  It is the path from the executable to the root.  This is used so that dynamic libraries can be found relative the executable.  This is not the same as `$origin` on linux where the path is relative to currently library or plugin.  So as long as your executable is in `bin`, the path should be `@executable_path/../` followed by the path from the root to where the libraries are stored.  I think in the example here `${library_dir}` defaults to `KWIVER_DEFAULT_LIBRARY_DIR`, which is just `lib`.  And `lib` has no `/` in the path so `${lib_subdir_path_to_root}` ends up empty.

mleotta:
For the record, this was error message before this fix.
```
CPack: - Run preinstall target for: TeleSculptor
CPack: - Install project: TeleSculptor []
CMake Error at /opt/local/share/cmake-3.19/Modules/BundleUtilities.cmake:454 (message):
  otool -l failed: 1


  /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/objdump:
  '@executable_path//lib/libmaptk.dylib': No such file or directory
```
This happened when packaging TeleSculptor on MacOS

PR #1253, opened 2021-03-24
Simple fix to metadata capability not getting set correctly

chetnieter:


arrows/ffmpeg/tests/test_video_input_ffmpeg.cxx
```diff
@@ -478,12 +482,19 @@ TEST_F(ffmpeg_video_input, no_sync_metadata)
...
+  EXPECT_TRUE( caps.capability( kwiver::vital::algo::video_input::HAS_METADATA ) );
+
   kwiver::vital::timestamp ts;
 
   unsigned int frame_num = 0;
   while (vif.next_frame(ts))
   {
     auto md_vect = vif.frame_metadata();
+
+    EXPECT_TRUE(md_vect.size() > 0)
```
russelldj, line 495: Just out of curiosity, this is true solely because of adding the video name to each frame?

russelldj (approved):
This fixes the issue for me. Approved pending clean CI.

arrows/ffmpeg/tests/test_video_input_ffmpeg.cxx
chetnieter: Yes - for this video the only metadata is the video filename. There is no actual metadata stored in the video file.

PR #1255, opened 2021-03-24
[VIAME] Add intersection and empty checks to refiners

russelldj:
This fixes failures in the refiner pipelines caused by detections which extended past the border of the image by doing the following:
* Check that the input image isn't null
* Intersect the detection rectangle with the image boundary
* If the intersection in null, simply return the original detection. Otherwise, process as normal. 
 
It's possible there are better ways to implement this.

### DIFF
--- a/arrows/ocv/refine_detections_grabcut.cxx
+++ b/arrows/ocv/refine_detections_grabcut.cxx
@@ -111,6 +111,7 @@ refine_detections_grabcut
   using ic = ocv::image_container;
   cv::Mat img = ic::vital_to_ocv( image_data->get_image(), ic::BGR_COLOR );
   cv::Rect img_rect( 0, 0, img.cols, img.rows );
+  bounding_box<double> vital_img_rect( 0, 0, img.cols, img.rows );
 
   if( !detections )
   {
@@ -120,7 +121,7 @@ refine_detections_grabcut
   auto result = std::make_shared< vital::detected_object_set >();
   for( auto const& det : *detections )
   {
-    auto&& bbox = det->bounding_box();
+    auto&& bbox = intersection( det->bounding_box(), vital_img_rect );
     // Determine context crop and translate the bounding box
     auto cbbox = vital::scale_about_center( bbox, d_->context_scale_factor );
     auto ctx_rect = bbox_to_mask_rect( cbbox ) & img_rect;
--- a/arrows/ocv/refine_detections_grabcut.cxx
+++ b/arrows/ocv/refine_detections_grabcut.cxx
@@ -111,7 +111,7 @@ refine_detections_grabcut
   using ic = ocv::image_container;
   cv::Mat img = ic::vital_to_ocv( image_data->get_image(), ic::BGR_COLOR );
   cv::Rect img_rect( 0, 0, img.cols, img.rows );
-  bounding_box<double> vital_img_rect( 0, 0, img.cols, img.rows );
+  bounding_box< double > vital_img_rect( 0, 0, img.cols, img.rows );
 
   if( !detections )
   {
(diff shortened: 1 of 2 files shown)


russelldj:
See [here](https://drive.google.com/file/d/1dHux-MDSTm5eGacheXarGycUrIoovL5s/view?usp=sharing) for a zip containing an MVE. Note that as far as I can tell, the only requirement is a bounding box extending beyond the image border, e.g. the `-11.5` entry.

russelldj:
Interestingly, grabcut only fails if the annotation is fully outside the image which is much less common but I believe I have seen happen in practice.

russelldj:
And watershed error:
```
OpenCV Error: Assertion failed (0 <= roi.x && 0 <= roi.width && roi.x + roi.width <= m.cols && 0 <= roi.y && 0 <= roi.height && roi.y + roi.height <= m.rows) in Mat, file /home/local/KHQ/david.russell/dev/VIAME/build_debug/build/src/fletch-build/build/src/OpenCV/modules/core/src/matrix.cpp, line 538
terminate called after throwing an instance of 'cv::Exception'
```
Backtrace:
```
#9  0x00007fffebb4aa4d in cv::Mat::operator() (this=0x7ffe540017b0, roi=...) at /home/local/KHQ/david.russell/dev/VIAME/build_debug/install/include/opencv2/core/mat.inl.hpp:830
#10 0x00007fffebb715f7 in kwiver::arrows::ocv::refine_detections_watershed::refine (this=0x555556693680, image_data=..., detections=...)
    at /home/local/KHQ/david.russell/dev/VIAME/src/packages/kwiver/arrows/ocv/refine_detections_watershed.cxx:158
#11 0x00007fffe4c74023 in kwiver::refine_detections_process::_step (this=0x55555669e460) at /home/local/KHQ/david.russell/dev/VIAME/src/packages/kwiver/sprokit/processes/core/refine_detections_process.cxx:122
#12 0x00007ffff4e6fde1 in sprokit::process::step (this=0x55555669e460) at /home/local/KHQ/david.russell/dev/VIAME/src/packages/kwiver/sprokit/src/sprokit/pipeline/process.cxx:387
#13 0x00007fffe32094df in sprokit::thread_per_process_scheduler::priv::run_process (this=0x555556004e20, process=...)
```

neal-siekierski:
Thanks for looking at this and providing the example. You're right that this is very easy to reproduce.

I've had the time to take a closer look and push up a more targeted fix [here](https://github.com/Kitware/kwiver/compare/viame/master...VIAME:dev/ocv-refine-detections-oob). There were two main issues:
- For watershed, the seed / initial mask images weren't being cropped before being used with the marker image
- For GrabCut, `cv::grabCut` was being called when the cropped context region lacked foreground or background (including when it was empty). This skips calling it.

That branch doesn't address the other issues like gracefully handling missing input imagery.

What the "right" fix is might lie in between what this branch currently has and what's in the branch I pushed up and depends on what "should" happen for partially / fully OOB boxes. The goal of the original code, which I've tried to preserve in the new branch, is to have control flow that's as similar as possible no matter the state of the input detection &ndash; in, partially in, or full out of bounds; existing mask or no existing mask. That flow is:
- Get a normalized input mask for the detection (based on the existing mask if there is one and configuration allows, otherwise a scaled rectangle as configured)
- Update the part of the mask that's within the image using the chosen segmentation method (the rest of the mask stays as is)
- Update the detection to reference the new mask

The advantages are that it's conceptually simple (at least to me, and the wrangling of the cropping aside) and more importantly that it guarantees that every detection has a mask image that's the "right" shape (as determined by `bbox_to_mask_rect`). The disadvantages are that the partially OOB newly minted scaled rectangles arguably leave an artifact when detections are mostly / fully out of bounds and that it's unclear if we even want masks for such detections.

neal-siekierski:
Sounds good. I created #1266. To even things out I added the check for the image being null. If it works when you test it later feel free to merge.

neal-siekierski:
With #1266 merged and, as I understand it, also working for you, should this PR be closed?

russelldj:
Yes, #1266 has been tested on the multiple years of sea lion data for both methods, including known prior failures.

PR #1261, opened 2021-03-27
dev/resection camera API

borovik135:
camera resection implementation based on OpenCV calibrateCamera method


arrows/ocv/resection_camera.h
```diff
@@ -0,0 +1,73 @@
...
+    std::vector<bool> &inliers, ///< [out] inlier flags for each point, the value is true if this pair is an inlier to the estimate
+    kwiver::vital::camera_intrinsics_sptr cal ///< [in] initial guess on intrinsic parameters of the camera
+  ) const override;
+
+  /// Estimate camera parameters for a frame from landmarks and tracks.
+  /// This is a convenience function, constructing the pts2d and pts3d from a frame, landmarks, tracks,
+  /// and then call resection() with the point correspondences.
+  /// \return estimated camera parameters
+  virtual kwiver::vital::camera_perspective_sptr
+  resection (kwiver::vital::frame_id_t frmID, ///< [in] frame number for which to estimate a camera
```
mleotta, line 57: See if you can just use `using vital::algo::resection_camera::resection;` here instead of defining this one again.

arrows/ocv/resection_camera.h
```diff
@@ -0,0 +1,73 @@
...
+  /// Set this algorithm's properties via a config block.
+  virtual void set_configuration(vital::config_block_sptr config);
+
+  /// Check that the algorithm's configuration config_block is valid.
+  virtual bool check_configuration(vital::config_block_sptr config) const;
+
+  /// Estimate the camera's pose and projection matrix from the 3D points and their corresponding projections.
+  /// \return estimated camera parameters
+  virtual kwiver::vital::camera_perspective_sptr
+  resection (std::vector<kwiver::vital::vector_2d> const &pts2d, ///< [in] 2d projections of pts3d in the same order as pts3d
```
mleotta, line 46: This is interesting doxygen style.  I haven't seen this before.  The problem is that it makes line lengths quite long.  The KWIVER styles is to put these parameter descriptions in the block above using `\\\ \param pts2d [in] description`

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,198 @@
...
+namespace kwiver {
+namespace arrows {
+namespace ocv
+{
+
+/// private implementation
+struct resection_camera::priv
+{
+  priv()
+    : confidence_threshold(0.99)
```
mleotta, line 31: It looks like these parameter are left over from PNP.  We can remove them.  Later you may want to add parameters relating to which camera parameters to estimate, but these are not needed in the first pass.

When we do add parameters for controlling which camera parameters to estimate, we might want to borrow some of that configuration logic from the Ceres solver arrow, which does the same thing.

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,111 @@
...
+    auto const fts = dynamic_pointer_cast<feature_track_state>(*track->find(frmID));
+    pts_projs.push_back(fts->feature->loc());
+  }
+
+  // camera pose from the points and their projections
+  resection_camera res_cam;
+  auto cams = cameras->cameras();
+  auto cam = dynamic_pointer_cast<camera_perspective>(cams[frmID]);
+  vector<bool> inliers;
+  auto est_cam = res_cam.resection(pts_projs, pts_3d, inliers, cam->intrinsics());
```
mleotta, line 47: We should probably have a test that calls the other API that works directly on tracks and landmarks to make sure that works as well.

Also we probably want a test that doesn't use the correct intrinsics as input to make sure we can recover the intrinsics as well.

config/mvg_initialize_cameras_landmarks.conf
```diff
@@ -47,6 +47,7 @@ block mvg
   essential_mat_estimator:type = vxl
 
   estimate_pnp:type = ocv
+  resection_camera:type = ocv
```
mleotta, line 50: we probably don't need this change yet.  The mvg cameras and landmarks initialization doesn't yet use this new resection API.

python/kwiver/vital/algo/trampoline/estimate_pnp_trampoline.txx
```diff
@@ -29,10 +29,10 @@
  */
 
 /**
- * \file estimate_homography_trampoline.txx
+ * \file estimate_pnp_trampoline.txx
```
mleotta, line 32: This looks like an error in the documentation for PNP python bindings.  You could push just the changes to this file in a separate PR if you'd like.  That would be merged quickly.  Generally bug fixes like this should be pushed as separate PR's and rebased onto the `release` branch and then targeted for merging into the `release` branch.  This way they land sooner and are back ported to fix the last release.

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+  /// \return name of this algorithm
+  static std::string static_type_name() { return "resection_camera"; }
+
+  /// Estimate camera parameters from 3D points and their corresponding projections.
+  /// \return estimated camera parameters
+  virtual
+  kwiver::vital::camera_perspective_sptr
+  resection(
+    std::vector<kwiver::vital::vector_2d> const & pts2d, ///< [in] 2d projections of pts3d in the same order as pts3d
```
mleotta, line 37: Again, this is not the doxygen style used elsewhere in KWIVER.  We should be consistent.

mleotta (changes requested):
Here is some initial feedback.

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,198 @@
...
+    vector<bool>& inliers,
+    kwiver::vital::camera_intrinsics_sptr cal
+) const
+{
+  if (cal==nullptr)
+  {
+    LOG_ERROR(d_->m_logger, "camera calibration guess should not be null in this implementation");
+    return vital::camera_perspective_sptr();
+  }
+  const unsigned
```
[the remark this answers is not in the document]
borovik135: OK, changed assuming that makes line commenting easier.

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,198 @@
...
+    minCnt = 3,
+    pts2cnt = pts2d.size(),
+    pts3cnt = pts3d.size();
+  if (pts2cnt < minCnt || pts3cnt < minCnt)
+  {
+    LOG_ERROR(d_->m_logger, "not enough points to resection camera");
+    return vital::camera_perspective_sptr();
+  }
+  if (pts2cnt != pts3cnt)
```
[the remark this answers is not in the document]
borovik135: OK, unsure why one-liners deserve scoping, but will follow the style for consistency.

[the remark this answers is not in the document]
mleotta: It's a little more verbose, but it can remove ambiguities and make code more readable.  Here is some discussion on the topic:
https://stackoverflow.com/questions/12193170/whats-the-purpose-of-using-braces-i-e-for-a-single-line-if-or-loop

arrows/ocv/resection_camera.cxx
borovik135: changed to
    : reproj_accuracy(DBL_EPSILON)
    , max_iterations(30)
as specified in cv::calibrateCamera()

arrows/tests/test_resection_camera.h
borovik135: we currently have two tests, one calling resection with points, while another one with tracks; I am working on a test with null intrinsics

config/mvg_initialize_cameras_landmarks.conf
borovik135: OK, removed that line for now.

python/kwiver/vital/algo/trampoline/estimate_pnp_trampoline.txx
borovik135: reverting this back to the erroneous line seems weird, but I can see the merge to release logic

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,198 @@
...
+  vector<Mat> vrvec, vtvec;
+
+  vector<vector<Point3f>> objPts = {Xs};
+  vector<vector<Point2f>> imgPts = {projs};
+  Size imgSize(cal->image_width(), cal->image_height());
+  int flags = CALIB_USE_INTRINSIC_GUESS;
+  const auto err = calibrateCamera(objPts, imgPts, imgSize, cv_K, dist_coeffs, vrvec, vtvec, flags);
+  const auto reproj_error = 4.;
+  if (err>reproj_error)
+    LOG_WARN(d_->m_logger, "estimated re-projection error " << err << " > " << reproj_error << " expected re-projection error");
```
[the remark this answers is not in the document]
borovik135: scoped and multi-lined

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,198 @@
...
+
+  Mat rvec=vrvec[0], tvec=vtvec[0];
+  vector<Point2f> prjPts; projectPoints(Xs, rvec, tvec, cv_K, dist_coeffs, prjPts);
```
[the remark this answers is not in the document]
borovik135: now in separate lines

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,198 @@
...
+  cv2eigen(rvec, rvec_eig);
+  cv2eigen(tvec, tvec_eig);
+  cv2eigen(cv_K, K);
+  rotation_d rot(rvec_eig);
+  res_cam->set_rotation(rot);
+  res_cam->set_translation(tvec_eig);
+  cal.reset(new simple_camera_intrinsics(K, dist_eig));
+  res_cam->set_intrinsics(cal);
+
+  if (!isfinite(res_cam->center().x()))
```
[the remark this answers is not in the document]
borovik135: now checking for x,y,z

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,198 @@
...
+              tvec.at<double>(1) << " " << tvec.at<double>(2));
+    LOG_DEBUG(d_->m_logger, "rotation angle " << res_cam->rotation().angle());
+    LOG_WARN(d_->m_logger, "non-finite camera center found");
+    return vital::camera_perspective_sptr();
+  }
+  return dynamic_pointer_cast<vital::camera_perspective>(res_cam);
+}
+
+kwiver::vital::camera_perspective_sptr
+resection_camera::resection(kwiver::vital::frame_id_t frame,
```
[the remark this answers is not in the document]
borovik135: done; thanks for the hint!

arrows/ocv/resection_camera.h
borovik135: reverted to the convention, although I still find it more convenient to have parameter descriptions next to the arguments themselves: that saves on name and order syncing

python/kwiver/vital/algo/trampoline/estimate_pnp_trampoline.txx
mleotta: You don't need to revert this change here, although you could.  I'm just suggesting factoring it out so that it also stands alone an can be fast-tracked into merging in the release branch.  If you have this change on a separate commit you could just cherry-pick that commit onto a new branch off of the release branch and push that as a new PR.

arrows/tests/test_resection_camera.h
borovik135: I changed the interface for the convenience method, requesting the image dimensions instead the initial calibration structure.

borovik135: added one more virtual resection(frmID, landmarks, tracks, cam.intrinsics), and a test of that

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,151 @@
...
+  landmark_map_sptr landmarks = kwiver::testing::init_landmarks(128);
+  landmarks = kwiver::testing::noisy_landmarks(landmarks, 1.0);
+
+  // camera sequence (elliptical path)
+  camera_map_sptr cameras = kwiver::testing::camera_seq();
+
+  // tracks from projections
+  track_set_sptr tracks = projected_tracks(landmarks, cameras);
+
+  const frame_id_t frmID = 0;
```
mwoehlke-kitware, line 28:
```suggestion
  constexpr frame_id_t test_frame = 0;
```
(Likewise in the other test.)

arrows/tests/test_resection_camera.h
mwoehlke-kitware: Maybe we should provide both? (Oh, look, we *do*.) I had similar thoughts ("why are we not testing conversion via the convenience method").

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,151 @@
...
+  resection_camera res_cam;
+  auto cams = cameras->cameras();
+  auto cam = dynamic_pointer_cast<camera_perspective>(cams[frmID]);
+  vector<bool> inliers;
+  auto est_cam = res_cam.resection(pts_projs, pts_3d, inliers, cam->intrinsics());
+
+  // true and computed camera poses
+  const auto
+    &camR = cam->rotation(),
+    &estR = est_cam->rotation();
```
mwoehlke-kitware, line 52:
```suggestion
  auto const& camR = cam->rotation();
  auto const& estR = est_cam->rotation();
```
...the repeated `&` is a great example of *why* "one declaration per statement" is so widely preferred.

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,151 @@
...
+
+using namespace kwiver::vital;
+using namespace kwiver::arrows::mvg;
+using namespace std;
+
+// ----------------------------------------------------------------------------
+// Test camera resection with ideal points.
+TEST(resection_camera, ideal_points)
+{
+  // landmarks at random locations
```
mwoehlke-kitware, line 18:
```suggestion
  // Landmarks at random locations
```

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,151 @@
...
+  auto est_cam = res_cam.resection(pts_projs, pts_3d, inliers, cam->intrinsics());
+
+  // true and computed camera poses
+  const auto
+    &camR = cam->rotation(),
+    &estR = est_cam->rotation();
+  cout << "cam R:\n" << camR.matrix() << endl
+  << "est R:\n" << estR.matrix() << endl
+  << "cam C = " << cam->center().transpose() << endl
+  << "est C = " << est_cam->center().transpose() << endl;
```
mwoehlke-kitware, line 56:
```suggestion
  cout << "cam R:\n" << camR.matrix() << endl
       << "est R:\n" << estR.matrix() << endl
       << "cam C = " << cam->center().transpose() << endl
       << "est C = " << est_cam->center().transpose() << endl;
```
(uncrustify is *supposed* to not botch this, but I've seen it have trouble indenting around streaming operators.)

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,151 @@
...
+
+  auto R_err = camR.inverse()*estR;
+  cout << "rotation error = " << R_err.angle()*180/pi << " degrees" << endl;
```
mwoehlke-kitware, line 59: Use `vital::rad_to_deg` here?

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,151 @@
...
+{
+  // landmarks at random locations
+  landmark_map_sptr landmarks = kwiver::testing::init_landmarks(128);
+  landmarks = kwiver::testing::noisy_landmarks(landmarks, 1.0);
+
+  // camera sequence (elliptical path)
+  camera_map_sptr cameras = kwiver::testing::camera_seq();
+
+  // tracks from the projections
+  auto tracks = dynamic_pointer_cast<feature_track_set>(projected_tracks(landmarks, cameras));
```
mwoehlke-kitware, line 81: I am having trouble finding where `projected_tracks` is defined (`arrows/mvg/projected_track_set.h`? That seems unlikely?), but I am surprised it does not already return a `feature_track_set_sptr`?

arrows/tests/test_resection_camera.h
```diff
@@ -0,0 +1,151 @@
...
+
+  // random noise to track image locations
```
mwoehlke-kitware, line 83:
```suggestion
  // Add some random noise to the track image locations
```

vital/algo/resection_camera.cxx
```diff
@@ -0,0 +1,86 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/**
+ * \file
+ * \brief resection_camera instantiation
+ */
```
mwoehlke-kitware, line 8:
```suggestion
/// \file
/// \brief resection_camera instantiation
```

vital/algo/resection_camera.cxx
```diff
@@ -0,0 +1,86 @@
...
+
+using namespace std;
+
+namespace kwiver {
+
+namespace vital {
+
+namespace algo {
+
+resection_camera
```
mwoehlke-kitware, line 25:
```suggestion
// ----------------------------------------------------------------------------
resection_camera
```
...and throughout.

vital/algo/resection_camera.cxx
```diff
@@ -0,0 +1,86 @@
...
+::resection_camera()
+{
+  attach_logger( "algo.resection_camera" );
+}
+
+camera_perspective_sptr
+resection_camera
+::resection( frame_id_t frmID,
```
mwoehlke-kitware, line 33:
```suggestion
::resection( frame_id_t frame,
```
...or `frame_id`, but in any case, please choose something more in line with prevailing naming conventions.

vital/algo/resection_camera.cxx
```diff
@@ -0,0 +1,86 @@
...
+  vector< vector_3d > pts_3d; // world points
+  vector< vector_2d > pts_projs; // corresponding image points
+  for( auto const& track : tracks->tracks() )
+  {
+    auto lm_id = track->id();
+    auto lm_it = landmarks->landmarks().find( lm_id );
+    pts_3d.push_back( lm_it->second->loc() );
+    auto const fts =
+      dynamic_pointer_cast< feature_track_state >( *track->find( frmID ) );
+    pts_projs.push_back( fts->feature->loc() );
```
mwoehlke-kitware, line 75: I think this will SEGV if the track has no state at the requested frame? (Note that this is quite likely to happen in actual use, as I expect callers will have a set of feature tracks covering multiple cameras that will not be filtered for only those feature tracks with points on the specified camera.)

While we're at it, we probably shouldn't assume a one-to-one correspondence between feature tracks and landmarks, either:
```suggestion
    auto const fti = track->find( frmID );
    if( fti != track->end() )
    {
      auto const fts = dynamic_pointer_cast< feature_track_state >( *fti );
      auto lmi = landmarks->landmarks().find( track->id() );
      if( lmi != landmarks->landmarks().end() )
      {
        pts_3d.push_back( lmi->second->loc() );
        pts_projs.push_back( fts->feature->loc() );
      }
    }
```

vital/algo/resection_camera.cxx
```diff
@@ -0,0 +1,86 @@
...
+  }
```
mwoehlke-kitware, line 48: Instead of duplicating this code, please refactor to call the overload that takes tracks, landmarks and intrinsics.

vital/algo/resection_camera.cxx
```diff
@@ -0,0 +1,86 @@
...
+  // resection camera using point correspondences and initial calibration guess
+  vector< bool > inliers;
+  return resection( pts_projs, pts_3d, inliers, cal );
+}
+
+} // end algo
+
+} // end vital
+
+} // end kwiver
```
mwoehlke-kitware, line 86:
```suggestion
} // namespace algo

} // namespace vital

} // namespace kwiver
```

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/**
+ * \file
+ * \brief resection_camera algorithm definition
+ */
```
mwoehlke-kitware, line 8:
```suggestion
/// \file
/// \brief resection_camera algorithm definition
```

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+/// Shared pointer type of base resection_camera algorithm definition class
+typedef std::shared_ptr< resection_camera > resection_camera_sptr;
+
+} // namespace algo
+
+} // namespace vital
+
+}     // end namespace
+
+#endif // VITAL_ALGO_RESECTION_CAMERA_H_
```
mwoehlke-kitware, line 97:
```suggestion
} // namespace kwiver

#endif
```

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+    kwiver::vital::landmark_map_sptr landmarks,
+    kwiver::vital::feature_track_set_sptr tracks,
+    kwiver::vital::camera_intrinsics_sptr cal ) const;
+
+protected:
+  resection_camera();
+};
+
+/// Shared pointer type of base resection_camera algorithm definition class
+typedef std::shared_ptr< resection_camera > resection_camera_sptr;
```
mwoehlke-kitware, line 89:
```suggestion
/// Shared pointer type of base resection_camera algorithm definition class.
using resection_camera_sptr = std::shared_ptr< resection_camera >;
```

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+  kwiver::vital::camera_perspective_sptr
+  resection(
+    kwiver::vital::frame_id_t frmID,
+    kwiver::vital::landmark_map_sptr landmarks,
+    kwiver::vital::feature_track_set_sptr tracks,
+    unsigned width, unsigned height ) const;
+
+  /// Estimate camera parameters for a frame from landmarks and tracks.
+  /// This is a convenience function, callin internally
+  /// resection(pts2d, pts3d, ...) with the recoverd point correspondences.
```
mwoehlke-kitware, line 70:
```suggestion
  /// Estimate camera parameters for a frame from landmarks and tracks.
  ///
  /// This is a convenience overload; the default implementation calls
  /// resection(pts2d, pts3d, ...) with the recovered point correspondences.
  ///
```
...although even better would be to say something like "extracts the raw point locations and calls <link to appropriate overload>". (I'm not sure if there's an easy way to write that link, however.)

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+  resection(
+    std::vector< kwiver::vital::vector_2d > const& pts2d,
+    std::vector< kwiver::vital::vector_3d > const& pts3d,
+    std::vector< bool >& inliers,
+    kwiver::vital::camera_intrinsics_sptr cal ) const = 0;
+
+  /// Estimate camera parameters for a frame from landmarks and tracks.
+  /// This is a convenience function, callin internally
+  /// resection(pts2d, pts3d, ...) with the recoverd point correspondences.
+  /// \param [in] frmID frame number for which to estimate a camera
```
mwoehlke-kitware, line 54: (See naming note in `.cxx`.)

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \param [in] landmarks 3D landmarks locations to constrain camera
+  /// \param [in] tracks 2D feature tracks in image coordinates
+  /// \param [in] width image matrix column count
+  /// \param [in] height image matrix row count
```
mwoehlke-kitware, line 58: "{row,column} count" is rather unexpected here. Strictly speaking, the expected values are twice the x/y value of the camera center point, as given in the same coordinate system as the feature track locations. Practically speaking, of course, that's something like "{width,height} of camera image".

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+    std::vector< kwiver::vital::vector_2d > const& pts2d,
+    std::vector< kwiver::vital::vector_3d > const& pts3d,
+    std::vector< bool >& inliers,
+    kwiver::vital::camera_intrinsics_sptr cal ) const = 0;
+
+  /// Estimate camera parameters for a frame from landmarks and tracks.
+  /// This is a convenience function, callin internally
+  /// resection(pts2d, pts3d, ...) with the recoverd point correspondences.
+  /// \param [in] frmID frame number for which to estimate a camera
+  /// \param [in] landmarks 3D landmarks locations to constrain camera
```
mwoehlke-kitware, line 55: Possibly you meant:
```suggestion
  /// \param [in] landmarks 3D landmark locations to constrain camera
```

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \param [in] pts3d 3d points in a 1-1 correspondence with pts2d
+  /// \param [out] inliers inlier flags for the point pairs
+  /// \param [in] cal initial guess on intrinsic parameters of the camera
+  /// \return estimated camera parameters
+  virtual
+  kwiver::vital::camera_perspective_sptr
+  resection(
+    std::vector< kwiver::vital::vector_2d > const& pts2d,
+    std::vector< kwiver::vital::vector_3d > const& pts3d,
+    std::vector< bool >& inliers,
```
mwoehlke-kitware, line 48: You might consider moving this so it is not mixed in with the input parameters.

vital/algo/resection_camera.h
```diff
@@ -0,0 +1,97 @@
...
+{
+public:
+  /// \return name of this algorithm
+  static std::string
+  static_type_name() { return "resection_camera"; }
+
+  /// Estimate camera parameters from 3D points and their corresponding
+  /// projections.
+  /// \param [in] pts2d 2d projections of pts3d in the same order as pts3d
+  /// \param [in] pts3d 3d points in a 1-1 correspondence with pts2d
```
mwoehlke-kitware, line 39:
```suggestion
  /// \param [in] image_points 2D
  ///   projections of \p world_points in camera image space
  /// \param [in] world_points
  ///   points in world space corresponding to the \p image_points
```
(...and consider propagating the name change. Alternately, `camera_points`; no particular opinion which is better.)

I dropped the correspondence bit. Possibly you could add that back, but it felt a little awkward

mwoehlke-kitware (changes requested):
Please at least fix the SEGV 🙂.

borovik135:
> SEGV

I am not sure where the segmentation violation is happening, and thus unable to fix that. The test on camera resection runs and passes OK, i.e.
$ tests/bin/test-ocv-resection_camera
...
[ RUN      ] resection_camera.noisy_points_cal
cam R:
   -0.150843     0.988558 -1.66533e-16
     0.46268    0.0705999     -0.88371
   -0.873598    -0.133302    -0.468035
est R:
  -0.150356    0.988632 0.000114703
    0.46263   0.0704618   -0.883747
  -0.873708   -0.132824   -0.467965
cam C =   3.9197 0.598104      2.1
est C =  3.91452 0.597655  2.09986
rotation error = 0.0293237 degrees
[       OK ] resection_camera.noisy_points_cal (5 ms)
[----------] 4 tests from resection_camera (3025 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (3025 ms total)
[  PASSED  ] 4 tests.

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/**
+ * \file
+ * \brief OCV resection_camera algorithm implementation
+ */
```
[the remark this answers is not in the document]
borovik135: done

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
...
+
+#include "resection_camera.h"
+
+#include <cmath>
+#include <opencv2/calib3d/calib3d.hpp>
+#include <opencv2/core/eigen.hpp>
+
+#include "camera_intrinsics.h"
```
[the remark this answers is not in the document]
borovik135: removed #include <cmath> as unneeded, moved more specific headers upwards

mwoehlke-kitware:
> I am not sure where the segmentation violation is happening [...] The test on camera resection runs and passes OK

That is because the test is providing neat, "sanitized" data. Try again with feature tracks that don't contain features for the requested frame, and contain features with no corresponding landmark. The latter "might" never happen in practice, though we should still guard against it. The former I am almost certain *will* happen. Basically, both of those `find`s can (*and will*, sometimes, but not in the test as currently set up) return invalid iterators, which your current code will happily try to dereference... which Will Not End Well™ 🙂.

In fact, it might not hurt to add a test, maybe without fuzzing the point positions, that pseudo-randomly (for a unit test, you may want to use a constant seed so the test is deterministic) drops some percentage of the landmarks (25%?), some percentage of the feature tracks (15%? 25%?), and some percentage of the feature points *within* those tracks (50%?). I'll bet you *that* test crashes.

I haven't *actually* seen it SEGV, I'm just confident that it *will*.

mleotta:
As a point of reference, we do have other test for bundle adjustment that take the full set of test data and extract subsets to make sure the algorithms work with sparse subsets of data.  For example, see
https://github.com/Kitware/kwiver/blob/master/arrows/tests/test_bundle_adjust.h#L219
and
https://github.com/Kitware/kwiver/blob/master/arrows/tests/test_bundle_adjust.h#L269
you can use these as an example

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
...
+
+using namespace std;
+using namespace cv;
+using namespace kwiver::vital;
```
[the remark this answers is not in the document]
borovik135: That is my inclination towards the Occam's razor, if you will. Rational being: even short namespaces prefixes add to a long expression, and we all have limited time to look at the code, hence the preference for shorter, clearly written, transparent expressions, and usage of qualified names where it clearly makes a difference. Things like cout, cerr, endl most certainly come from std, for instance. I removed using namespace kwiver::vital; keeping the others for now.

mleotta:
Also, please note that the ghostflow CI is failing due to several commits adding white space at the end of line or end of file.  Please set your editor to remove or highlight trailing whitespace or use a command line tool to remove it before checking in code.  If you run `git diff` before commiting code, git should also highlight this extra whitespace.  For example, I run the following to remove trailing whitespace if I find any:
```bash
#/bin/bash
find $1 -type f -exec sed -i 's/[[:space:]]*$//' '{}' ';'
```

It's not enough to simply remove white space on new commit to fix this.  Ghostflow checks all commit.  So you will need to interactively rebase an fix each commit it is complaining about.  However, it might be easier to just squash all the commit into one and fix it on that new commit, unless there is history in your changes that should be preserved.

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
...
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace ocv {
+
+/// private implementation
```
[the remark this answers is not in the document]
borovik135: done

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
...
+  vital::logger_handle_t m_logger;
+
+  priv() : m_logger( vital::get_logger( "arrows.ocv.resection_camera" ) )
+  {
+  }
+};
+
+resection_camera
+::resection_camera()
+  : d_( new priv )
```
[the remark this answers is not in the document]
borovik135: done, although I thought that member initialization would use (); not sure how {} would be better in this case.

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
...
+    LOG_ERROR( d_->m_logger, "camera calibration guess should not be null" );
+    return vital::camera_perspective_sptr();
+  }
+
+  const unsigned minCnt = 3;
+  const unsigned pts2cnt = pts2d.size();
+  const unsigned pts3cnt = pts3d.size();
+  if( pts2cnt < minCnt || pts3cnt < minCnt )
+  {
+    LOG_ERROR( d_->m_logger, "not enough points to resection camera" );
```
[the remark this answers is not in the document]
borovik135: not a sentence, so ignoring the capitalization for now

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
...
+    LOG_WARN( d_->m_logger,
+              "counts of 3D points and projections do not match" );
+  }
+
+  vector< Point2f > projs;
+  vector< Point3f > Xs;
+  for( const auto& p : pts2d )
+  {
+    projs.push_back( Point2f( static_cast< float >( p.x() ),
+                              static_cast< float >( p.y() ) ) );
```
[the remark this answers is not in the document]
borovik135: done

arrows/ocv/resection_camera.cxx
```diff
@@ -0,0 +1,221 @@
...
+  for( const auto& X : pts3d )
+  {
+    Xs.push_back( Point3f( static_cast< float >( X.x() ),
+                           static_cast< float >( X.y() ),
+                           static_cast< float >( X.z() ) ) );
+  }
+
+  matrix_3x3d K = cal->as_matrix();
+  Mat cv_K;
+  eigen2cv( K, cv_K );
```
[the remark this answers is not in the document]
borovik135: ignoring this change since we need matrix_3x3d K later in the procedure

arrows/ocv/tests/test_resection_camera.cxx
```diff
@@ -0,0 +1,36 @@
...
+using namespace kwiver::arrows;
+using namespace std;
+
+using ocv::resection_camera;
+
+static constexpr double
+  ideal_rotation_tolerance = 1e-6,
+  ideal_center_tolerance = 1e-6,
+  noisy_rotation_tolerance = 0.008,
+  noisy_center_tolerance = 0.05;
```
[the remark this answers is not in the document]
borovik135: noted, but leaving as is for now

arrows/tests/test_resection_camera.h
borovik135: done, tnx for the hint

arrows/tests/test_resection_camera.h
borovik135: removed the extra pointer cast

vital/algo/resection_camera.cxx
mwoehlke-kitware: Another bug we both missed, but the compiler caught... `landmarks->landmarks()` is a temporary; it needs to be assigned to something.

borovik135:
assigning it now to a variable within a point getting function

vital/algo/resection_camera.cxx
borovik135: addressed

vital/algo/resection_camera.cxx
borovik135: addressed

vital/algo/resection_camera.h
borovik135: addressed

vital/algo/resection_camera.h
borovik135: using size in the x/y dimension in pixels

vital/algo/resection_camera.h
borovik135: done

vital/algo/resection_camera.h
borovik135: not sure where to move them

vital/algo/resection_camera.h
mwoehlke-kitware: Ahem: ...felt a little awkward as part of the parameter documentation. It might be a good candidate for a `\note`. (Not sure what happened, possibly I got distracted and forgot to finish writing the comment. Sorry.)

vital/algo/resection_camera.cxx
```diff
@@ -0,0 +1,85 @@
...
+             landmark_map_sptr landmarks,
+             feature_track_set_sptr tracks,
+             kwiver::vital::camera_intrinsics_sptr cal ) const
+{
+  auto world_points = vector< vector_3d >{};
+  auto camera_points = vector< vector_2d >{};
+
+  auto const& real_landmarks = landmarks->landmarks();
+
+  for( auto const& track : tracks->tracks() )
```
mleotta, line 61: `feature_track_set` has a helper to make this more efficient. `tracks->frame_feature_track_states(frame_id)`

```c++
for( auto const& fts : tracks->frame_feature_track_states(frame_id) )
{
  auto lmi = real_landmarks.find( fts->track()->id() );
  ...
```

mleotta: This is not critical, just an optimization.  We can revise this later if it would help get this branch landed sooner.

vital/algo/resection_camera.h
mleotta: Let's hold off on further naming/style changes until we get a version of this landed.  I don't want to keep holding it up.  We can revise in a future PR if desired.

vital/algo/resection_camera.cxx
borovik135: done

borovik135:
I think I am done with the changes. Let's review and merge, I suppose.

borovik135:
I have created an alternative PR #1269 with a much shorter commit/revision history; shall we kill this PR then?

mleotta:
Yes, that's fine.  closing in favor of #1269

PR #1267, opened 2021-04-05
Fixed error in doxygen file name

mleotta:


mleotta:
@borovik135 This is what I meant about applying this bug fix change as a separate PR for the release branch.

borovik135 (approved):
I have recently learned that doxygen allows to leave the \file operand blank, in which case the containing file name is supplied during doxy compilation.

borovik135:
thanks for taking care of this, it looks good to me, tnx!

mleotta:
@dstoup @Cookt2 Is the Window CI machine down.  We have a number of branches that are waiting on results from the Windows build.

mleotta:
This only changes doxygen strings so I don't think I need to rerun the CI.  Merging.

PR #1272, opened 2021-04-12
Fixup SKBUILD linking order and RPATH

johnwparent:
Reorders FFMPEG dependency linking, fixes CMake bug when SKBUILD flag is set, when SKBUILD is set, adds install location of wheel libs to RPATH of libs built with `kwiver_add_library` 

Changes made to facilitate production of Kwiver python wheels.


dstoup (commented):
Looks good to me.

johnwparent:
Glad to hear it, while I'm here, do you think it's worth reordering the remainder of the FFmpeg dependencies, or are we ready to merge and move on?

dstoup:
Looks like I was supposed to comment, sorry for the delay. I don't think we *need to do anything more but if it makes sense to reorder FFmpeg deps and it doesn't cause undo harm, then I would say yes. Otherwise, we can merge and move on.

johnwparent:
Ok, in that case, I'll give them a quick reorder as per the FFmpeg FAQ and then we can go from there.

PR #1282, opened 2021-04-20
Update sprokit tools search path

russelldj:
Update the `sprokit` applets to search in the recently-added `share/kwiver/<version>/config/` directory. This is very similar to #1258, which fixes the same issue for the `pipeline_runner.` The first commit applies the fix, the second is purely formatting.


russelldj:
It look like the Windows build may be hung.

russelldj:
It still appears that there's an issue with the Windows build. @dstoup @Cookt2 is there a known issue with the Windows runner?

mwoehlke-kitware (approved):

PR #1289, opened 2021-04-26
Remove python virtual env and dependency install from Kwiver cmake workflow

johnwparent:
Removes python dependency work from Kwiver's build process

Instead generates a requirements.txt file from the cmake based on cmake options and adds pip install step to the kwiver setup script

Python tests no longer have guarantee of requirements being met when being run unless setup script is executed, but will still be run by whatever version of python `${KWIVER_PYTHON_MAJOR_VERSION}` points to.


johnwparent:
Currently this requires manual maintenance of the dependencies to be included in the requirements via CMake. A quick revision to the check_deps script would allow us to dynamically determine python dependency requirements, but would not allow for very deterministic package versioning. If we would like a more dynamic approach,  I can provide that fairly easily.

dstoup (changes requested):
Mostly looks good. There is one comment deserving of a response and a few minor nits. Reviewing CI failures now.

CMake/kwiver-setup-python.cmake
```diff
@@ -195,6 +195,27 @@ set(PYTHON_ABIFLAGS "${_python_abi_flags}"
...
+if(SKBUILD)
+  set(PYTHON_REQS "${PYTHON_REQS};scikit-build<=0.11.1")
+endif(SKBUILD)
+
+if(KWIVER_ENABLE_PYTORCH)
+  set(PYTHON_REQS "${PYTHON_REQS};opencv-python>=3.4.2.17,<=4.0.0;pillow>=7.0.0,<=7.1.2;scipy>=1.2,<=1.5")
+endif(KWIVER_ENABLE_PYTORCH)
+
+if(KWIVER_ENABLE_ARROWS)
+  set(PYTHON_REQS "${PYTHON_REQS};torch==1.4.0;torchvision==0.5.0")
```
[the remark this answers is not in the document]
johnwparent: You are 100% correct, will fix.

CMake/kwiver-setup-python.cmake
```diff
@@ -195,6 +195,27 @@ set(PYTHON_ABIFLAGS "${_python_abi_flags}"
...
+#
+set(PYTHON_REQS "numpy>=1.13.0,<=1.19.0;six>=1.10.0,<=1.13.0")
+
+if(SKBUILD)
+  set(PYTHON_REQS "${PYTHON_REQS};scikit-build<=0.11.1")
+endif(SKBUILD)
+
+if(KWIVER_ENABLE_PYTORCH)
+  set(PYTHON_REQS "${PYTHON_REQS};opencv-python>=3.4.2.17,<=4.0.0;pillow>=7.0.0,<=7.1.2;scipy>=1.2,<=1.5")
+endif(KWIVER_ENABLE_PYTORCH)
```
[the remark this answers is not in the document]
johnwparent: Will normalize

CMake/kwiver-setup-python.cmake
```diff
@@ -212,117 +233,20 @@ mark_as_advanced(PYTHON_ABIFLAGS)
...
-      endif()
-      DEACTIVATE_VENV()
-    endif()
-  endif()
+  message(STATUS "KWIVER_PYTHON and KWIVER_TESTING enabled.")
+  message(STATUS "Python tests will be added to CTest.")
+  message(STATUS "Tests are executed by PYTEST, for sucessful test execution, please install the appropriate pytest for your python distribution.")
+  message(STATUS "Testing dependencies will be added to Python requirements.")
+  set(PYTHON_TEST 1)
+  set(PYTHON_REQS "${PYTHON_REQS};nose>1.2;coverage>=4.4.1,<5.0.0;pytest>=4.6,<=6.0;multimethod>=1.2,<=1.4")
```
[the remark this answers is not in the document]
johnwparent: Will make that change!

dstoup:
In addition to the code review comments, I think this patch should target release since it also contains the VEnv and we will want the fix in the minor release.

The Travis CI is failing because of when it's trying to install python deps. It's doing it before the script runs and there isn't a python/requirements.txt file anymore. We can probably remove that line from the 'before_script' section and let it happen at the bottom of the file when - . ./setup_KWIVER.sh is called.
**before_script:
    - pip3 install -q -r python/requirements.txt**

The MAC CI is having some issues with the python install as well but it's less clear what the solution is. Here is the block of output which contains the error message:

Collecting zipp>=0.5 (from importlib-metadata>=0.12; python_version < "3.8"->pytest<=6.0,>=4.6->-r /Users/kitware/Jenkins/workspace/KwiverMacPR/build/python/requirements.txt (line 7))
  Cache entry deserialization failed, entry ignored
  Downloading https://files.pythonhosted.org/packages/0f/8c/715c54e9e34c0c4820f616a913a7de3337d0cd79074dd1bed4dd840f16ae/zipp-3.4.1-py3-none-any.whl
Cache entry deserialization failed, entry ignored
Requirement already up-to-date: pyparsing>=2.0.2 in /Users/kitware/Library/Python/3.6/lib/python/site-packages (from packaging->pytest<=6.0,>=4.6->-r /Users/kitware/Jenkins/workspace/KwiverMacPR/build/python/requirements.txt (line 7))
Installing collected packages: numpy, six, coverage, multimethod, pillow, py, typing-extensions, zipp, importlib-metadata, more-itertools, iniconfig, attrs, toml, packaging
  Found existing installation: numpy 1.17.0
    Uninstalling numpy-1.17.0:
      Successfully uninstalled numpy-1.17.0
  Rolling back uninstall of numpy
Exception:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pip/req/req_install.py", line 851, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pip/req/req_install.py", line 1064, in move_wheel_files
    isolated=self.isolated,
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pip/wheel.py", line 345, in move_wheel_files
    clobber(source, lib_dir, True)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pip/wheel.py", line 323, in clobber
    shutil.copyfile(srcfile, destfile)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
PermissionError: [Errno 13] Permission denied: '/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/numpy/__config__.py'
You are using pip version 9.0.3, however version 21.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

johnwparent:
@dstoup I will absolutely retarget to release, I had planned to and PR'd on autopilot I guess. Seems like the MAC issue is that the version of python being called by calls to `$ python3` is a system install of python, which if I recall, on MacOS, cannot be modified by the user. We have a couple options here. One: we can create a venv before running the MacOS tests/setup script (This is what I had in mind) or when we search for Python in cmake we can specify no MacOS system installs and use that python to install dependencies and run the tests. The drawback to the second approach being that it would prevent users from using a venv if kwiver was not built with an active venv on the path.

dstoup:
> @dstoup I will absolutely retarget to release, I had planned to and PR'd on autopilot I guess. Seems like the MAC issue is that the version of python being called by calls to `$ python3` is a system install of python, which if I recall, on MacOS, cannot be modified by the user. We have a couple options here. One: we can create a venv before running the MacOS tests/setup script (This is what I had in mind) or when we search for Python in cmake we can specify no MacOS system installs and use that python to install dependencies and run the tests. The drawback to the second approach being that it would prevent users from using a venv if kwiver was not built with an active venv on the path.

No worries on the branch targetting .. I do that all the time too. It's hard to remember release when making the branch AND submitting the PR :)

Regarding the MAC issue, I guess I'm not sure what the right approach is here. Certainly for CI, we can have the deps installed or have an env created that is activated during the testing process. That's to say, I think the solution should be one that works best for users and not what's best for CI, since we can likely overcome CI issues.

dstoup:
> > Regarding the MAC issue, I guess I'm not sure what the right approach is here. Certainly for CI, we can have the deps installed or have an env created that is activated during the testing process. That's to say, I think the solution should be one that works best for users and not what's best for CI, since we can likely overcome CI issues.
> 
> Right. I have it configured the way I do so that a user can build kwiver with or without any kind of python virtual environment and the dependencies will be installed after the build to whatever the activate version of python is on the path at that time, whether that's a venv activated at any time or the system. The same is the case for the tests. I think our best approach then is to leave the implementation, and load a venv before running the setup script or tests on macOS.

Thanks .. sounds right to me. I don't know if I have rights on the mac CI machine. I will have Taylor create the env and I can add the activation step to the CI scripts.

dstoup:
Jenkins test this please

dstoup:
MAC machine should be set up with the virtual env now. Hoping for a pass!

johnwparent:
> MAC machine should be set up with the virtual env now. Hoping for a pass!

Fantastic! As soon as we get a pass I'll rebase and then retarget this branch.

dstoup:
Shoot. I just remembered that the fixes I made to release CI didn't extend down to kwiver builds yet. I will fix the CI and rerun.

dstoup:
I think I fixed the Windows CI machines to properly use the target branch to properly build branches targetting release to use the correct Fletch. I'm going to rerun now with the expectation that Windows should pass (after Fletch release is done building) and that all others will still fail. If that assertion holds true, I will fix up the CI configs on the other platforms as well.

dstoup:
Jenkins test this please

dstoup:
BTW, I know I said today that we can probably merge this PR, but I think I would like to wait until everything passes CI. This is the only Kwiver PR currently targeting release, so it will be a good test for the work I'm trying to do.

johnwparent:
> BTW, I know I said today that we can probably merge this PR, but I think I would like to wait until everything passes CI. This is the only Kwiver PR currently targeting release, so it will be a good test for the work I'm trying to do.

Sounds good!

CMake/kwiver-setup-python.cmake
```diff
@@ -195,6 +195,28 @@ set(PYTHON_ABIFLAGS "${_python_abi_flags}"
...
+# to requirements.txt file
+#
+list(APPEND PYTHON_REQS "numpy>=1.13.0,<=1.19.0" "six>=1.10.0,<=1.13.0")
+
+if(SKBUILD)
+  list(APPEND PYTHON_REQS "scikit-build<=0.11.1")
+endif()
+
+if(KWIVER_ENABLE_PYTORCH)
+  list(APPEND PYTHON_REQS "opencv-python>=3.4.2.17,<=4.0.0"
```
Erotemic, line 211: This is a tricky package. See this discussion: https://github.com/opencv/opencv-python/issues/467 Do we want opencv-python or opencv-python-headless? For my pip packaged dependencies I settled on not specifying any variant of cv2 in the skbuild.setup requirements. (I did add a graphics, and headless extra_install, but the user has to explicitly pick and install the one they want, otherwise pip installing a package that specifies one or the other my end up breaking somebodies virtualenv). 

Also, does restricting maximum versions prevent a future and potentially compatible version from running? (Warning is ok).

Erotemic: Thinking about it, it would be nice if the python requirements spec differentiated between maximum versions where you don't know if it will work or not, and maximum versions that prevent known actually broken code from running.

johnwparent: As to the first point, I have no real experience with the distinction between those packages in production, but as Kwiver does use Qt it seems like the same conflicts you experienced could present themselves. With this current setup we could always make it a user configurable option with the default being neither. What are your thoughts there?

These upper limits will restrict future, compatible versions yes. I believe there was a previous discussion on an old PR about how we're maintaining the versions in the requirements.txt and I believe there was at least some small consensus that we would impose upper limits for version ranges to future proof against incompatible packages.

What would be the distinction between those different types of ranges? Maybe I'm missing something, but wouldn't those essentially translate to the same thing in a requirements.txt file? Or did you mean more of an offer two options for version ranges for the packages, one that we know works, and one that may or may not break?

dstoup:
Jenkins test this please

CMake/kwiver-setup-python.cmake
Erotemic: The second point is a musing. I don't think there exists any mechanism in pypi to allow it. Just something to think about.

The first point we should make a decision. It might be right to use opencv-python-headless here instead (or make it configurable). I guess what I need clarification on is what these reqs are being used for. Will these be translated to the setuptools install_requires option? (In which case that will influence users who grab the wheel from pypi) Or is this something else?

dstoup:
Jenkins test this please

dstoup:
The Travis build is getting messed up because doc/release-notes/master.txt exists. There looks to be an assumption in the install-deps.sh file that if master exists in the release-notes dir, then we're building master, otherwise, we build release.

I suspect that file got copied in at some point while creating a new release. @mleotta , I'm thinking we should remove master.txt and ensure that's part of the release process, I couldn't see a better way to determine what the target branch is in Travis.

CMake/kwiver-setup-python.cmake
dstoup: > Also, does restricting maximum versions prevent a future and potentially compatible version from running? (Warning is ok).

I am probably the one who suggested max versions and yes, it's to prevent compatibility issues. A lot of packages are good about deprecating features, API, etc and keeping it as a warning for a reasonable amount of time. But, packages like Numpy (maybe they have changed their ways over the years) have at times been very quick to convert a warning to an error which can cause a lot of immediate panic and scramble.

Either approach requires consistent maintenance but I generally prefer the maintenance coming when someone wants newer features from a package than when a package is left at the latest and breaks one day. That said, if you have a better idea, I am happy to hear it. I'm not deeply familiar with some of the Python best-practices on the topic.

dstoup (approved):
Pending any comments from Jon.

Erotemic:
@dstoup Any comment I had is out of the scope of this PR. It can be addressed at a later time.

So LGTM

mleotta:
Yeah, we should make a checklist for whoever is doing releases.  When
release is reset to master we need to move the contents of `master.txt`
into `release.txt` in preparation for the next release.  The release branch
should never contain a `master.txt`.  Likewise, the versioned releases
should never contain either `master.txt` or `release.txt`.

PR #1291, opened 2021-04-27
[VIAME] Teach refine_detections_watershed an option "uncertain_scale_factor"

neal-siekierski:
This PR adds an option to `refine_detections_watershed` to parametrize how much of the area around a seed is marked as uncertain &ndash; that is, needing to be filled in by the underlying call to the watershed routine. The current behavior is to use the bounding box of the associated detection. The new parameter, `uncertain_scale_factor`, allows scaling that bounding box about its center by a provided factor with a default of 1, which gives the current behavior.

Note that while values greater than 1 may help prevent undersegmentation, anything segmented for the detection outside of its original bounding box won't be included in the mask.

I'm able to run this locally and changing the parameter changes the resulting segmentations in a reasonable way.

### DIFF
--- a/arrows/ocv/refine_detections_watershed.cxx
+++ b/arrows/ocv/refine_detections_watershed.cxx
@@ -36,6 +36,7 @@ public:
 
   bool seed_with_existing_masks = true;
   double seed_scale_factor = 0.2;
+  double uncertain_scale_factor = 1;
 };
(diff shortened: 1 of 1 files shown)


russelldj (approved):
Code looks good and this produces meaningful results for me.

PR #1297, opened 2021-05-03
WIP: Python non-sprokit test refactor to pytest

johnwparent:
This branch refactors the python tests not in sprokit to use pytest compatible frameworks (in most cases unittest or python builtins)
This is another step toward remove the long deprecated nose dependency.


dstoup:
I think these changes look good. We probably need to add pytest and remove nose from the requirements.txt file to make this work correctly.

PR #1298, opened 2021-05-03
WIP: Refactor Python Spokit tests and related CMake

johnwparent:
Sprokit python tests on Windows are currently not running due to issues in the cmake related to multi-config generators. This PR is aimed at mitigating that issue, and refactoring the older CMake used to run the sprokit python testing framework to utilize the same conventions as the rest of the Kwiver python tests. To that end, the sprokit python tests themselves were refactored to support this new approach.


dstoup:
Mac is failing because of the following tests.

The following tests FAILED:
	720 - test-run-pysimple_pipeline-sync (Failed)
	721 - test-run-pysimple_pipeline-thread_per_process (Failed)
	722 - test-run-pysimple_pipeline-pythread_per_process (Failed)
	856 - test-python-serialize_json_utils (Failed)
	857 - test-python-test_serialize_json (Failed)

Ghostflow is complaining about trailing whitespace as well. Since that seems to be on the last commit pushed, it would be good to fix those before moving on to fixing the tests. If you're comfortable rewriting the history so the error goes away with that commit, great. If not, no worries. That doesn't matter as much as having it fixed before merging.

python/kwiver/sprokit/tests/sprokit/adapters/test-embedded_pipeline.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
-# ckwg +28
-# Copyright 2020 by Kitware, Inc.
+#ckwg +28
```
dstoup, line 2: Minor nit .. probably should keep that space in the copyright before "ckwg +28"

Also, here and above, did you mean to set the year back to 2011? At some point we should update the copyright headers to the new style anyway.

python/kwiver/sprokit/tests/sprokit/pipeline/CMakeLists.txt
```diff
@@ -21,14 +21,20 @@ set(__simple_testable_modnames
 
 # KWIVER_ENABLE_TESTS is always true for now (because we are in the test
 # subdir), but the python tests should be moved to the python source dir.
-if (KWIVER_ENABLE_TESTS)
-  include(kwiver-test-setup)
-  # Register individual tests with ctest
+if (KWIVER_ENABLE_TESTS AND PYTEST_RUNNER)
+
+# Register individual tests with ctest
```
dstoup, line 26: Again, really minor but the space before the comment should stay so intending is correct.

dstoup (commented):
CMake looks good to me. It would be good to have someone else look over the Python changes though. Otherwise, nothing but minor space/indent comments and also a little confused by all the copyright date changes from 2012 to 2011 but that will all go away once we replace those headers with the new copyright header that Matthew cooked up.

dstoup:
I will look into the Linux dashboards now to see why they are reporting success and also why they aren't configuring. I have made a ton of CI 'improvements' to make things build correctly with regards to the target branch release vs. master but must have messed something up in the process.

dstoup:
Jenkins test this please

johnwparent:
@dstoup The two failing python tests have been repaired (locally at least) but the remaining three tests are c++ tests that seem to have been failing for some time now, but CTest has been reporting them as having passed. An example of that on the MacOS nightly builds can be found [here](https://open.cdash.org/test/396661258)

dstoup:
Thanks for the heads up. I will look into that issue. Likely test failure output like this is passing through the regular error regex.

PR #1299, opened 2021-05-04
Minor fix so functions in core are exposed in python

chetnieter:
Minor fix so the functions in kwiver.arrows.core are properly exposed and available for use. In particular, before this fix 

```
from kwiver.arrows.core import mesh_triangulate
``` 
was failing.


johnwparent (approved):

johnwparent:
Looks good for a quick patch. I think a more ideal solution would be to restructure the python arrows to avoid the redundant name import, but that can be handled at another time.

dstoup:
Jenkins test this please

PR #1302, opened 2021-05-10
GCC 11 warns us when we are making data copies in range for loop

dstoup:
This PR fixes the range-loop-construct warnings in gcc 11. In all of these cases we are making copies of data in the for loop. I've investigated each instance and believe these changes are all safe to access as const&


mwoehlke-kitware (approved):
LGTM; also, thanks for putting `const` in the correct place! (Mostly thinking of those places where it was "wrong" previously.)

PR #1314, opened 2021-05-21
WIP: [VIAME] Add polygon support to detections

russelldj:
The issues I'm having is this does not appear to properly register the `polygon` attribute when it's used in Python. Do I need to add something to the `vital/bindings/c/types/detected_object.h` file for these new attributes?


mattdawkins:
Did you try both .polygon and .polygon() ? Does the point class used in the vector have python wrappers as well?

python/kwiver/arrows/core/detected_object_set_output_coco.py
```diff
@@ -105,6 +105,9 @@ def write_set(self, detected_object_set, file_name):
                 ],
                 score=det.confidence(),
             )
+            # TODO, figure out how to use masks if they're present 
+            d['segmentations']=det.polygon()
```
tao558, line 109: I think Matt is right. Try changing this line to
```suggestion
            d['segmentations']=det.polygon
```
since `polygon` was bound as an attribute, not a method.

russelldj: That didn't work either. Doing a `dir()` call shows that a number of the bound attributes and method are not present.

russelldj: It appears that there's a version from my first build which is getting installed in `./install/lib/python3.6/site-packages/kwiver/vital/types/detected_object.so` which actually contains the new changes and works properly. However, it's pulling from an outdated file at
`./install/lib/python3.6/site-packages/vital/types/detected_object.so` which is not getting updated.

PR #1315, opened 2021-05-24
Add support for polygons to detections

russelldj:


russelldj:
This should be a clean ff merge that I'm confident in, I just don't have permissions to do it.

russelldj:
This should be updated in the future to use the `polygon` class.

russelldj:
This still needs to land for polygons to work.

dstoup:
It looks like there is some mismatch from the travis build with regards to proj.

**In file included from /home/travis/build/Kitware/kwiver/arrows/proj/geo_conv.cxx:12:0:

/opt/kitware/fletch/include/proj_api.h:37:2: error: #error 'To use the proj_api.h you must define the macro ACCEPT_USE_OF_DEPRECATED_PROJ_API_H'

 #error 'To use the proj_api.h you must define the macro ACCEPT_USE_OF_DEPRECATED_PROJ_API_H'**

I see this is targeting viame. There seems to be some mismatch between the hashes viame is using and the originating hash of this PR. Just speculating there ... The Windows PR is failing to find the fletch it needs:

D:\workspace>tar -xf fletch_install_viame/next.tar 
tar: Error opening archive: Failed to open 'fletch_install_viame/next.tar'

It's possibly a similar issue with Mac, which can't find some fundamental package requirements, e.g. log4cplus. Usually that is caused by a missing or failed Fletch.

mattdawkins:
We can push this into master after

PR #1323, opened 2021-06-01
Uncrustified submodules of Arrows

quantumech3:
Uncrustified the following submodules:

- ceres
- burnout
- core
- cuda
- darknet


mleotta:
This looks like too many files to review in one PR.  I would not do more than one arrow per PR.  Maybe even only part of an arrow for the bigger ones.

PR #1348, opened 2021-06-08
Add working computation of GSD and VNIIRs

russelldj:
This adds computation for GSD and VNIIRS based on the reference implementation provided by NGA. It needs style fixes and, as I understand it, should not modify the input metadata vector. I think this requires implementing the assignment operator for `metadata_item`. Also, some comment formatting needs to be brought in line with the style guide.

### DIFF
--- a/arrows/core/derive_metadata.cxx
+++ b/arrows/core/derive_metadata.cxx
@@ -8,37 +8,25 @@
  */
 
 #include <arrows/core/derive_metadata.h>
+
+#include <vital/math_constants.h>
 #include <vital/types/geo_point.h>
 #include <vital/types/metadata.h>
 #include <vital/types/metadata_tags.h>
 #include <vital/types/metadata_traits.h>
 
 #include <cmath>
+#include <memory>
 
 namespace kv = kwiver::vital;
 using namespace kwiver::vital::algo;
 
 namespace kwiver {
-
 namespace arrows {
-
 namespace core {
-
 namespace {
 
 static ::kwiver::vital::metadata_traits meta_traits;
-
-double
-deg2rad( double deg )
-{
-  return deg * M_PI / 180.0;
-}
-
-double
-rad2deg( double rad )
-{
-  return rad * 180.0 / M_PI;
-}
 
 /*
  * Rotations needed to compute view angle
@@ -51,7 +39,7 @@ RotRoll( double* in, double theta, double* out )
   double in2 = in[ 2 ];
   double sint, cost;
 
-  theta = deg2rad( theta );
+  theta = theta * kv::deg_to_rad;
   sint = sin( theta );
   cost = cos( theta );
 
@@ -70,7 +58,7 @@ RotPitch( double* in, double theta, double* out )
   double in2 = in[ 2 ];
   double sint, cost;
 
-  theta = deg2rad( theta );
+  theta = theta * kv::deg_to_rad;
   sint = sin( theta );
   cost = cos( theta );
 
@@ -89,7 +77,7 @@ RotYaw( double* in, double theta, double* out )
   double in2 = in[ 2 ];
   double sint, cost;
 
-  theta = deg2rad( theta );
+  theta = theta * kv::deg_to_rad;
   sint = sin( theta );
   cost = cos( theta );
 
@@ -136,7 +124,7 @@ view_pitch(
                   right_unrotated ), 0, -platform_pitch, -platform_roll,
           right_unrotated );
 
-  return rad2deg( asin( -forward_unrotated[ 2 ] ) );
+  return asin( -forward_unrotated[ 2 ] ) * kv::rad_to_deg;
 }
 
 double
@@ -148,19 +136,19 @@ compute_gsd( int rows, int cols, kwiver::vital::metadata_sptr md )
   }
 
   /* RP1201 method */
-  auto const platform_pitch_angle = md->find(
+  auto const platform_pitch_deg = md->find(
     kv::VITAL_META_PLATFORM_PITCH_ANGLE ).as_double();
-  auto const platform_roll_angle =
+  auto const platform_roll_deg =
     md->find( kv::VITAL_META_PLATFORM_ROLL_ANGLE ).as_double();
-  auto const sensor_relative_azimuth_angle = md->find(
+  auto const sensor_relative_azimuth_deg = md->find(
     kv::VITAL_META_SENSOR_REL_AZ_ANGLE ).as_double();
-  auto const sensor_relative_elevation_angle = md->find(
+  auto const sensor_relative_elevation_deg = md->find(
     kv::VITAL_META_SENSOR_REL_EL_ANGLE ).as_double();
-  auto const sensor_relative_roll_angle = md->find(
+  auto const sensor_relative_roll_deg = md->find(
     kv::VITAL_META_SENSOR_REL_ROLL_ANGLE ).as_double();
-  auto const fov_h_deg =
+  auto const sensor_horizontal_fov =
     md->find( kv::VITAL_META_SENSOR_HORIZONTAL_FOV ).as_double();
-  auto const fov_v_deg =
+  auto const sensor_vertical_fov =
     md->find( kv::VITAL_META_SENSOR_VERTICAL_FOV ).as_double();
   auto slant_range = md->find( kv::VITAL_META_SLANT_RANGE ).as_double();
   auto const sensor_location = kv::any_cast< kv::geo_point >(
@@ -172,30 +160,31 @@ compute_gsd( int rows, int cols, kwiver::vital::metadata_sptr md )
   auto const target_width =
     md->find( kv::VITAL_META_TARGET_WIDTH ).as_double();
 
-  if( platform_pitch_angle >= -20 && platform_pitch_angle <= 20 &&
-      platform_roll_angle >= -50 && platform_roll_angle <= 50 &&
-      sensor_relative_azimuth_angle >= 0 &&
-      sensor_relative_azimuth_angle <= 360 &&
-      sensor_relative_elevation_angle >= -180 &&
-      sensor_relative_elevation_angle <= 180 &&
-      sensor_relative_roll_angle >= 0 && sensor_relative_roll_angle <= 360 &&
-      fov_h_deg >= 0 && fov_h_deg <= 180 &&
-      fov_v_deg >= 0 && fov_v_deg <= 180 &&
+  if( platform_pitch_deg >= -20 && platform_pitch_deg <= 20 &&
+      platform_roll_deg >= -50 && platform_roll_deg <= 50 &&
+      sensor_relative_azimuth_deg >= 0 &&
+      sensor_relative_azimuth_deg <= 360 &&
+      sensor_relative_elevation_deg >= -180 &&
+      sensor_relative_elevation_deg <= 180 &&
+      sensor_relative_roll_deg >= 0 && sensor_relative_roll_deg <= 360 &&
+      sensor_horizontal_fov >= 0 && sensor_horizontal_fov <= 180 &&
+      sensor_vertical_fov >= 0 && sensor_vertical_fov <= 180 &&
       slant_range > 0 && slant_range <= 5000000 )
   {
-    double gsd_horiz, gsd_vert;
-
-    double const absolute_elevation_angle = view_pitch(
-      platform_pitch_angle,
-      platform_roll_angle,
-      sensor_relative_azimuth_angle,
-      sensor_relative_elevation_angle,
-      sensor_relative_roll_angle );
+    double const absolute_elevation_deg = view_pitch(
+      platform_pitch_deg,
+      platform_roll_deg,
+      sensor_relative_azimuth_deg,
+      sensor_relative_elevation_deg,
+      sensor_relative_roll_deg );
 
     // Convert angles to radians
-    auto const fov_h_rad = deg2rad( fov_h_deg );
-    auto const fov_v_rad = deg2rad( fov_v_deg );
-    auto const elev_rad = deg2rad( absolute_elevation_angle );
+    auto const sensor_horizontal_fov_rad =
+      sensor_horizontal_fov * kv::deg_to_rad;
+    auto const sensor_vertical_fov_rad =
+      sensor_vertical_fov  * kv::deg_to_rad;
+    auto const absolute_elevation_rad =
+      absolute_elevation_deg * kv::deg_to_rad;
 
     // If Slant Range is missing, compute from Elevation Angle and Sensor
     // Altitude */
@@ -205,27 +194,30 @@ compute_gsd( int rows, int cols, kwiver::vital::metadata_sptr md )
         19000 )
     {
       slant_range = ( sensor_altitude - frame_center_elevation ) /
-                    sin( -elev_rad );
+                    sin( -absolute_elevation_rad );
     }
 
     // GSD horizontal
-    gsd_horiz = 2.0 * slant_range * tan( ( fov_h_rad / 2.0 ) / cols );
+    double const gsd_horizontal = 2.0 * slant_range *
+                     tan( ( sensor_horizontal_fov_rad / 2.0 ) / cols );
 
     // GSD vertical
-    gsd_vert  = ( 2.0 * slant_range * tan( ( fov_v_rad / 2.0 ) / rows ) ) /
-                sin( -elev_rad );
+    double const gsd_vertical = ( 2.0 * slant_range *
+                     tan( ( sensor_vertical_fov_rad / 2.0 ) / rows ) ) /
+                   sin( -absolute_elevation_rad );
 
     // GSD = Geometric mean of horiz & vert GSD
-    return 1000.0 * sqrt( gsd_horiz * gsd_vert );
+    return 1000.0 * sqrt( gsd_horizontal * gsd_vertical );
   }
 
   /* RP 1201 horizontal axis only */
 
   if( slant_range > 0 && slant_range <= 5000000 &&
-      fov_h_deg >= 0 && fov_h_deg <= 180 )
+      sensor_horizontal_fov >= 0 && sensor_horizontal_fov <= 180 )
   {
     return 2000.0 * slant_range *
-           tan( ( deg2rad( fov_h_deg ) / 2.0 ) / (float) cols );
+           tan( ( sensor_horizontal_fov * kv::deg_to_rad / 2.0 ) /
+                (float) cols );
   }
 
   /* Target width horizontal axis only */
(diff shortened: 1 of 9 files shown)


russelldj:
Note that the computation of `VNIIRs` differs from that provided in the reference implementation in that it does not include any image terms. We will add those later and for now only use the term based on GSD. A deeper understanding of the equation may be useful to determine whether we should multiply/scale this computation to account for the likely value of the missing terms.

russelldj:
We should use the KWIVER versions of utility functions instead of reimplementing them here.

russelldj:
It looks like `5ccc757` adds bad whitespace. So if you could do an iterative rebase, `git rebase -i 30ae5c1` to fix that, that would be great. Actually, that namespace file is what `uncrustify` produces so it's fine to leave as it was before. So I'd suggest just dropping that commit.

daniel-riehm:
Should the public-facing compute_derived_metadata fail silently if the values can't be computed (e.g. just return an unchanged copy of the `metadata` parameter)? Or should it throw an exception to be caught / crash? I don't know enough about the context of how this will be called to decide this...

russelldj:
> Should the public-facing compute_derived_metadata fail silently if the values can't be computed (e.g. just return an unchanged copy of the `metadata` parameter)? Or should it throw an exception to be caught / crash? I don't know enough about the context of how this will be called to decide this...

In my opinion it would be best to compute everything we can and avoid forwarding any errors to the user. It seems like the only situation where this is applicable is deriving `slant_range` since both `GSD` and `VNIIRs` rely on the same as well as the slant range. It's likely this process will be a (relatively small) part of a pipeline that's run on a variety of videos where we cannot assume we'll have complete metadata.

russelldj:
> Should the public-facing compute_derived_metadata fail silently if the values can't be computed (e.g. just return an unchanged copy of the `metadata` parameter)? Or should it throw an exception to be caught / crash? I don't know enough about the context of how this will be called to decide this...

In my opinion it would be best to compute everything we can and avoid forwarding any errors to the user. It seems like the only situation where this is applicable is deriving `slant_range` since both `GSD` and `VNIIRs` rely on the same as well as the slant range. It's likely this process will be a (relatively small) part of a pipeline that's run on a variety of videos where we cannot assume we'll have complete metadata.

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,374 @@
...
+kwiver::vital::metadata_vector
+compute_derived_metadata( kwiver::vital::metadata_vector metadata_vec,
+                          size_t frame_width, size_t frame_height )
+{
+  kv::metadata_vector updated_values;
+
+  for( auto const metadata : metadata_vec )
+  {
+    // Deep copy metadata
+    auto updated_metadata =
```
russelldj, line 333: I can't suggest changes on my own PR I guess...but this should be updated to `std::make_shared< kv::metadata >( metadata->deep_copy() )`. My own branch is pretty dirty right now.

russelldj:
I'd advocate for calling this complete so we can start doing integration. We can open a follow-on PR for the unit test and evaluation of correctness.

vital/types/metadata.cxx
```diff
@@ -97,6 +97,22 @@ metadata
 ::metadata()
 { }
 
+// ---------------------------------------------------------------------
+metadata
+metadata
+::deep_copy()
```
mwoehlke-kitware, line 103: I wonder if this should instead be `clone( clone_type = clone_type::DEEP )`...

vital/types/metadata.cxx
```diff
@@ -97,6 +97,22 @@ metadata
...
+{
+  metadata copy;
+  copy.set_timestamp( m_timestamp );
+
+  for( auto const& md_item: m_metadata_map )
```
mwoehlke-kitware, line 108:
```suggestion
  for( auto const& md_item : m_metadata_map )
```

mwoehlke-kitware (changes requested):

vital/types/metadata.cxx
russelldj: I'd imagine so. I wasn't aware of this pattern.

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,373 @@
...
+      // Compute VNIIRS
+      auto const rer_value = compute_rer();
+      auto const snr_value = compute_snr();
+      auto const vniirs_value = compute_vniirs( gsd_value, rer_value, snr_value );
+      auto const& vniirs_trait = meta_traits.find( kv::VITAL_META_VNIIRS );
+      updated_metadata->add( vniirs_trait.create_metadata_item( vniirs_value ) );
+    }
+    catch ( kv::invalid_value const& e )
+    {
+      // Fail silently
```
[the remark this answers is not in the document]
daniel-riehm: This section fills in three metadata values - slant range, gsd, and vniirs. If it fails while calculating gsd, shouldn't it still pass on the slant range value, which it was able to calculate without error?

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    updated_values.push_back( updated_metadata );
+  }
+
+  return updated_values;
+}
+
+} // namespace core
```
daniel-riehm, line 355: If we are going to accept blank lines between namespace openings / closings because that's how uncrustify does it, shouldn't the style guide be changed to reflect this?

mwoehlke-kitware: Possibly 🙂. There was some notion of eventually "fixing" Uncrustify to be able to remove those, but meanwhile some people have been pushing to just accept them. So I guess the "answer" is that this is a good question that I'd be inclined to possibly raise to a larger audience. (But I suspect no one cares enough about the no-blanks style...)

arrows/core/derive_metadata.cxx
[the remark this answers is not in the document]
mwoehlke-kitware: Ah, right; for some reason I was thinking if an exception happened here, `updated_metadata` would be unmodified. So, yeah, this is probably fine.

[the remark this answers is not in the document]
russelldj: I agree with Daniel.

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,373 @@
...
+  {
+    sensor_fov[ 1 ] = sensor_fov[ 0 ] * frame_height / frame_width;
+  }
+
+  // Intermediate Values
+  kv::rotation_d const total_rotation = get_total_rotation( metadata );
+  double yaw, pitch, roll;
+  total_rotation.get_yaw_pitch_roll( yaw, pitch, roll );
+
+  double const slant_range = get_slant_range( metadata );
```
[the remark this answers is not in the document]
daniel-riehm: When I wrote that I thought there was a compelling reason, but looking back at it there really isn't. Changed.

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,350 @@
...
+    kv::rotation_d const total_rotation = get_total_rotation( metadata );
+    double yaw, pitch, roll;
+    total_rotation.get_yaw_pitch_roll( yaw, pitch, roll );
+
+    // Determine the altitude of the sensor above the frame center
+    kv::geo_point const sensor_location = get_sensor_location( metadata );
+    double const sensor_altitude = sensor_location.location()[ 2 ];
+    kv::geo_point const frame_center = get_frame_center( metadata );
+    double const frame_center_altitude = frame_center.location()[ 2 ];
+    double const altitude_difference = sensor_altitude - frame_center_altitude;
```
russelldj, line 200: I was initially concerned this was wrong because the `SENSOR_LOCATION` claims to have an altitude while the `FRAME_CENTER` claims to have an elevation. I read the former as height above the ground while the later was height above sea level. However, a read of the [MISB standard ](https://gwg.nga.mil/misb/docs/standards/ST0601.6.pdf) suggests the code is correct and these two terms are actually synonymous. Perhaps adding some comment here would be useful, or additionally updating the documentation in `metadata_tags.h` would be useful.

vital/types/metadata.cxx
daniel-riehm: From what I can tell, a shallow copy is not really a thing here, since the `metadata` class takes ownership of its `metadata_item`s, so by necessity they would be copied as well. @mwoehlke-kitware Would you recommend still accepting the `clone_type` parameter and implementing the same behavior for both options, or just a rename from `deep_copy` to `clone`...?

russelldj: Would copying `m_metadata_map` not result in a shallow copy, since this is a `std::map< vital_metadata_tag, std::shared_ptr< metadata_item >`?

mwoehlke-kitware: That was my thought also; it seems like a shallow clone is just the copy ctor...

daniel-riehm: Oh, yes, you're right. Oops.

daniel-riehm: But consider `vital/types/metadata.h:371`. We don't handle both cases, and the case in which `item_ptr` is a `unique_ptr`, the copy constructor will fail.

arrows/core/tests/test_derive_metadata.cxx
```diff
@@ -0,0 +1,127 @@
...
+} // end namespace
+
+class derive_metadata : public ::testing::Test
+{
+  void
+  SetUp()
+  {
+    auto meta = make_metadata();
+    size_t const frame_height{ 720 };
+    size_t const frame_width{ 1080 };
```
mwoehlke-kitware, line 88: This is... an interesting way to declare variables?
```suggestion
    constexpr auto frame_height = size_t{ 720 };
    constexpr auto frame_width = size_t{ 1080 };
```

...although TBH these can probably just be passed to `compute_derived_metadata` directly.

arrows/core/tests/test_derive_metadata.cxx
```diff
@@ -0,0 +1,127 @@
...
+    derived_metadata = kwiver::arrows::core::compute_derived_metadata(
+      meta, frame_width, frame_height );
+  }
+
+public:
+  kv::metadata_vector derived_metadata;
+};
+
+// ----------------------------------------------------------------------------
+TEST_F(derive_metadata, compute_derived)
```
mwoehlke-kitware, line 99:
```suggestion
TEST_F( derive_metadata, compute_derived )
```

arrows/core/tests/test_derive_metadata.cxx
```diff
@@ -0,0 +1,127 @@
...
+{
+  kv::metadata_item const& gsd_value = derived_metadata.at( 0 )->find(
+    kv::VITAL_META_AVERAGE_GSD );
+  kv::metadata_item const& vniirs_value = derived_metadata.at( 0 )->find(
+    kv::VITAL_META_VNIIRS );
+  kv::metadata_item const& slant_range_value = derived_metadata.at( 0 )->find(
+    kv::VITAL_META_SLANT_RANGE );
```
mwoehlke-kitware, line 106:
```suggestion
  kv::metadata_item const& gsd_value =
    derived_metadata.at( 0 )->find( kv::VITAL_META_AVERAGE_GSD );
  kv::metadata_item const& vniirs_value =
    derived_metadata.at( 0 )->find( kv::VITAL_META_VNIIRS );
  kv::metadata_item const& slant_range_value =
    derived_metadata.at( 0 )->find( kv::VITAL_META_SLANT_RANGE );
```

(Also, below)

arrows/core/tests/test_derive_metadata.cxx
```diff
@@ -0,0 +1,127 @@
...
+
+  EXPECT_FLOAT_EQ( gsd_value.as_double(), 0.38253304 );
+  EXPECT_FLOAT_EQ( vniirs_value.as_double(), 5.183558 );
+  EXPECT_FLOAT_EQ( slant_range_value.as_double(), 300 );
```
mwoehlke-kitware, line 110: Two issues:
- Use `EXPECT_DOUBLE_EQ`, not `EXPECT_FLOAT_EQ`
- Assertions take arguments as `expected, actual`

```suggestion
  EXPECT_DOUBLE_EQ( 0.38253304, gsd_value.as_double() );
  EXPECT_DOUBLE_EQ( 5.183558, vniirs_value.as_double() );
  EXPECT_DOUBLE_EQ( 300, slant_range_value.as_double() );
```

(Also, below)

arrows/core/tests/test_derive_metadata.cxx
russelldj: Everywhere else I've seen uses no spaces for `TEST*`google test macros. I was fairly sure this was by convention.

vital/types/metadata.cxx
daniel-riehm: We might even consider deleting the copy constructor for the class altogether to prevent any code which would break when `item_ptr` is a `unique_ptr`

arrows/core/tests/test_derive_metadata.cxx
mwoehlke-kitware: It *is* de-facto convention, but it's an exception with no real justification, and I'm not aware that anyone is really arguing for that exception; rather, it is an accident. It's also getting "corrected" as we start working to apply Uncrustify across the code base.

(On that note, as currently configured, Uncrustify wants to also add a space in `TEST (`, however that's considered a bug / misconfiguration.)

russelldj: I take that back, it looks like a mixed bag. But still, it appears `[ws.space]` suggests this style.

mwoehlke-kitware: Hmm, so it does. I'm not sure *why* that's the case, though, other than "inertia". Perhaps, like blank lines between namespaces, we should revisit that...

vital/types/metadata.cxx
mwoehlke-kitware: Ah, right. Um... on second thought, maybe this should just *be* the copy ctor. Especially as the default ctor is broken (and being used in this PR!).

So, yes, please ensure the copy ctor is properly specified (i.e. either implemented or explicitly deleted). If you turn this into a copy ctor, please also explicitly default the move ctor and move assignment operator. (You might then delete the non-move assignment operator as not currently needed, perhaps with a note that it can be implemented if needed. Or you could implement it in terms of making a local copy and then move-assigning the copy to `this`. IOW, `T& op=(T const& o) { auto c = o; *this = std::move(c); return *this; }`.)

arrows/core/tests/test_derive_metadata.cxx
russelldj: If you view the raw `code-style.rst` files, there's a subitem under that stating this is mostly historical with no good justification, but that's not rendering properly.

vital/types/metadata.cxx
russelldj: Making this the copy ctor was my initial reaction. However, I was scared this could represent a breaking change that might be hard to track down across code unrelated to this PR. If you're comfortable with it, I'd agree that makes sense.

mwoehlke-kitware: (Oh, and there is precedent — within the metadata classes, even — for `clone` not taking a `clone_type`; if it doesn't, I believe that implies that only deep copies are possible. But it should return a pointer type. I'd prefer not a raw pointer, as that can lead to memory leaks, but there is unfortunately precedent for doing so.)

arrows/core/tests/test_derive_metadata.cxx
mwoehlke-kitware: Ah... and it suggests nuking the exception if/when we start using automated formatting. Guess what we're starting to do 🙂.

(It *is* rendering correctly. It's a reST comment; the correct rendering is to not render, which it is correctly doing.)

vital/types/metadata.cxx
mwoehlke-kitware: Heh. On that note, there are a couple issues if the `unique_ptr` code was enabled. Which it *ought* to be, darn it; the use of `shared_ptr` is supposed to only be for broken STLs. I'm working on a PR for that. Meanwhile, yeah, I'm fine making this the copy ctor; any code currently using the copy ctor is currently broken. (Which includes *this PR*, but if you make this a copy ctor, that spot should naturally fix itself.)

russelldj:
I think this is ready to land. Any final objections? Also, given the number of commits here, should we rebase to squash some of them?

vital/types/metadata.cxx
russelldj: Forgot to close this; I believe everything should be addressed.

mleotta (changes requested):
See inline comments. For some of them (like requiring both h_fov and v_fov), if this is what the reference code we have does then we should probably leave that as is for initial testing compatibility, but later we might want to change this to be most robust.

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,352 @@
...
+      std::make_shared< kv::metadata >( *metadata.get() );
+
+    try
+    {
+      // Compute slant range. Must be inserted before GSD calculation
+      auto const slant_range_value = compute_slant_range( updated_metadata );
+      auto const& slant_range_trait =
+        meta_traits.find( kv::VITAL_META_SLANT_RANGE );
+      updated_metadata->add(
+        slant_range_trait.create_metadata_item( slant_range_value ) );
```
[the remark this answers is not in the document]
russelldj: We only compute a new slant range if it is not present.

arrows/core/derive_metadata.cxx
```diff
@@ -308,7 +308,7 @@ compute_derived_metadata( kwiver::vital::metadata_vector const& metadata_vec,
   {
     // Deep copy metadata
     auto updated_metadata =
-      std::make_shared< kv::metadata >( metadata->deep_copy() );
+      std::make_shared< kv::metadata >( *metadata.get() );
```
mwoehlke-kitware, line 311: BTW, `.get()` is superfluous here:
```suggestion
      std::make_shared< kv::metadata >( *metadata );
```

mwoehlke-kitware (commented):
Squashing is probably good. What's the current story on handling only one FOV?

russelldj:
> Squashing is probably good. What's the current story on handling only one FOV?

Initially, we were just going to land this as-is and then fix the single-FOV on a later branch. I'll squash these changes and push them up. If it doesn't land before I fix the FOV thing, I'll at that as well.

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,352 @@
...
+  return slant_range;
+}
+
+// ----------------------------------------------------------------------------
+double
+compute_gsd( kwiver::vital::metadata_sptr const& metadata,
+             size_t frame_width, size_t frame_height )
+{
+  if ( frame_width == 0 || frame_height == 0 ) {
+    VITAL_THROW( kv::invalid_value, "frame dimensions cannot be zero" );
```
mwoehlke-kitware, line 214: Minor (and not critical if this doesn't happen), but...
```suggestion
  if ( frame_width == 0 || frame_height == 0 )
  {
    VITAL_THROW( kv::invalid_value, "frame dimensions cannot be zero" );
```

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,352 @@
...
+} // namespace <anonymous>
+
+// ----------------------------------------------------------------------------
+kwiver::vital::metadata_vector
+compute_derived_metadata( kwiver::vital::metadata_vector const& metadata_vec,
+                          size_t frame_width, size_t frame_height )
+{
+  kv::metadata_vector updated_values;
+
+  for( auto const metadata : metadata_vec )
```
mwoehlke-kitware, line 307:
```suggestion
  for( auto const& metadata : metadata_vec )
```

arrows/core/derive_metadata.cxx
```diff
@@ -0,0 +1,352 @@
...
+  {
+    // Deep copy metadata
+    auto updated_metadata =
+      std::make_shared< kv::metadata >( *metadata );
```
mwoehlke-kitware, line 311: (Optional)
```suggestion
    auto updated_metadata = std::make_shared< kv::metadata >( *metadata );
```

arrows/core/tests/test_derive_metadata.cxx
```diff
@@ -0,0 +1,115 @@
...
+  m1->add< kv::VITAL_META_TARGET_WIDTH >( 100 );
+
+  // Add the geo point traits
+  m1->add< kv::VITAL_META_SENSOR_LOCATION >(
+    { kv::geo_point::geo_3d_point_t{ 0, 0, 12 }, kv::SRID::lat_lon_WGS84 } );
+  m1->add< kv::VITAL_META_FRAME_CENTER >( {
+      kv::geo_point::geo_3d_point_t{ 0, 0, 0 }, kv::SRID::lat_lon_WGS84 } );
+
+  // Replicate these values except without the slant range field
+  auto const m2 = std::make_shared< kv::metadata >( *m1.get() );
```
mwoehlke-kitware, line 53: Heh, missed this one:
```suggestion
  auto const m2 = std::make_shared< kv::metadata >( *m1 );
```

mwoehlke-kitware (commented):
With the understanding that we're punting on the FOV thing for now (n.b. @mleotta's comment), this looks good aside from use of `.get()` is now inconsistent 🙂.

mleotta (approved):

russelldj:
I'd like to see this land first and then have a follow-on branch to make the computations more correct. I was looking through the current approach and it seems like there are some non-trivial differences with the reference implementation.

PR #1358, opened 2021-06-25
Fix depth formula and near plane.

danlipsa:


danlipsa:
@mleotta @borovik135 Please review. The real culprit is the non linear dependency between value in the depth buffer and depth.
See Comparison of Depth Buffer Precision plot at http://www.songho.ca/opengl/gl_projectionmatrix.html

danlipsa:
What Matt suggested - computing the near and far depth might still be needed at one point, but not for this dataset. Right now I use the bounding box corners to compute that instead of the actual mesh points.

danlipsa:
Also, presenting the user with the current near and far values, and allowing them to modify would work well for difficult datasets where even computing near and far from the mesh points would not work - For example when you have some irrelevant data that makes your [near, far] interval too big.

borovik135:
Image-to-mesh projection results now indeed show occlusion removal, react to occlusion threshold values, and can be used in analysis of camera model quality, e.g. regarding lens distortion estimation. Yet in some regions they look a bit confusing, blacking-out non-occluded mesh regions. I realize that what we have is an approximation, but ideally the decision should not depend on a threshold, should it? a 3D point is either occluded or it is not, given a camera model, right?
Screenshot from 2021-06-25 13-47-22

danlipsa:
This is what I get for Occlusion threshold 0. The round removed portion before the building is the frustum intersected with the mesh I think.

danlipsa:
Unless you use different data than I do.

danlipsa:
Probably lens distortion plays a role as well.

borovik135:
The black spot in front of the building is the artifact of the lens distortion estimation error, and once corrected, it will be gone, so that is not relevant to the occlusion. I am more concerned with the black out on the roof and to the side of the building: those areas do not seem to be distorted, and should be colored, but are not, as shown below.
Screenshot from 2021-06-25 13-49-29

borovik135:
I think we are using the same data. I guess I am still struggling with the occlusion dependency on a threshold, since to me it appears to be a binary decision: a point is either occluded or not.

danlipsa:
Probably you should leave the threshold 0 (the default) in most cases - increasing it would add back a surface that was judged occluded but was not. I think we were trying to account for very close surfaces. With my newly gained understanding, I think adding (near, far) depths as user parameters, where you can modify the computed values, would be more effective and more intuitive as well.

danlipsa:
I don't get the black on the roof. Let me know if you want to connect in a hangout to make sure we get the same results.

danlipsa:
@borovik135 Thinking more about your suggestions:
- The occlusion test I have seems correct: Initially all points are occluded. If a point tests not occluded I show it. A point occlusion status cannot change depending on the order of traversal. It is either occluded or not.
- The initialization of the array where I read the depth buffer does not matter I think. The whole depth buffer is initialized by OpenGL and copied over my array.

danlipsa:
This picture shows the depths I compute from the depth buffer and the depth I get from kwiver. Interestingly, the range of values for the opengl depths is shifted to the right with about 20.
depths

borovik135:
@danlipsa, thanks for the tests and thoughts. Frankly, I still think that the vertex traversal order matters, as some vertices/triangles may come in front for the already traversed vertices, and the latter need to be uncolored. I also think the initial depth values in our z-buffer should be the far plane or even infinity. I will be happy to take a look at the algorithm implementation, if you send me a link to it.

danlipsa:
Vertex coloring is not used in the occlusion test. Because of that, I don't think the order matters.

You can see the occlusion test after the depthBufferValue computation in the diff. 
depthBuffer[ idData ].Buffer is the depth buffer for a certain frame = idData (sorry not my variable names)

I am inclined to think that the shift I see in the range is the cause of both: 1. the ac units on the roof not showing occlusion; 2. the behavior you see where the whole roof is occluded (maybe in your case the shift is in the other direction).
Maybe an uninitialized value in the conversion in vtkKwiverCamera?

danlipsa:
For a different camera the difference in depth is less pronounced: about 5 close to near plane, 10 around the near corner of the building and around 100 closer to the far plane - this does look like a precision issue.
This picture shows the difference between depth buffer depth and quiver depth, yellow is the init value.

difference

danlipsa:
Another thought: does camera distortion affect depth computation?

borovik135:
> Another thought: does camera distortion affect depth computation?
it may, if the ray tracing take into account lens distortion

borovik135:
> For a different camera the difference in depth is less pronounced: about 5 close to near plane, 10 around the near corner of the building and around 100 closer to the far plane - this does look like a precision issue.
> This picture shows the difference between depth buffer depth and quiver depth, yellow is the init value.
> 
> difference
depth estimations seem to agree OK in the area between the camera and building speculatively because in that area the model is nearly pin-hole; we may want to try a camera model with no distortion, e.g. by setting all the distortion coefficients to 0 in the KRtd files, to see what the differences in depth estimation may be, as well as what vertex coloring may bring as a result.

arrows/vtk/mesh_coloration.cxx
```diff
@@ -649,7 +650,7 @@ ::render_depth_buffer(
   // we only render points in front of the camera
   if( depthRange[ 0 ] < 0 )
   {
-    depthRange[ 0 ] = 0;
+    depthRange[ 0 ] = 0.01;
```
mleotta, line 653: This is potentially problematic for a several reasons.  First, this value of `0.01` only kicks in for values less than zero.  So if the actual near depth is zero this offset is not applied.  The same is true for any values less than 0.01.  To correct this hack we also need to change line 650 to `if( depthRange[ 0 ] < 0.01 )`.

However, even with the above change, this 0.01 is very arbitrary and makes assumptions about the scale of the scene.  This will not work for a scene where the overall scale is very small.  The "right" way to do this is to project all the vertices and check all vertices in front of the camera that lie in the image.  But that is expensive.  An alternate hack that is likely safer than 0.01 is to scale that threshold relative to `depthRange[ 1 ]`, so maybe ` depthRange[ 0 ] = depthRange[ 1 ] * 0.01` or similar.

mleotta (changes requested):

arrows/vtk/mesh_coloration.cxx
danlipsa: Sounds good.

danlipsa:
@mleotta Is this ready?

danlipsa:
@mleotta Do you want to also merge this for the release. It is certainly an improvement to what we had before.

mleotta (commented):
Can you add a release note describing this bug fix to `doc/release_notes/release.txt`?

mleotta:
I'm going to get this merged and then add the release note

danlipsa:
Great. Thanks Matt. Sorry, I have not seen your request to add a note in the release.txt.

PR #1359, opened 2021-06-28
Create Colmap arrow

quantumech3:
Created a colmap arrow with a build configuration for colmap and a small test to verify that the library links properly


mleotta:
It looks like you are missing all the files in the 'colmap' subdirectory.  These need to be added to the commit.  Also, you should not have the commit for the template arrow on this PR.  This PR should be just adding the colmap arrow.

quantumech3:
> It looks like you are missing all the files in the 'colmap' subdirectory. These need to be added to the commit. Also, you should not have the commit for the template arrow on this PR. This PR should be just adding the colmap arrow.

Fixed

mleotta (changes requested):
This is a good start, but see inline comments.

arrows/colmap/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+)
+
+kwiver_install_headers(
+    ${CMAKE_CURRENT_BINARY_DIR}/kwiver_algo_colmap_export.h
+    NOPATH SUBDIR arrows/colmap
+)
+
+set(
+    plugin_colmap_sources
+    colmap.cpp
```
chetnieter, line 18: We use the .cxx extension for source files in kwiver rather than .cpp.

arrows/colmap/colmap.cpp
```diff
@@ -0,0 +1,11 @@
+#include <arrows/colmap/colmap.h>
+#include <colmap/mvs/image.h>
```
chetnieter, line 2: Compilation fails for me here. This should be resolved by dealing with Matt's comment about finding the colmap library.

chetnieter (changes requested):

arrows/CMakeLists.txt
```diff
@@ -102,3 +102,6 @@ if( KWIVER_ENABLE_VTK)
 endif()
 
 add_subdirectory( serialize )
+
+# Register the colmap arrow with CMake.
+add_subdirectory( colmap )
```
chetnieter, line 107: This need to be wrapped in ```if```, ```endif``` with KWIVER_ENABLE_COLMAP. See what was done for Ceres:

https://github.com/Kitware/kwiver/blob/master/arrows/CMakeLists.txt#L17-L20

chetnieter (changes requested):

arrows/CMakeLists.txt
chetnieter: This still needs to be wrapped.

```
if ( KWIVER_ENABLE_COLMAP )
  add_subdirectory( colmap )
endif()
```

arrows/colmap/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+    NOPATH SUBDIR arrows/colmap
+)
+
+set(
+    plugin_colmap_sources
+    colmap.cxx
+)
+
+kwiver_add_library(
+    kwiver_algo_colmap
```
chetnieter, line 22: I am now getting several errors at link time

```
undefined reference to `boost::system::generic_category()'
```

This may be due to issues with multiple copies of boost on my system. Does everything compile and link for you?

chetnieter (changes requested):

arrows/colmap/CMakeLists.txt
quantumech3: Yes, this was an issue I had before. I actually addressed these issues earlier but I must have wiped my changes that made this arrow use the package variables and wrapped the including of the arrow in an if statement

CMake/kwiver-depends-COLMAP.cmake
```diff
@@ -0,0 +1,11 @@
+# Optionally find and configure COLMAP dependency
+
+option( KWIVER_ENABLE_COLMAP
+  "Enable COLMAP dependent code and plugins (Arrows)"
+  ON
```
chetnieter, line 5: Let's change this to OFF by default

chetnieter (changes requested):
This is looking good. I was able to build things on my linux with the colmap arrow enabled. I think the next step would to create a simple test that calls your version() method.

arrows/colmap/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+target_link_libraries(
+    kwiver_algo_colmap
+    PUBLIC vital
+    ${COLMAP_LIBRARIES}
+)
+
+if (KWIVER_ENABLE_TESTS)
+  add_subdirectory(tests)
+endif()
```
chetnieter, line 45: Please add a newline at the end of the file

arrows/colmap/colmap.cxx
```diff
@@ -0,0 +1,65 @@
+#include <arrows/colmap/colmap.h>
```
chetnieter, line 1: Let's rename this file `image_container.cxx` and the header file `image_container.h` to be consistent with the other arrows.

arrows/colmap/colmap.h
```diff
@@ -0,0 +1,33 @@
...
+#include <vital/types/image_container.h>
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace colmap_arrow {
+
+KWIVER_ALGO_COLMAP_EXPORT
+void version();
```
chetnieter, line 16: Let's remove this method since we put it in there just test we where correctly accessing the library.

arrows/colmap/tests/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+set(CMAKE_FOLDER "Arrows/colmap/Tests")
+
+include(kwiver-test-setup)
+
+set(test_libraries      colmap vital vital_vpm kwiver_algo_colmap kwiver_algo_ocv)
+
+##############################
+# Algorithms Colmap tests
+##############################
+kwiver_discover_gtests(colmap colmap                 LIBRARIES ${test_libraries})
```
chetnieter, line 12: Add a newline

arrows/colmap/tests/test_colmap.cxx
```diff
@@ -0,0 +1,73 @@
...
+#include <fstream>
+#include <gtest/gtest.h>
+#include <iostream>
+#include <test_gtest.h>
+#include <vital/algo/image_io.h>
+#include <vital/plugin_loader/plugin_manager.h>
+#include <vital/types/image.h>
+#include <vital/types/image_container.h>
+
+char* logo_path;
```
chetnieter, line 12: We should put in the file path to the test image relative to the test directory. Take a look at how it is done in the ocv arrow - [here](https://github.com/Kitware/kwiver/blob/11bce47e962ed6577b6425887522bbcd54e98132/arrows/ocv/tests/test_inpaint.cxx#L20-L22),  [here](https://github.com/Kitware/kwiver/blob/11bce47e962ed6577b6425887522bbcd54e98132/arrows/ocv/tests/test_inpaint.cxx#L36),  [here](https://github.com/Kitware/kwiver/blob/11bce47e962ed6577b6425887522bbcd54e98132/arrows/ocv/tests/test_inpaint.cxx#L46), and [here](https://github.com/Kitware/kwiver/blob/11bce47e962ed6577b6425887522bbcd54e98132/arrows/ocv/tests/test_inpaint.cxx#L83). You will need to copy the test image to `data` in the test directory.

chetnieter (changes requested):
Looking good. The biggest issue is fixing how to pass the file path of the test image to the tests. If done correctly you should be able to run `ctest -R colmap` in your build and it will run all the colmap tests.

arrows/colmap/tests/test_image.cxx
```diff
@@ -0,0 +1,70 @@
+#include <arrows/colmap/colmap.h>
```
chetnieter, line 1: This need to change to

```
#include <arrows/colmap/image_container.h>
```

chetnieter (changes requested):

chetnieter:
Looks good - good work!

PR #1365, opened 2021-07-09
minimize re-projection error over multiple focal scales

borovik135:
This small change makes our camera calibration more robust to initial focal lengths, and naturally slows it down a bit.

### DIFF
--- a/arrows/ocv/resection_camera.cxx
+++ b/arrows/ocv/resection_camera.cxx
@@ -211,10 +211,12 @@ resection_camera
   MatD cv_K;
   eigen2cv( K, cv_K );
   auto const dc0 = dist_coeffs;
-  auto const focal_scales = {0.5, 1.0, 2.0};
-  auto focal_scale = 0.0; // preferred focal scale after optimization
+  auto const focal_scales = { 0.5, 1.0, 2.0 };
+  // focal scale search parameter for optimization
+  auto focal_scale = 0.0;
+  // minimize re-projection error over multiple focal scales
   for (auto const scale : focal_scales)
-  { // minimize re-projection error over multiple focal scales
+  {
     auto dc = dc0;
     vmat rv,tv;
     MatD cvK;
@@ -236,7 +238,7 @@ resection_camera
     }
   }
 
-  LOG_INFO( d_->m_logger, "re-projection error=" << err <<
+  LOG_DEBUG( d_->m_logger, "re-projection error=" << err <<
 	    ", focal scale=" << focal_scale);
   if( err > reproj_error )
   {
--- a/arrows/ocv/resection_camera.cxx
+++ b/arrows/ocv/resection_camera.cxx
@@ -105,7 +105,8 @@ resection_camera
                                                      d_->reproj_accuracy );
   d_->max_iterations = config->get_value< int >( "max_iterations",
                                                  d_->max_iterations );
-  std::stringstream ss(config->get_value< std::string > ( "focal_scales", "1" ));
+  std::stringstream ss( config->get_value< std::string > ( "focal_scales", "1" ) );
+  d_->focal_scales.clear();
   ss >> d_->focal_scales;
 }
 
@@ -114,7 +115,6 @@ bool
 resection_camera
 ::check_configuration( vital::config_block_sptr config ) const
 {
-  auto good_conf = true;
   auto const reproj_accuracy =
     config->get_value< double >( "reproj_accuracy", d_->reproj_accuracy );
   if( reproj_accuracy <= 0.0 )
@@ -122,7 +122,7 @@ resection_camera
     LOG_ERROR( d_->m_logger,
                "reproj_accuracy parameter is " << reproj_accuracy <<
                ", but needs to be positive." );
-    good_conf = false;
+    return false;
   }
 
   auto const max_iterations =
@@ -132,10 +132,22 @@ resection_camera
     LOG_ERROR( d_->m_logger,
                "max iterations is " << max_iterations <<
                ", needs to be greater than zero." );
-    good_conf = false;
+    return false;
   }
-  // TODO: focal_scales
-  return good_conf;
+
+  std::stringstream ss( config->get_value< std::string > ( "focal_scales", "1" ) );
+  vectorf focal_scales;
+  ss >> focal_scales;
+  auto m = std::min_element( focal_scales.begin(), focal_scales.end() );
+  if( m == focal_scales.end() || *m <= 0 )
+  {
+    LOG_ERROR( d_->m_logger,
+               "focal_scales: " << focal_scales <<
+               ", needs to be greater than zero." );
+    return false;
+  }
+
+  return true;
 }
 
 // ----------------------------------------------------------------------------
@@ -235,9 +247,6 @@ resection_camera
     flags |= cv::CALIB_FIX_K4 | cv::CALIB_FIX_K5 | cv::CALIB_FIX_K6;
   }
 
-  auto const reproj_error = d_->reproj_accuracy;
-
-  auto err = std::numeric_limits< double >::infinity();
   vital::matrix_3x3d K = cal->as_matrix();
   cv::TermCriteria term_criteria{
     cv::TermCriteria::COUNT + cv::TermCriteria::EPS,
@@ -251,6 +260,7 @@ resection_camera
   // focal scale search parameter for optimization
   auto focal_scale = 1.0;
   // minimize re-projection error over multiple focal scales
+  auto err = std::numeric_limits< double >::infinity();
   for( auto const scale : d_->focal_scales )
   {
     auto dc = dc0;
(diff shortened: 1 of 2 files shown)


dstoup:
Jenkins test this please

borovik135:
it would be nice to merge this into master, as this small change significantly improves our camera calibration function

arrows/ocv/resection_camera.cxx
```diff
@@ -205,15 +202,42 @@ ::resection(
...
+      cv_K = cvK;
+      dist_coeffs = dc;
+      vrvec = rv;
+      vtvec = tv;
+      focal_scale = scale;
+      err = e;
+    }
+  }
 
+  LOG_INFO( d_->m_logger, "re-projection error=" << err <<
```
mleotta, line 239: I'd suggest that this be `LOG_DEBUG` instead of `LOG_INFO`.

arrows/ocv/resection_camera.cxx
```diff
@@ -205,15 +202,42 @@ ::resection(
...
+  vital::matrix_3x3d K = cal->as_matrix();
+  cv::TermCriteria term_criteria{
+    cv::TermCriteria::COUNT + cv::TermCriteria::EPS,
+    d_->max_iterations, DBL_EPSILON };
+  using MatD = cv::Mat_<double>;
+  MatD cv_K;
+  eigen2cv( K, cv_K );
+  auto const dc0 = dist_coeffs;
+  auto const focal_scales = {0.5, 1.0, 2.0};
+  auto focal_scale = 0.0; // preferred focal scale after optimization
```
mleotta, line 215: I have a preference for comments on the line before a statement rather than at the end of a line after the statement, but it seems our style guide doesn't comment on this, so I guess it's okay.  The style guide does say to start the comment with a capital letter.

arrows/ocv/resection_camera.cxx
```diff
@@ -205,15 +202,42 @@ ::resection(
...
+  auto err = std::numeric_limits<double>::infinity();
+  vital::matrix_3x3d K = cal->as_matrix();
+  cv::TermCriteria term_criteria{
+    cv::TermCriteria::COUNT + cv::TermCriteria::EPS,
+    d_->max_iterations, DBL_EPSILON };
+  using MatD = cv::Mat_<double>;
+  MatD cv_K;
+  eigen2cv( K, cv_K );
+  auto const dc0 = dist_coeffs;
+  auto const focal_scales = {0.5, 1.0, 2.0};
```
mleotta, line 214: Are these three scale always sufficient?  Not critical, but it might be nice to make this list a config parameter at some point (doesn't need to hold up this PR).

arrows/ocv/resection_camera.cxx
```diff
@@ -205,15 +202,42 @@ ::resection(
...
+  auto focal_scale = 0.0; // preferred focal scale after optimization
+  for (auto const scale : focal_scales)
+  { // minimize re-projection error over multiple focal scales
```
mleotta, line 217: same applies to this comment

mleotta (changes requested):
Looks good except for a few minor nit-picks

arrows/ocv/resection_camera.cxx
borovik135: done

arrows/ocv/resection_camera.cxx
borovik135: done

arrows/ocv/resection_camera.cxx
borovik135: those values appear to be sufficient for the projects where we use manual calibration; having them in a configuration file may indeed be a good idea; I may need some guidance on how to represent a vector of floats in a config file

arrows/ocv/resection_camera.cxx
borovik135: done

borovik135:
> Looks good except for a few minor nit-picks

I believe I have addressed those minor nit-picks; can we merge this PR into master?

mleotta:
Jenkins test this please

arrows/ocv/resection_camera.cxx
```diff
@@ -58,9 +62,10 @@ ::get_configuration() const
     vital::algo::resection_camera::get_configuration();
   d_->get_configuration( config );
   config->set_value( "reproj_accuracy", d_->reproj_accuracy,
-                     "desired re-projection positive accuracy" );
+                     "desired re-projection positive accuracy for inlier points" );
   config->set_value( "max_iterations", d_->max_iterations,
                      "maximum number of iterations to run optimization [1, INT_MAX]" );
+  // TODO: focal_scales
```
mleotta, line 68: Adding options to set the actual focal scales would be nice, but at a minimum, I think what this branch needs for now is a way to disable the use of focal scales through configuration.  It is often the case that you know the intrinsic model and want the best resectioned camera using that exact intrinsic model, without allowing the focal length to change.  So we need a way of turning off the multiple focal scales.  If we can just set the focal scales in the config then we could set this with just one scale of 1.0.  If that's too trick for now, we should at least have a boolean option to enable or disable these scales.

borovik135: When discussed with BC and ML this changed seemed like a good idea for our BA needs, which currently relies just on trusted tracks, but still uses the notion of fixed landmarks which are supplied via an index. Given no other needs, I suppose this change can be dropped.

borovik135 (commented):
I think these changes have been overcome by the newer commits.

borovik135:
@mleotta , I think this PR is ready to land

arrows/ocv/resection_camera.cxx
```diff
@@ -74,22 +105,24 @@ ::set_configuration( vital::config_block_sptr config )
...
+  d_->focal_scales.clear();
+  ss >> d_->focal_scales;
 }
 
 // ----------------------------------------------------------------------------
 bool
 resection_camera
 ::check_configuration( vital::config_block_sptr config ) const
 {
-  auto good_conf = true;
```
mleotta, line 84: In `check_configuration` it is good practice to do all the checks before returning false if one failed.  The reason for this is that a config file may have multiple errors and we want all of those errors logged before failing.  Otherwise, the user will just see the first error, fix it, run again, and then see the second error, fix it, run again, etc.  It can take many more iterations to find all the problems.  It's better just to report all the errors you can find.

arrows/ocv/resection_camera.cxx
```diff
@@ -99,9 +132,22 @@ ::check_configuration( vital::config_block_sptr config ) const
...
                ", needs to be greater than zero." );
-    good_conf = false;
+    return false;
+  }
+
+  std::stringstream ss( config->get_value< std::string > ( "focal_scales", "1" ) );
+  vectorf focal_scales;
+  ss >> focal_scales;
+  auto m = std::min_element( focal_scales.begin(), focal_scales.end() );
+  if( m == focal_scales.end() || *m <= 0 )
```
mleotta, line 142: I think `focal_scales` being empty should generate a different error message than the error for values less than zero.

I'm also wondering if we should require that "1" be in the list.  Or maybe this should be a list of _additional_ focal scales and "1" is just implicitly assumed to be always in the list.  In that case if the list was empty by default you would just get "1" and if the list was "0.5 2" you would end up using scales of 0.5, 1, and 2.

borovik135: We now ensure that focal_scales is never empty, by default containing 1.0, but allowing custom lists specified in a configuration, including the case omitting 1.0 for some special cases, when 1.0 is not of interest, thus achieving the most flexibility.

borovik135:
I think all of the change requests have been addressed and checks passed, so it is probably time to land this PR.

arrows/ocv/resection_camera.cxx
```diff
@@ -101,6 +140,21 @@ ::check_configuration( vital::config_block_sptr config ) const
...
     good_conf = false;
   }
+
+  std::stringstream ss( config->get_value< std::string >( "focal_scales",
+                                                          "1" ) );
+  vectorf focal_scales;
+  ss >> focal_scales;
+
+  auto m = std::min_element( focal_scales.begin(), focal_scales.end() );
+  if( *m <= 0 )
```
mleotta, line 150: If `focal_scales` is empty then then this will dereference `focal_scales.end()`, which will likely crash.  You should check if `focal_scales` is empty.  It should be a different error message if the `focal_scales` are empty.  We need at least one positive value unless you are going to treat `1` as a special value that is always implicitly in the list.  However, currently that is not the case that 1 is always forced to be in the list.

mleotta (changes requested):
There are still issues related to ensuring that you have valid focal scales specified.

arrows/ocv/resection_camera.cxx
borovik135: I think the prior line
std::stringstream ss( config->get_value< std::string >( "focal_scales", "1" ) );
ensures that we always have at least one value for focal_scales, so it is never empty; should I leave a comment about that in the code?

mleotta: Not exactly.  That "1" provides a default value if the option is unspecified.  If the option is specified, but is left blank then you will still get an empty string.  Or if the values specified are not a valid list of number the stringstream may give you an empty vector.

borovik135: now requiring that focal_scales is non-empty

mleotta:
Jenkins test this please

PR #1368, opened 2021-07-13
Uncrustified vital/algo (1)

quantumech3:


dstoup:
Jenkins test this please

dstoup (commented):
Looks fine to me as long as CI runs pass. I am a bit confused as to why some of the "// end namespace" comments are indented more than the other comments. I am guessing it happens in the cases where the former end was " } } }  // end namespace". Guessing uncrustify is just replacing the two "} } with a space. Might be worth just manually fixing them if you have a minute.

PR #1369, opened 2021-07-16
Refactor std_0102_lds

daniel-riehm:
Consolidate two locally-declared versions of the std_0102_lds class,
which remains empty and unimplemented, into a new metadata_types file.
This new file is intended to eventually contain a number of new classes
and enums during future refactoring commits.


daniel-riehm:
jenkins test this please

arrows/klv/convert_0104_metadata.cxx
```diff
@@ -207,7 +208,7 @@ case klv_0104::KN:                                        \
       CASE( START_DATE_TIME_UTC );
       CASE2( MISSION_START_TIME, MISSION_START_TIME_UTC );
       CASE( SECURITY_CLASSIFICATION );
-      CASE( CLASSIFICATION );
```
russelldj, line 210: Lacking context here but what happened to this classification tag?

vital/tests/test_metadata_types.cxx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// ----------------------------------------------------------------------------
+TEST( metadata, std_0102_lds )
+{
+  auto const obj = std_0102_lds{};
+  std::stringstream ss;
+  ss << obj;
+  // When std_0102_lds is actually implemented, this will fail as a reminder
+  //   to write an actual unit test in this file
```
russelldj, line 28: very minor nitpick. Weird indent.

russelldj (approved):
I think I need some additional context to wholeheartedly approve this but none of the changes seem dangerous. I think what would be most useful is a better understanding at a very high level of what 0102 is. I'd assume it's another NGA standard? Also, is `classification` just an alias for `security_local_md_set` or is it just happening to get deleted in this PR?

arrows/klv/convert_0104_metadata.cxx
daniel-riehm: It was a duplicate of `SECURITY_LOCAL_MD_SET`. It appears in both the 0104 (as `CLASSIFICATION`) and 0601 (as `SECURITY_LOCAL_MD_SET`) standards and I guess whoever wrote them didn't realize this at the time. I chose to stick with the `SECURITY_LOCAL_MD_SET` name since it's more descriptive.

russelldj: Thanks, makes sense.

daniel-riehm:
@russelldj Requested context: 0102 is indeed another MISB KLV standard (just like e.g. 0601) found [here](https://gwg.nga.mil/misb/docs/standards/ST0102.12.pdf). Its tags all have to do with security classification - who should be able to view this footage, is it top secret, etc. It's possible to embed this 0102 KLV set inside an 0601 or 0104 set as just another one of the values. The actual parsing of the 0102 standard has not been implemented yet, but it should be identical no matter where it is embedded. Before this PR, there were two separate KWIVER tag names and two separate classes for the 0102 data, one for each of (0601, 0104). This PR pulls them out and combines them into a single class in a new file, with the `SECURITY_LOCAL_MD_SET` absorbing the identical `CLASSIFICATION` tag.

russelldj:
Ok thanks, this all seems good. My only frustration is the name they use, `security metadata local set`, is hard to understand at first glance, but I understand the rationale for sticking with the standard.

daniel-riehm:
> Ok thanks, this all seems good. My only frustration is the name they use, `security metadata local set`, is hard to understand at first glance, but I understand the rationale for sticking with the standard.

Do you have a suggested replacement? I have no problem with changing the tag name from what it is now.

russelldj:
From my initial read, `security_{classification, designation}`, or `clearance_{classification, designation}` seem reasonable, but you know better than me. I think including the `metadata` part is confusing when it's a metadata tag. And I again could be wrong, but `local set` seems to refer to an implementation detail?

daniel-riehm:
What about `security_specification`? `security_classification` is actually the name of a field within 0102, which could be confusing.

russelldj:
That sounds good.

daniel-riehm:
Ready to land, then?

russelldj:
> Ready to land, then?

Sounds good to me.

PR #1373, opened 2021-07-26
Implement reading and writing KLV integer formats

daniel-riehm:
Create flexible, robust, and error-checking functions for reading and writing three KLV integer formats. Begin the much-needed process of writing KLV unit tests here. Add two new exception types to differentiate the types of runtime errors one might run across when dealing with low-level KLV reading and writing.

Floating-point and string readers/writers soon to follow.

### DIFF
--- a/arrows/klv/CMakeLists.txt
+++ b/arrows/klv/CMakeLists.txt
@@ -6,9 +6,9 @@ set( sources
   convert_metadata.cxx
   convert_0601_metadata.cxx
   convert_0104_metadata.cxx
-  klv_parse.cxx
   klv_data.cxx
   klv_key.cxx
+  klv_parse.cxx
   klv_0601.cxx
   klv_0104.cxx
   misp_time.cxx
@@ -16,9 +16,9 @@ set( sources
 
 set( public_headers
   convert_metadata.h
-  klv_parse.h
   klv_data.h
   klv_key.h
+  klv_parse.h
   klv_0601.h
   klv_0104.h
   misp_time.h
--- a/arrows/klv/klv_read_write.txx
+++ b/arrows/klv/klv_read_write.txx
@@ -1,0 +1,348 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/// \file
+/// Implementation of the templated basic KLV read/write functions.
+
+#include "klv_read_write.h"
+
+#include <vital/exceptions.h>
+
+#include <limits>
+#include <numeric>
+#include <string>
+
+#include <cmath>
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace klv {
+
+// ---------------------------------------------------------------------------
+// Macros for verifying type attributes
+// Message-less static_assert not available in C++11
+
+#define KLV_ASSERT_INT( T ) \
+  static_assert( std::is_integral< T >::value, "must be an integer type" )
+
+#define KLV_ASSERT_UINT( T )                                                \
+  KLV_ASSERT_INT( T );                                                      \
+  static_assert( std::is_unsigned< T >::value, "must be an unsigned type" )
+
+#define KLV_ASSERT_SINT( T )                                           \
+  KLV_ASSERT_INT( T );                                                 \
+  static_assert( std::is_signed< T >::value, "must be a signed type" )
+
+#define KLV_ASSERT_UINT8_ITERATOR( ITER )                              \
+  static_assert(                                                       \
+    std::is_same< typename std::decay< decltype( *ITER ) >::type,      \
+                  uint8_t >::value, "iterator must point to uint8_t" )
+
+// ---------------------------------------------------------------------------
+// Return number of bits required to store the given signed or unsigned int.
+template < class T >
+KWIVER_ALGO_KLV_EXPORT
+size_t
+_int_bit_length( T value );
+
+// ---------------------------------------------------------------------------
+// Return whether left-shifting by given amount would overflow type T.
+template < int shift_amount, typename T >
+KWIVER_ALGO_KLV_EXPORT
+bool
+_left_shift_overflow( T value );
+
+// ---------------------------------------------------------------------------
+template < class T, class Iterator >
+T
+klv_read_int( Iterator& data, size_t length )
+{
+  // Ensure types are compatible with our assumptions
+  KLV_ASSERT_INT( T );
+  KLV_ASSERT_UINT8_ITERATOR( data );
+
+  if( sizeof( T ) < length )
+  {
+    VITAL_THROW( kwiver::vital::metadata_type_overflow,
+                 "integer will overflow given type" );
+  }
+
+  // Avoid integer underflow in later expressions
+  if( !length )
+  {
+    return 0;
+  }
+
+  // We have to work with unsigned for proper behavior of << >> operators
+  using UnsignedT = typename std::make_unsigned< T >::type;
+
+  // Functor to insert each successive byte into the output value
+  auto const accumulator = []( UnsignedT value, uint8_t byte ){
+                             return ( value << 8 ) | byte;
+                           };
+
+  // Reduce span to final result
+  auto result =
+    std::accumulate( data, data + length, static_cast< UnsignedT >( 0 ),
+                     accumulator );
+
+  // Extend sign bit
+  UnsignedT const result_sign_bit = 1ull << ( 8 * length - 1 );
+  if( std::is_signed< T >::value && sizeof( T ) != length &&
+      ( result_sign_bit & result ) )
+  {
+    result |= ~UnsignedT{ 0 } << ( 8 * length );
+  }
+
+  data += length;
+
+  return result;
+}
+
+// ---------------------------------------------------------------------------
+template < class T, class Iterator >
+void
+klv_write_int( T value, Iterator& data, size_t length )
+{
+  // Ensure types are compatible with our assumptions
+  KLV_ASSERT_INT( T );
+  KLV_ASSERT_UINT8_ITERATOR( data );
+
+  auto const value_length = klv_int_length( value );
+  if( value_length > length )
+  {
+    VITAL_THROW( kwiver::vital::metadata_type_overflow,
+                 "integer not representable using given length" );
+  }
+
+  using UnsignedT = typename std::make_unsigned< T >::type;
+
+  auto const unsigned_value = static_cast< UnsignedT >( value );
+
+  for( size_t i = 0; i < length; ++i, ++data )
+  {
+    auto const shift_amount = ( length - i - 1 ) * 8;
+    *data =
+      static_cast< uint8_t >( 0xFF & ( unsigned_value >> shift_amount ) );
+  }
+}
+
+// ---------------------------------------------------------------------------
+template < class T >
+size_t
+klv_int_length( T value )
+{
+  return ( _int_bit_length( value ) + 7 ) / 8;
+}
+
+// ---------------------------------------------------------------------------
+template < class T, class Iterator >
+T
+klv_read_ber( Iterator& data, size_t max_length )
+{
+  // Ensure types are compatible with our assumptions
+  KLV_ASSERT_UINT( T );
+  KLV_ASSERT_UINT8_ITERATOR( data );
+
+  if( !max_length )
+  {
+    VITAL_THROW( kwiver::vital::metadata_buffer_overflow,
+                 "BER decoding overruns end of data buffer" );
+  }
+
+  // Short form - first bit is 0, remaining bits are the value itself
+  if( !( 0x80 & *data ) )
+  {
+    auto const result = *data;
+    ++data;
+    return result;
+  }
+
+  // Long form - first bit is 1, remaining bits are the length of the
+  // following
+  size_t const total_length = ( 0x7F & *data ) + 1;
+
+  if( total_length > max_length )
+  {
+    VITAL_THROW( kwiver::vital::metadata_buffer_overflow,
+                 "BER decoding overruns end of data buffer" );
+  }
+
+  auto const rewind = data;
+  try
+  {
+    return klv_read_int< T >( ++data, total_length - 1 );
+  }
+  catch ( const kwiver::vital::metadata_type_overflow& e )
+  {
+    data = rewind;
+
+    throw e;
+  }
+}
+
+// ---------------------------------------------------------------------------
+template < class T, class Iterator >
+void
+klv_write_ber( T value, Iterator& data, size_t max_length )
+{
+  KLV_ASSERT_UINT( T );
+  KLV_ASSERT_UINT8_ITERATOR( data );
+
+  auto const value_length = klv_ber_length( value );
+  if( value_length > max_length )
+  {
+    VITAL_THROW( kwiver::vital::metadata_buffer_overflow,
+                 "BER encoding overruns end of data buffer" );
+  }
+
+  if( value < 128 )
+  {
+    *data = static_cast< uint8_t >( value );
+    ++data;
+  }
+  else
+  {
+    *data = static_cast< uint8_t >( ( 0x7F & value_length - 1 ) | 0x80 );
+    klv_write_int( value, ++data, value_length - 1 );
+  }
+}
+
+// ---------------------------------------------------------------------------
+template < class T >
+size_t
+klv_ber_length( T value )
+{
+  KLV_ASSERT_UINT( T );
+  return ( value > 127 ) ? klv_int_length( value ) + 1 : 1;
+}
+
+// ---------------------------------------------------------------------------
+template < class T, class Iterator >
+T
+klv_read_ber_oid( Iterator& data, size_t max_length )
+{
+  KLV_ASSERT_UINT( T );
+  KLV_ASSERT_UINT8_ITERATOR( data );
+
+  auto const rewind = data;
+
+  // Highest bit set = not final byte
+  T value = 0;
+  for( uint8_t prev_byte = 0x80; prev_byte & 0x80; ++data )
+  {
+    if( !max_length )
+    {
+      data = rewind;
+      VITAL_THROW( kwiver::vital::metadata_buffer_overflow,
+                   "BER-OID decoding overruns end of data buffer" );
+    }
+    --max_length;
+
+    if( _left_shift_overflow< 7, T >( value ) )
+    {
+      data = rewind;
+      VITAL_THROW( kwiver::vital::metadata_type_overflow,
+                   "BER-OID value will overflow given type" );
+    }
+
+    value <<= 7;
+    value |= 0x7F & ( prev_byte = *data );
+  }
+
+  return value;
+}
+
+// ---------------------------------------------------------------------------
+template < class T, class Iterator >
+void
+klv_write_ber_oid( T value, Iterator& data, size_t max_length )
+{
+  KLV_ASSERT_UINT( T );
+  KLV_ASSERT_UINT8_ITERATOR( data );
+
+  auto value_length = klv_ber_oid_length( value );
+  if( value_length > max_length )
+  {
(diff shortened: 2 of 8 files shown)


daniel-riehm:
jenkins test this please

arrows/klv/CMakeLists.txt
```diff
@@ -6,19 +6,21 @@ set( sources
...
   )
 
 set( public_headers
   convert_metadata.h
-  klv_parse.h
   klv_data.h
   klv_key.h
+  klv_parse.h
+  klv_read_write.h
+  klv_read_write.txx
```
mwoehlke-kitware, line 23: Is this really a *public* header?

mwoehlke-kitware (commented):
I'm going to punt on a more detailed review for now, and instead, can you run Uncrustify on the new files?

daniel-riehm:
> I'm going to punt on a more detailed review for now, and instead, can you run Uncrustify on the new files?

Done. Is any work being done on the uncrustify that I should pull from somewhere? It still leaves / causes some amount of crust in need of manual cleanup.

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,378 @@
...
+  KLV_ASSERT_INT( T );                                                 \
+  static_assert( std::is_signed < T > ::value, "must be a signed type" )
+
+#define KLV_ASSERT_UINT8_ITERATOR( ITER )                              \
+  static_assert(                                                       \
+    std::is_same < typename std::decay < decltype( *ITER ) > ::type,      \
+    uint8_t > ::value, "iterator must point to uint8_t" )
+
+// ---------------------------------------------------------------------------
+// Returns number of bits required to store the given signed or unsigned int.
```
russelldj, line 42: I'm not saying this is wrong, but the full indents before code is surprising to me._

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,378 @@
...
+                                    ( --value_length * 7 );
+          auto const top_bit = static_cast < uint8_t >
+                               ( value_length ? 0x80 : 0x00 );
+          auto const bottom_bits =
+            static_cast < uint8_t > ( ( value >> shift_amount ) & 0x7F );
+          *( data++ ) = top_bit | bottom_bits;
+        }
+      }
+
+// ---------------------------------------------------------------------------
```
russelldj, line 306: I'm always surprised there's no stdlib implementation for this.

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,437 @@
...
+  KLV_WRITE_INT( int64_t,  8, 0x1122334455667788 )
+
+  // Signed, negative - arbitrary numbers
+  KLV_WRITE_INT( int8_t,   1, -0x11 )
+  KLV_WRITE_INT( int16_t,  2, -0x1122 )
+  KLV_WRITE_INT( int32_t,  4, -0x11223344 )
+  KLV_WRITE_INT( int64_t,  8, -0x1122334455667788 )
+
+  // Lowest representable value
+  KLV_WRITE_INT( uint8_t,  1, std::numeric_limits< uint8_t  >::min() )
```
russelldj, line 138: There's an internal email thread about `std::numeric_limits` not behaving in an intuitive fashion but I forget whether it's at all applicable here. I'll try to pull it up after the review.

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,437 @@
...
+#define KLV_READ_BER_ERR( T, length, e, ... )                 \
+  {                                                           \
+    vec_t const data = { __VA_ARGS__ };                       \
+    auto it = data.begin();                                   \
+    EXPECT_THROW( klv_read_ber< T >( it, length ), e );       \
+    EXPECT_EQ( data.begin(), it );                            \
+  }
+
+  // Short form
+  KLV_READ_BER( uint8_t, 0,
```
russelldj, line 217: Odd formatting

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,437 @@
...
+{
+#define KLV_ROUND_TRIP( value, length )                              \
+  {                                                                  \
+    auto data = vec_t( length, 0xba );                               \
+    auto it = data.begin();                                          \
+    klv_write_ber( value, it, data.size() );                         \
+    EXPECT_EQ( it, data.end() );                                     \
+    try                                                              \
+    {                                                                \
+      EXPECT_NE( 128, data.at( 0 ) );                                \
```
russelldj, line 265: Maybe deserving of a comment

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,437 @@
...
+  KLV_WRITE_BER_ERR( 0ull,                                          0,
+                     kv::metadata_buffer_overflow )
+  KLV_WRITE_BER_ERR( 50ull,                                         0,
+                     kv::metadata_buffer_overflow )
+  KLV_WRITE_BER_ERR( ( 1ull << 7 ) - 1,                             0,
+                     kv::metadata_buffer_overflow )
+  KLV_WRITE_BER_ERR( ( 1ull << 7 ),                                 1,
+                     kv::metadata_buffer_overflow )
+  KLV_WRITE_BER_ERR(
+    std::numeric_limits< uint8_t  >::max(),        1,
```
russelldj, line 304: nitpick: double space here and below after `uint8_t`

arrows/klv/tests/test_klv_read_write.cxx
russelldj: It looks like `std::numeric_limits` returns odd values for floating points. The correct solution for floating points is `std::numeric_limits<double>::lowest()`. But this appears irrelevant here.

russelldj (approved):
Looks good to me.

daniel-riehm:
One thing I'd like input on from reviewers is whether the macro-based approach in the unit tests is the right way to go - the alternative being (sometimes templated) functions in lieu of macros. The advantages of macros are that the test case error message points to the line of the test case, not the line in the test function, and that their parameters are more flexible - e.g. insert a substring into the name of a function. The function-based approach would be less ugly, more C++ style-wise, work better with IDEs, allow unparenthesized commas in arguments, and avoid the finicky-ness of all those line terminators. Should I switch the unit tests over, or keep them as they are?

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,378 @@
...
+
+#include <limits>
+#include <numeric>
+#include <string>
+
+#include <cmath>
+
+namespace kwiver {
+  namespace arrows {
+    namespace klv {
```
mwoehlke-kitware, line 20: This... looks like Uncrustify treated this as C, not C++?

russelldj:
I'll defer on this one to others, but I think that macros are perfectly acceptable in this case. When I've written functions, I've hit the exact issue you mentioned about the incorrect line number showing up, which I found rather annoying. Since you have this implementation already, it seems reasonable to keep it this way.

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,378 @@
...
+  static_assert(                                                       \
+    std::is_same < typename std::decay < decltype( *ITER ) > ::type,      \
+    uint8_t > ::value, "iterator must point to uint8_t" )
+
+// ---------------------------------------------------------------------------
+// Returns number of bits required to store the given signed or unsigned int.
+      template < class T >
+      KWIVER_ALGO_KLV_EXPORT
+      size_t
+      _int_bit_requirement( T value );
```
mwoehlke-kitware, line 46: What is the purpose of forward declarations? Why do these start with `_`? If they're meant to be internal, putting them in an anonymous namespace (and also a `.cxx`) might be better?

arrows/klv/tests/test_klv_read_write.cxx
mwoehlke-kitware: This looks like a case of Uncrustify confused by the lack of a `;`? Maybe consider rewriting the macros wrapped with `do ... while(0)`.

mwoehlke-kitware:
An alternative to using purely macros would be to have *one* macro, maybe like:
```c++
#define CALL_TEST( func, ... ) \
  do { SCOPED_TRACE( #func ); func( __VA_ARGS__ ); } while( 0 )
```

arrows/klv/klv_read_write.txx
daniel-riehm: ...Huh. I suppose forward declarations aren't necessary for use within template functions. My bad.

I don't think putting them in a `.cxx` would work because they're templated and used within templates. Doesn't the compiler need the full template definition in the header? You are correct that the `_` was meant to mark them as internal, since (correct me if I'm wrong here) an anonymous namespace wouldn't make sense in a header. An alternative would be explicitly instantiating the helper templates for all foreseen types, then moving the definitions into a `.cxx`, but that seems less elegant to me.

arrows/klv/tests/test_klv_read_write.cxx
daniel-riehm: Thanks, I'll keep that in mind for the floating-point readers/writers.

arrows/klv/klv_read_write.txx
daniel-riehm: Turns out that removing forward declarations does cause a compile error. Even with templates, functions need to be declared before use.

daniel-riehm:
Okay, I think I've addressed all the review comments. Let me know if anything else needs to change.

arrows/klv/klv_read_write.txx
mwoehlke-kitware: Let me rephrase that, then; is there a reason your definitions aren't in order of dependence? 🙂

daniel-riehm:
I suppose I should add tests for the newly-public functions. I will push those in a bit.

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,346 @@
...
+
+  if( _left_shift_overflow< T >( value, 7 ) )
+  {
+    VITAL_THROW( kwiver::vital::metadata_type_overflow,
+                 "BER-OID value will overflow given type" );
+  }
+
+  // Highest bit unset = final byte
+  value <<= 7;
+  value |= *( data++ );
```
mwoehlke-kitware, line 258: Somewhat weak suggestion:

```suggestion
  do
  {
    if( !max_length )
    {
      // TODO restore iterator
      VITAL_THROW( kwiver::vital::metadata_buffer_overflow,
                   "BER-OID decoding overruns end of data buffer" );
    }
    --max_length;

    if( _left_shift_overflow< T >( value, 7 ) )
    {
      // TODO restore iterator
      VITAL_THROW( kwiver::vital::metadata_type_overflow,
                   "BER-OID value will overflow given type" );
    }
    value <<= 7;
    value |= 0x7F & *data;
    
    // Highest bit unset = final byte
  } while( 0x80 & *( data++ ) )
```

Please double-check my logic 🙂. Also, there are various ways this could be done (e.g. still using a `for` or using a `while( true )`); the idea is mainly to avoid having separate code paths to add the last byte vs. other bytes.

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,346 @@
...
+  auto value_length = klv_ber_oid_length( value );
+  if( value_length > max_length )
+  {
+    VITAL_THROW( kwiver::vital::metadata_buffer_overflow,
+                 "BER-OID encoding overruns end of data buffer" );
+  }
+
+  if( !value )
+  {
+    *( data++ ) = 0;
```
mwoehlke-kitware, line 280:
```suggestion
    *data = 0;
    ++data
```

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,346 @@
...
+    return;
+  }
+
+  while( value_length )
+  {
+    auto const shift_amount = static_cast< uint8_t >( --value_length * 7 );
+    auto const top_bit = static_cast< uint8_t >( value_length ? 0x80 : 0x00 );
+    auto const bottom_bits =
+      static_cast< uint8_t >( ( value >> shift_amount ) & 0x7F );
+    *( data++ ) = top_bit | bottom_bits;
```
mwoehlke-kitware, line 290:
```suggestion
    *data = top_bit | bottom_bits;
    ++data;
```

mwoehlke-kitware (commented):
Have not re-reviewed the unit tests, but wanted to get this published...

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,346 @@
...
+  return i;
+}
+
+// ---------------------------------------------------------------------------
+template < class T >
+bool
+_left_shift_overflow( T value, T shift_amount )
+{
+  KLV_ASSERT_UINT( T );
+  return static_cast< T >( value << shift_amount ) >> shift_amount != value;
```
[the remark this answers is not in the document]
daniel-riehm: As per https://en.cppreference.com/w/cpp/language/implicit_conversion#Integral_promotion, both arguments are implicitly promoted to `int`. For types smaller than `int` (e.g. `uint8_t`), this means that what would normally overflow does not, and the test returns the wrong result. The `static_cast` thus enforces the overflow true to `T`.

Your `constexpr` idea is maybe a smarter way of doing this anyway.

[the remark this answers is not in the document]
mwoehlke-kitware: D'oh, right, I was forgetting about this possibly *truncating*.

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,626 @@
...
+test_write_ber_oid( T value, size_t length )
+{
+  auto data = vec_t( length, 0xba );
+  auto it = data.begin();
+  klv_write_ber_oid< T >( value, it, length );
+  EXPECT_EQ( data.end(), it );
+  // Avoid technically correct but bad-form encoding ( leading zero bytes )
+  EXPECT_NE( data[ 0 ], 0x80 );
+  it = data.begin();
+  EXPECT_NE( 0x80, *it );
```
mwoehlke-kitware, line 553: These two expects seem redundant?

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,626 @@
...
+  CALL_TEST( test_read_ber< uint8_t >, 0x42, { 0x42 } );
+  CALL_TEST( test_read_ber< uint8_t >, 0x7F, { 0x7F } );
+
+  // Long form
+  CALL_TEST( test_read_ber< uint8_t >,  0xFF,       { 0x81, 0xFF } );
+  CALL_TEST( test_read_ber< uint16_t >, 0x102,      { 0x82, 0x01, 0x02 } );
+  CALL_TEST( test_read_ber< uint32_t >, 0x010203,
+             { 0x83, 0x01, 0x02, 0x03 } );
+  CALL_TEST( test_read_ber< uint32_t >, 0xFF428012, { 0x84, 0xFF, 0x42, 0x80,
+                                                      0x12 } );
```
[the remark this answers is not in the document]
daniel-riehm: This is what uncrustify produces - I will remember to look for this in later commits.

arrows/klv/tests/test_klv_read_write.cxx
daniel-riehm: Oops - good catch

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,626 @@
...
+
+  // Truncated values
+  CALL_TEST( test_read_ber_oid_buffer_overflow< uint64_t >, {} );
+  CALL_TEST( test_read_ber_oid_buffer_overflow< uint64_t >, { 0x81 } );
+  CALL_TEST( test_read_ber_oid_buffer_overflow< uint64_t >, { 0xFF, 0x81 } );
+
+  // Values too large for native type
+  CALL_TEST( test_read_ber_oid_type_overflow< uint8_t  >, { 0x82, 0X00 } );
+  CALL_TEST( test_read_ber_oid_type_overflow< uint16_t >,
+             { 0x84, 0x81, 0X00 } );
```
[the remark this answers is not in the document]
daniel-riehm: No - probably a typo

arrows/klv/klv_read_write.txx
```diff
@@ -0,0 +1,346 @@
...
+}
```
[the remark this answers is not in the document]
daniel-riehm: I don't know if it's a bug or what, but on GCC 9.3.0 I get `error: left operand of shift expression ‘(-1 << 1)’ is negative` on line 343.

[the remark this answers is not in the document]
daniel-riehm: (Which is why I didn't push an implementation it the first time you suggested it)

arrows/klv/tests/test_klv_read_write.cxx
```diff
@@ -0,0 +1,626 @@
...
+  EXPECT_EQ( it, data.cbegin() + offset );
+}
+
+// ----------------------------------------------------------------------------
+template < class T >
+void
+test_read_int_type_overflow( vec_t const& data, size_t offset, size_t length )
+{
+  test_read_int_throw< T, kv::metadata_type_overflow >( data, offset, length );
+}
```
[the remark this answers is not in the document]
daniel-riehm: Yeah, I might keep it this way just for the consistency.

arrows/klv/tests/test_klv_read_write.cxx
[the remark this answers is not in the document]
mwoehlke-kitware: If there are no breaks, yes, I'm not surprised. Are you saying if you reformat as shown, Uncrustify *reverts* that? (That would be surprising and probably a bug, as other lines are wrapped as per the suggestion.)

arrows/klv/klv_read_write.txx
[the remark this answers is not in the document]
mwoehlke-kitware: It's unclear what 343 is in this context. Can you paste the full error?

I'm not seeing an issue [here](https://godbolt.org/z/4PPY1v7rh)...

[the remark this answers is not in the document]
daniel-riehm: Update: same with GCC 10.3.0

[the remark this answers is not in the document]
daniel-riehm:
```arrows/klv/klv_read_write.txx: In instantiation of ‘bool kwiver::arrows::klv::_left_shift_overflow(T) [with int shift_amount = 7; T = unsigned char]’:
arrows/klv/klv_read_write.txx:245:37:   required from ‘T kwiver::arrows::klv::klv_read_ber_oid(Iterator&, size_t) [with T = unsigned char; Iterator = __gnu_cxx::__normal_iterator<const unsigned char*, std::vector<unsigned char> >; size_t = long unsigned int]’
arrows/klv/tests/test_klv_read_write.cxx:474:3:   required from ‘void test_read_ber_oid(T, const vec_t&) [with T = unsigned char; vec_t = std::vector<unsigned char>]’
arrows/klv/tests/test_klv_read_write.cxx:507:3:   required from here
arrows/klv/klv_read_write.txx:340:59: error: left operand of shift expression ‘(-1 << 1)’ is negative [-fpermissive]
  340 |   constexpr auto mask = static_cast< T >( ~UnsignedT{ 0 } << retained_bits );
      |                                           ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~```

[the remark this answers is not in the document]
daniel-riehm: Changing `~Unsigned{ 0 }` to `~0ul` seems to remove the error. I suspect this has something to do with the 'promote sub-`int`-sized arithmetic operators to `int`' behavior, which might not be guaranteed to preserve signed-ness?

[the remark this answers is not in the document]
mwoehlke-kitware: Ah, yes. I would use `0ull`, though, or you may run into issues on MSVC. (As a bonus, the type alias can go away.)

mwoehlke-kitware (approved):

PR #1374, opened 2021-07-28
Move configure_logging to kwiver tool setup

as6520:
This PR moves _configure_logging to kwiver tool setup. This prevents it from running when vital is imported and changing log levels.

### DIFF
--- a/python/kwiver/kwiver_tools.py
+++ b/python/kwiver/kwiver_tools.py
@@ -55,6 +55,7 @@ def _setup_environment():
 
 
 def _kwiver_tools(tool_name, args):
+    vital_logging._configure_logging()
     assert tool_name in KWIVER_SUPPORTED_TOOLS, "Unsupported tool {0} specified".format(tool_name)
     tool_environment = _setup_environment()
     tool_path = os.path.join(KWIVER_BIN_DIR, tool_name)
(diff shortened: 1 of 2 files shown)


Purg (approved):
Looks like this should work.

Since the ``_kwiver_tools`` function was not *added* in this PR, this isn't a blocking comment, but this really should have a doc-string describing what's happening and the side-effects of its use (like enabling logging at a certain level, `args` is modified in place, etc.).

Additionally, this function should **return** the integer exit code output of subprocess result (``run_output.return_code``) to be passed as the constructor arg of ``SystemExit``. The integer passed to ``SystemExit`` will be the exit code. Currently the `check=True` will cause an exception to be raised on command failure, but the exit code will always be 1, eating the actual exit code of the underlying tool (unless I'm missing something).

```bash
$ python -c "import subprocess; subprocess.run(['ls', 'sdfsdf'], shell=False, check=True); raise SystemExit(None)"
ls: cannot access 'sdfsdf': No such file or directory
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib64/python3.6/subprocess.py", line 438, in run
    output=stdout, stderr=stderr)
$ echo $?
1
```

as6520:
I have added docstring and return code in https://github.com/Kitware/kwiver/pull/1376

PR #1376, opened 2021-07-28
Update kwiver tools

as6520:
This PR updates the kwiver tools in the following way:
1.  Adds return code to helper function `_kwiver_tools`
2. Adds docstring to the functions
3. Add typing hints to the functions
4. Replaces `format` with f strings.

Depends on #1374

### DIFF
--- a/python/kwiver/kwiver_tools.py
+++ b/python/kwiver/kwiver_tools.py
@@ -31,12 +31,7 @@ def _create_env_var_string(values: List[str]) -> str:
     Returns:
         Colon separated list
     """
-    env_var_value = ""
-    # Append values
-    for value in values:
-        assert isinstance(value, str), "environment variables must be string, {0} specified".format( value )
-        env_var_value += "{0}:".format(value)
-    return env_var_value
+    return ":".join(values)
 
 
 def _setup_environment() -> Dict:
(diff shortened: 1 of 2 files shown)


python/kwiver/kwiver_tools.py
```diff
@@ -12,21 +20,37 @@
...
-def _create_env_var_string( values ):
-    env_var_value = ""
-    # Append values
-    for value in values:
-        assert isinstance(value, str), "environment variables must be string, {0} specified".format( value )
-        env_var_value += "{0}:".format(value)
-    return env_var_value
 
-def _setup_environment():
+def _create_env_var_string(values: List[str]) -> str:
```
Purg, line 24: If this function is only used in this module, which I assume so given it's "private" underscore prefix, we can probably replace it's uses with the wrapped code. `":".join(...)` is simple enough that I don't think it warrants a wrapping function, as well as this function is only used just once.

python/kwiver/kwiver_tools.py
```diff
@@ -54,20 +78,44 @@ def _setup_environment():
...
+    subprocess_complete = subprocess.run(args, shell=False, check=False, env=tool_environment)
+    return subprocess_complete.returncode
+
 
+def plugin_explorer() -> None:
+    """
+    Console script function for plugin_explorer.
 
-def plugin_explorer():
+    Returns:
```
Purg, line 106: Stating a "returns None" seems superfluous to me.

Purg (approved):
LGTM

PR #1379, opened 2021-08-09
Implement read/write for KLV string format

daniel-riehm:
KLV strings are stored literally, with no null terminator, making this a fairly simple task. However, there is one special case to handle - the empty string is stored as a single null byte in order to give it a positive length, since KLV uses "value with zero length" to mean "unknown value" across all data types, and the empty string is not an unknown string. (As a consequence of this, one cannot store the string `"\0"` in KLV, should one ever want to do that.)


daniel-riehm:
jenkins test this please

PR #1381, opened 2021-08-12
Refactor KLV int read / write

daniel-riehm:
Move KLV read / write function implementations for integer types from a .txx file to a .cxx file with explicit instantiations.

### DIFF
--- a/arrows/klv/CMakeLists.txt
+++ b/arrows/klv/CMakeLists.txt
@@ -11,6 +11,7 @@ set( sources
   klv_data.cxx
   klv_key.cxx
   klv_parse.cxx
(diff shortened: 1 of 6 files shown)


daniel-riehm:
jenkins test this please

daniel-riehm:
This is giving me annoying errors - apparently on Mac `deque::const_iterator + size_t` does not yield another `deque::const_iterator`, but some sort of different wrapper type. Windows doesn't seem to be processing the macros correctly in the first place. This behavior is confusing enough (and will remain confusing enough for future users of these functions) that I don't think the effort is worth it.

PR #1383, opened 2021-08-16
Cleanup old cmake stuffs

Purg:
While perusing some files I found dissonence between the currently stated CMake minimum version and some other CMake bits that operate in the context of being under that version, which I believe no longer makes sense. This PR removes those two instances discovered.


dstoup (commented):
Looks good to me. I will add Matthew as a reviewer as well. Would like his opinion.

Purg:
I missed adding a release note, I will update that momentarily.

If making sure people add release/change notes, I might recommend getting the `changelog/topic` check added to the ghostflow-director settings for this repo to pay attention to `*.txt` files in the `docs/release-notes` directory. Error for gitHUB is noticeable with a red-X check result, while WARNING may be appropriate for gitLAB is gf will comment warnings emitted there.

mwoehlke-kitware (approved):

PR #1386, opened 2021-08-17
Fix a bug typo in a cmake file that was missing a closing bracket

Purg:
This came up as an issue trying to locally build the python wheel.

### DIFF
--- a/CMake/utils/kwiver-utils-python.cmake
+++ b/CMake/utils/kwiver-utils-python.cmake
@@ -149,7 +149,7 @@ function (kwiver_add_python_module path     modpath    module)
   # in source build this would be kwiver_python_install_path/project_name
   if(SKBUILD)
     set(pyfile_dst "${CMAKE_BINARY_DIR}/${modpath}/${module}.py")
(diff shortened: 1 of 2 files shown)


mleotta (changes requested):
this is a bug fix and should probably target the release branch instead.

Purg:
Valid. I will get to rebasing and changing release note target after a meeting.

Purg:
So it looks like this fix is on the release branch but it is not on master. What's correct thing to do here?

mleotta:
If that's true then something went wrong in our process.  Whenever a branch
lands in release we are supposed to merge release into master (it would be
nice to automate that in the future if we can, or at least test in CI if
release is not reachable by from master in Git history)

It sounds like we need to merge release into master to get back to a valid
state.

Purg:
Strike that, I made a rookie mistake of not updating my local release branch checkout. I can see the bug actually on release. Let me fixup and get a force update.

johnwparent (approved):

Purg:
Checking back in. This is now based on release and has a positive review from @johnwparent. What's the next action here / who's responsible for merging?

mleotta (approved):

PR #1387, opened 2021-08-17
Implement read/write for KLV float

daniel-riehm:
The "float" format in KLV is just a big-endian serialized [ IEEE-754 ](https://en.wikipedia.org/wiki/IEEE_754#Basic_and_interchange_formats) single- or double-precision floating-point number, same as is standard for most systems to operate on internally (like `float` and `double` in C/C++). It is one of three ways MISB KLV data stores floating-point values. The other two formats attempt to compress the value by limiting its allowed range, while this format straightforwardly allows any representable value.

### DIFF
--- a/arrows/klv/klv_read_write.h
+++ b/arrows/klv/klv_read_write.h
@@ -214,6 +214,40 @@ size_t
 klv_ber_oid_length( T value );
 
 // ---------------------------------------------------------------------------
+/// Read an IEEE-754 floating-point number from a sequence of bytes
+/// (big-endian).
+///
+/// \param[in,out] data Iterator to sequence of \c uint8_t. Set to end of read
+/// bytes on success, left as is on error.
+/// \param length Number of bytes to read. Must be \c sizeof(float) or \c
+/// sizeof(double).
+///
+/// \returns Floating-point value read from \p data.
+///
+/// \throws invalid_value When \p length is not \c sizeof(float) or \c
+/// sizeof(double).
+template < class Iterator >
+KWIVER_ALGO_KLV_EXPORT
+double
+klv_read_float( Iterator& data, size_t length );
+
+// ---------------------------------------------------------------------------
+/// Write an IEEE-754 floating-point number to a sequence of bytes
+/// (big-endian).
+///
+/// \param[in,out] data Writeable iterator to sequence of \c uint8_t. Set to
+/// end of written bytes on success, left as is on error.
+/// \param length Number of bytes to write. Must be \c sizeof(float) or \c
+/// sizeof(double).
+///
+/// \throws invalid_value When \p length is not \c sizeof(float) or \c
+/// sizeof(double).
+template < class Iterator >
+KWIVER_ALGO_KLV_EXPORT
+void
+klv_write_float( double value, Iterator& data, size_t length );
+
+// ---------------------------------------------------------------------------
 /// Read an IMAP-encoded floating-point value from a sequence of bytes.
 ///
 /// For an explanation of IMAP, see the MISB ST1201 document.
--- a/arrows/klv/klv_read_write.txx
+++ b/arrows/klv/klv_read_write.txx
@@ -14,6 +14,7 @@
 #include <string>
 
 #include <cmath>
+#include <cstring>
 
 namespace kwiver {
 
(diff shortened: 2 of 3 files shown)


arrows/klv/klv_read_write.txx
```diff
@@ -55,6 +55,29 @@ KWIVER_ALGO_KLV_EXPORT
...
 
+// ---------------------------------------------------------------------------
+template < class Float > union _float_int_union;
+
+// ---------------------------------------------------------------------------
+template <>
+union _float_int_union< float >
+{
+  float float_value;
+  uint32_t int_value;
```
mwoehlke-kitware, line 66: So... I know this is a "typical" way to do [type-punning](https://en.wikipedia.org/wiki/Type_punning), but as I understand it, absent [`std::bit_cast`](https://en.cppreference.com/w/cpp/numeric/bit_cast) (C++20), the only non-UB way to do this is with `memcpy` ([note](https://stackoverflow.com/a/31080901/1644352)). (Which is somewhat less elegant in some ways, but means you probably don't need this `union` mess, and the compiler should also optimize away the `memcpy`. In fact, "should" here is used in the sense "most folks, *including compiler vendors*, would consider it a compiler bug if it doesn't".)

(Just to keep things exciting, this is true *in C++*. A `union` *is* actually the correct method in C99 and later.)

arrows/klv/klv_read_write.txx
```diff
@@ -343,6 +366,57 @@ klv_ber_oid_length( T value )
...
+klv_read_float( Iterator& data, size_t length )
+{
+  static_assert( std::numeric_limits< float  >::is_iec559 &&
+                 std::numeric_limits< double >::is_iec559,
+                 "platform not IEEE-754 compliant" );
+  if( length == sizeof( float ) )
+  {
+    _float_int_union< float > u;
+    u.int_value = klv_read_int< decltype( u.int_value ) >( data, length );
+    return u.float_value;
```
mwoehlke-kitware, line 381: Per the above:
```suggestion
    static_assert( sizeof( uint32_t ) == sizeof( float ) );
    uint32_t int_value;
    float float_value;
    int_value = klv_read_int< decltype( u.int_value ) >( data, length );
    std::memcpy( &float_value, &int_value, sizeof( float ) );
    return float_value;
```
...and similar in the other places.

arrows/klv/klv_read_write.txx
```diff
@@ -343,6 +366,57 @@ klv_ber_oid_length( T value )
...
 }
 
+// ---------------------------------------------------------------------------
+template < class Iterator >
+double
+klv_read_float( Iterator& data, size_t length )
+{
+  static_assert( std::numeric_limits< float  >::is_iec559 &&
+                 std::numeric_limits< double >::is_iec559,
+                 "platform not IEEE-754 compliant" );
```
mwoehlke-kitware, line 376:
```suggestion
                 "platform is not IEEE-754 compliant" );
```
...or maybe "non-IEEE-754 platform is not supported", but either is fine, and really, even missing the "is" isn't a big deal.

mwoehlke-kitware (changes requested):
I didn't look *super* closely at the test, but by now I'd be surprised if anything other than a stupid bug is hiding there. Main objection is use of UB type punning (see inline comments).

arrows/klv/klv_read_write.h
```diff
@@ -207,6 +207,40 @@ KWIVER_ALGO_KLV_EXPORT
 size_t
 klv_ber_oid_length( T value );
 
+// ---------------------------------------------------------------------------
+/// Read an IEEE-754 floating-point number from a sequence of bytes
+/// (big-endian).
+///
+/// \param[in,out] data Iterator to sequence of \c uint8_t. Set to end of read
+/// bytes on success, left as is on error.
```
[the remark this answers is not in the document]
daniel-riehm: Given the amount of code in this file which uses the non-indenting style, maybe it would be best to re-do the style for all functions all at once in a separate PR, once they're all merged in? I think all code that's going to end up in this file is already in a PR.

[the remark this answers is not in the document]
mwoehlke-kitware: Yeah, this is hardly world-ending 🙂.

mwoehlke-kitware (approved):

PR #1389, opened 2021-08-19
Remove dummy terms from VNIIRS equation

daniel-riehm:
Comments out the dummy placeholder values we previously put in for SNR and RER, opting instead to simply omit their terms from the equation.


daniel-riehm:
jenkins test this please

mwoehlke-kitware (approved):

PR #1397, opened 2021-08-24
Docker: support headless vtk applets

mleotta:
This change is require to run VTK applets like 'color-mesh' that require an x11 server to be running.  Xvfb sets up a virtual frame buffer to allow VTK to run correctly.


danlipsa:
Looks good to me.

PR #1400, opened 2021-09-14
Add klv_value

daniel-riehm:
Add the `klv_value` class, which is similar to `kwiver::vital::any`, but with built-in support for the `<`, `<<`, `==` operators. This ensures we can perform certain basic operations even on type-erased values.


gbrooksby:
This PR looks okay to me.

PR #1401, opened 2021-09-15
Take video start time from first decodable frame

daniel-riehm:
In the unusual case in which a video was poorly cut from a longer sequence, such that the first frame is not a keyframe, the ffmpeg reader counts these first non-decodable frames in its frame count, but is incapable of producing an image for them. It is desirable that the first displayable frame image is given the frame number of 1. This PR puts this into effect by, on file open, finding the first decodable frame and taking the "start" timestamp from there instead of the ffmpeg-provided "start" timestamp.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_input.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_input.cxx
@@ -279,7 +279,24 @@ public:
       this->f_frame_number_offset = 1;
     }
 
-    this->f_start_time = this->f_video_stream->start_time;
+    // Start time taken from the first decodable frame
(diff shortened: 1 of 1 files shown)


chetnieter (approved):
Do we have an example video that demonstrates the issue. Would it be possible assuming the video is small enough (or can be made small enough) to include a new test that includes this video. If we can't due to data restrictions or rights that is fine.

daniel-riehm:
> Do we have an example video that demonstrates the issue. Would it be possible assuming the video is small enough (or can be made small enough) to include a new test that includes this video. If we can't due to data restrictions or rights that is fine.

I don't believe the video which motivated this PR is approved for public release, unfortunately.

PR #1402, opened 2021-09-22
Derive wavelength from image source

daniel-riehm:
The string description of the sensor can sometimes contain the wavelength on which the sensor operates. "SENSOR_NOSE01_IR" likely operates on the infrared spectrum because we see the "IR" string in the sensor name. This PR attempts to fill in missing wavelength information by looking for common such substrings. Neither recall nor precision is expected to be 100% here.


gbrooksby:
This PR looks good to me.

gbrooksby:
Linux is exhibiting the same error as in the timestamp-as-metadata branch.
```
The following tests FAILED:
	 94 - mvg:integrate_depth_maps.integrate_distorted (Timeout)
```

daniel-riehm:
Yes, I'm unsure as to why this is suddenly timing out on all the PR tests. It doesn't look related to anything I'm doing. Maybe the test servers have high load or something? I'm going to profile it on my Linux machine and see if anything is made clearer.

PR #1403, opened 2021-09-23
Make metadata CSV capitalization consistent

daniel-riehm:
Capitalization should be consistent with metadata field names: title case.


hdefazio (approved):
LGTM

PR #1405, opened 2021-09-27
Add config options to limit metadata export frequency

daniel-riehm:
Add `every_n_microseconds` and `every_n_frames` configuration options to the metadata CSV exporter, which allow the user to choose how often to add a new line to the CSV file, in terms of time or frames elapsed. A video may have one or more metadata packets per frame, which results in exported CSV files of significant size when given long videos or very detailed metadata. In use cases where large output files are undesirable, users may choose to export only sparse metadata using these new options.

### DIFF
--- a/arrows/core/metadata_map_io_csv.cxx
+++ b/arrows/core/metadata_map_io_csv.cxx
@@ -8,6 +8,7 @@
 #include "metadata_map_io_csv.h"
 
 #include <vital/any.h>
+#include <vital/exceptions/algorithm.h>
 #include <vital/exceptions/io.h>
 #include <vital/types/geo_point.h>
 #include <vital/types/geo_polygon.h>
@@ -54,6 +55,8 @@ public:
   std::vector< std::string > column_names;
   std::string overrides_string;
   std::vector< std::string > column_overrides;
+  uint64_t every_n_microseconds;
+  uint64_t every_n_frames;
 };
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


arrows/core/metadata_map_io_csv.cxx
```diff
@@ -222,7 +230,8 @@ bool
 metadata_map_io_csv
 ::check_configuration( VITAL_UNUSED vital::config_block_sptr config ) const
```
hdefazio: If this is only going to be used for the export frequencies, could it be renamed to check_export_configuration or something similar?

hdefazio: Could the comment in the header file be expanded to explain what this function checks for?

hdefazio: VITAL_UNUSED can be removed

arrows/core/metadata_map_io_csv.cxx
```diff
@@ -332,10 +350,37 @@ ::save_( std::ostream& fout,
...
+        if( !timestamp.has_valid_time() ||
+            timestamp.get_time_usec() < next_timestamp )
+        {
+          continue;
+        }
+        next_timestamp +=
+          ( ( timestamp.get_time_usec() - next_timestamp ) /
+            d_->every_n_microseconds + 1 ) * d_->every_n_microseconds;
+      }
+      if( d_->every_n_frames )
```
hdefazio, line 372: What happens to the export frequency if both every_n_microseconds and every_n_frames are set? Does there need to be a check added somewhere to make sure one isn't overwriting the other? And if both are set: which one gets used by default - or does it error?

daniel-riehm: The current behavior is that only frames which pass both checks will be exported. So if the frame check limits us to every 5 frames and the microseconds check (effectively) limits us to every 3 frames, only every 5 frames will be exported. However, ideally this should not matter because it will error out once `check_configuration()` fails.

arrows/core/metadata_map_io_csv.cxx
daniel-riehm: This function is virtual; it overrides `check_configuration()` in the base `algorithm` class  Therefore, it has to be named identically.

arrows/core/metadata_map_io_csv.cxx
hdefazio: Ok, sounds good. Is check_configuration called in a different file? And should it be added in here just to double check that only one of these is set before proceeding?

daniel-riehm: Added.

hdefazio: Cool, LGTM now

PR #1407, opened 2021-09-30
Make vital tag traits dynamic

daniel-riehm:
This rather large pull request implements a change - changing vital tag traits access from compile-time to runtime - and then fixes / updates everywhere in the code that breaks as a result. Previously, the traits of vital's metadata tags (name, data type, description, etc) were defined in one giant macro. This was used to instantiate one templated class for each tag, which could only be accessed via `Tag` template parameter. This meant that these traits could only be directly accessed when the tag was known at compile time. A similar strategy was deployed for the `metadata_item` class, resulting in only being able to add a tag to a metadata collection when the identity of the tag was known at compile time. This sort of interface is difficult to work with and difficult to maintain, resulting in a lot of cumbersome, inelegant code (on net 600 lines removed here!).

Since the existing `vital_meta_trait` class was only accessible via template, there was a helper `metadata_trait` class which provided dynamic lookup capabilities. However, this class was instantiated in _n_ different places, incurring runtime costs as it built _n_ separate identical lookup tables. This whole system has been replaced with a better traits interface which only instantiates the traits once and allows lookup from anywhere. We can remove the 100+ separate `typed_metadata` classes and provide better functionality, now allowing dynamic insertion of runtime-determined tags using the new `metadata::add_any` overload.

There is still one ugly / hard-to-maintain code pattern left here, which the new code did have to add a few more instances of:
```
if( type == typeid( bool );
else if( type == typeid( int ) );
...
```
However, this will (hopefully) eventually be deleted. The `vital` metadata system already encourages/enforces/assumes use of only a handful of small types, which as of right now are all separately listed in the pattern above. Given the fact that these constraints are already being put on the data type, a `variant` is a much better pick for a metadata value representation than `any`, and will allow for more elegant treatment of these types. Though we don't have C++17 yet, we do have MPark's C++11 STL-compliant `variant` implementation already copy-pasted in KWIVER. If this PR goes through, using `variant` instead of `any` would be a reasonable next step in this cleanup.


arrows/serialize/protobuf/tests/test_convert_protobuf.cxx
```diff
@@ -516,56 +516,32 @@ TEST( convert_protobuf, geo_polygon )
...
+    kwiver::vital::geo_point::geo_2d_point_t geo_2d{ 42.50, 73.54 };
+    kwiver::vital::geo_point pt{ geo_2d, kwiver::vital::SRID::lat_lon_WGS84 };
+    meta.add< kwiver::vital::VITAL_META_FRAME_CENTER >( pt );
   }
 
   {
-    const auto& info = traits.find( kwiver::vital::VITAL_META_UNIX_TIMESTAMP );
-    auto item = info.create_metadata_item( kwiver::vital::any((uint64_t)12345678) );
-    meta.add( std::move( item ) );
+    kwiver::vital::geo_point::geo_3d_point_t geo{ 42.50, 73.54, 16.33 };
```
gbrooksby, line 531: The change here seems to replace a TIMESTAMP test with a geographical point test.

arrows/serialize/protobuf/tests/test_serialize_metadata.cxx
```diff
@@ -39,47 +39,32 @@ TEST( serialize_metadata, metadata )
...
+    kwiver::vital::geo_point::geo_2d_point_t geo_2d{ 42.50, 73.54 };
+    kwiver::vital::geo_point pt{ geo_2d, kwiver::vital::SRID::lat_lon_WGS84 };
+    meta_sptr->add< kwiver::vital::VITAL_META_FRAME_CENTER >( pt );
   }
 
   {
-    const auto& info = traits.find( kwiver::vital::VITAL_META_UNIX_TIMESTAMP );
-    auto item = info.create_metadata_item( kwiver::vital::any((uint64_t)12345678) );
-    meta_sptr->add( std::move( item ) );
+    kwiver::vital::geo_point::geo_3d_point_t geo{ 42.50, 73.54, 16.33 };
```
gbrooksby, line 54: The change here seems to replace a TIMESTAMP test with a geographical point test.

python/kwiver/vital/types/metadata.cxx
```diff
@@ -18,61 +18,94 @@
...
-  self.add<VITAL_META_ ## TAG>(vital_meta_data_ ## TAG); \
-  break; \
+  auto const& type = tag_traits_by_tag( tag ).type();
+  if( type == typeid( bool ) )
+  {
+    return data.cast< bool >();
   }
-
-  switch (t)
+  else if ( type == typeid( uint64_t ) )
```
gbrooksby, line 29: Curious why the change from a switch statement to an if-else structure.

gbrooksby (commented):
This commit looks good to me.  I made a few comments, but I found nothing that needs to be reworked.

python/kwiver/vital/types/metadata.cxx
daniel-riehm: Previously, it was switching on the metadata tag. The `KWIVER_VITAL_METADATA_TAGS` macro was used to generate the ~100 case statements for each of the tags, which was necessary because we don't know the internal type of `data`, but that can be statically determined from the tag. However, the major change of this pull request removed that macro. We can still achieve the same behavior by checking all the possible types of `data` manually, as we do here. It is a little ugly (as was the macro), but as I mentioned above I am planning a later change which would allow a more elegant way to go through all supported types, so ideally this is just temporary until then.

arrows/serialize/protobuf/tests/test_serialize_metadata.cxx
daniel-riehm: This is where git's line-by-line changes are a bit deceptive; the actual replacement for `TIMESTAMP` is at line 44. This is replacing lines 64-69.

arrows/serialize/protobuf/tests/test_convert_protobuf.cxx
daniel-riehm: See my reply to similar comment above.

PR #1408, opened 2021-10-01
Refactor klv_key

daniel-riehm:
Refactor the `klv_key` classes, `klv_uds_key` and `klv_lds_key`. First, `klv_lds_key` was reduced to just an alias for an unsigned integer, since that is all it is. The base `klv_key` and derived `klv_uds_key` classes could then be merged. There is some general style/simplicity cleanup here, along with additions of `klv_read_*`, `klv_write_*`, and `klv_length_*` functions to match the rest of the new KLV read/write API. The only major functional change to `klv_uds_key` was to implement the somewhat-obscure stipulation that byte 7 does not play a role in key uniqueness - in other words, two UDS keys which only differ in byte 7 are consider to be identical.

### DIFF
--- a/arrows/klv/convert_metadata.cxx
+++ b/arrows/klv/convert_metadata.cxx
@@ -61,7 +61,7 @@ convert_metadata
 void convert_metadata
 ::convert( klv_data const& klv, kwiver::vital::metadata& md )
 {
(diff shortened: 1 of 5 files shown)


gbrooksby:
This PR looks good to me.  I recommend it be merged.

PR #1411, opened 2021-10-06
Add non-numeric KLV formats

daniel-riehm:
Add data formats for `string` and `blob` types.


gbrooksby (approved):
This PR looks ok to merge.

PR #1412, opened 2021-10-06
Add support for alternate metadata representations

daniel-riehm:
This branch was meant to solve the problem of interfacing with data structures for specific metadata formats (e.g. KLV) in a general way. We currently have the `kwiver::vital::metadata` class, but it only maintains an internal lookup table of simple tag-to-value pairs, and can't necessarily hold all the structured information KLV requires to read and write the data back out identically. Thus, we needed a way to both preserve the native KLV data structure while also allowing algorithms to interact with its metadata generically.

Originally, the idea was to make the `metadata` class an abstract interface, with `simple_metadata` and, eventually `klv_metadata` derived classes implementing that interface. However, this ran into a few wrinkles. First, there are many places in the existing codebase where `metadata` is instantiated directly, and we would have to go change all of those to `simple_metadata`, touching lots of code all over KWIVER and hopefully not breaking any of it. Second, functions which were extremely simple for the `map`-based implementation, like `size()`, become a lot more complicated once each tag has to be found and translated from another format. Third, KLV does not support e.g. the `FRAME_NUMBER` vital tag, but we don't want the program to crash or drop tags when an unsupported one is added to an `metadata` object, since that would break the interface. Suffice it to say, it solves many problems to just maintain a `map` of vital tags even in the derived classes, and the alternate representation (e.g. KLV) besides. Any modification of data happens to the two representations in parallel.

All this to justify why the below code is different from the usual KWIVER _modus operandi_ of abstract base classes and concrete implementations. We add two virtual functions `add_internal()` and `erase_internal()`, which serve as an interface between derived and base classes, with the base class implementing all the `map` functionality as before, and the derived class modifying the alternate representation as needed.

Feedback about these design decisions as well as the usual fare would be appreciated.

### DIFF
--- a/vital/types/metadata.cxx
+++ b/vital/types/metadata.cxx
@@ -10,6 +10,7 @@
 #include "metadata.h"
 #include "metadata_traits.h"
 
+#include <vital/logger/logger.h>
 #include <vital/types/metadata_traits.h>
 #include <vital/util/demangle.h>
 
@@ -223,6 +224,14 @@ metadata
 #else
   this->m_metadata_map[ tag ] = item_ptr{ item.release() };
 #endif
+  if( !add_internal( *item ) )
+  {
+    std::stringstream ss;
+    ss  << demangle( typeid( *this ).name() ) << ": tag "
+        << tag_traits_by_tag( tag ).enum_name()
+        << " not supported by internal data structure";
+    LOG_DEBUG( get_logger( "metadata" ), ss.str() );
+  }
 }
 
 // ---------------------------------------------------------------------
@@ -238,7 +247,16 @@ metadata
   // Since the design intent for this map is that the metadata
   // collection owns the elements, we will clone the item passed in.
   // The original parameter will be freed eventually.
-  this->m_metadata_map[ item->tag() ] = item_ptr{ item->clone() };
+  auto const tag = item->tag();
+  this->m_metadata_map[ tag ] = item_ptr{ item->clone() };
+  if( !add_internal( *item ) )
+  {
+    std::stringstream ss;
+    ss  << demangle( typeid( *this ).name() ) << ": tag "
+        << tag_traits_by_tag( tag ).enum_name()
+        << " not supported by internal data structure";
+    LOG_DEBUG( get_logger( "metadata" ), ss.str() );
+  }
 }
 
 // -------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


vital/types/metadata.cxx
```diff
@@ -223,6 +224,14 @@ ::add( std::unique_ptr< metadata_item >&& item )
 #else
   this->m_metadata_map[ tag ] = item_ptr{ item.release() };
 #endif
+  if( !add_internal( *item ) )
+  {
+    std::stringstream ss;
+    ss  << demangle( typeid( *this ).name() ) << ": tag "
+        << tag_traits_by_tag( tag ).enum_name()
+        << " not supported by internal data structure";
+    LOG_DEBUG( get_logger( "metadata" ), ss.str() );
```
mleotta, line 233: This is probably going to happen very frequently for some fields in the KLV backend case, and it will generate a lot of debug messages.  I'd suggest using `LOG_TRACE` instead.

daniel-riehm:
> This approach can be made to work, but it does come with some serious risks. Anytime you try to maintain two parallel copies of data in memory (in the same format or converted) you open yourself to the risk of those copies getting out of sync. It's not always obvious how this can happen, but there are usually code paths you haven't thought of that might allow this. For example, if someone can get access to a pointer to the raw KLV data they may change it and those changes would not be reflected in the vital copy. Another example, the assignment operator is probably broken as currently written because it would not update the internal data.
> 
> That's said this may still be a reasonable approach to take if we can enforce that no one outside the class gains access to the internal memory to change it. For example, `find()` should no longer return the item as `const&` because someone could cast away `const` and change the value.

I agree it is not optimal. I would say that the `const_cast` situation isn't as much of an issue, because `const_cast` is pretty much willfully breaking code anyway - same with using `reinterpret_cast` to cast a `metadata` pointer to the internal `std::map`. However, I take your point about things like assignment operators being messy. There is a third option, perhaps, which is going with the abstract base class and `simple_metadata` derived class, and just having an `as_map()` function, which returns by value a built-on-demand `std::map` representation of whatever underlying data there is. This wouldn't be ideal performance-wise, but we would basically already have to do this to emulate things like `size()`, `begin()`, and `end()` anyway. Thoughts?

daniel-riehm:
I think that is a good idea - I will begin implementing that design.

daniel-riehm:
A future implementation of this will create a new PR.

PR #1415, opened 2021-10-08
Allow constructing Point2d etc. in Python with a location only

predicative:
This increases parity with C++, where the covariance argument is already optional. This is based on a similar fix I made for VIAME (b89916d89e9e5b8bf35c73ce2daade3d421ee3a8).


tao558:
@neal-siekierski The tests look perfect, thanks for adding them. I'll approve once the dashboard is passing. Who merges?

PR #1416, opened 2021-10-08
Add KLV tag traits lookup tables

daniel-riehm:
Since there are multiple KLV standards, there are multiple sets of KLV tags. Each needs to be able to look up a tag's traits based on one of the tag's identifying features. This `klv_tag_traits_lookup` class is created to avoid having to re-implement that lookup behavior for each new standard, and to allow some functionality to operate in a standard-agnostic way by just being given an instance of this class.


gbrooksby (approved):
This PR looks good to merge.

PR #1417, opened 2021-10-11
Add floating-point KLV formats

daniel-riehm:
Add `klv_data_format`s for the several KLV floating-point types. These are mostly wrapper classes like the other formats, though they do perform some of their own calculations while printing out the values. The various floating-point formats offer various levels of precision, but the resulting values are all stored internally as `double`s. When printing the values, only the number of digits actually known should be printed out. So if some format offers 5.7 digits of precision, we might print its value as `"12.3457"`, while a more precise format with 8.1 digits of precision might print the same value as `"12.3456789"`.

### DIFF
--- a/arrows/klv/klv_data_format.cxx
+++ b/arrows/klv/klv_data_format.cxx
@@ -7,11 +7,26 @@
 
 #include "klv_data_format.h"
 
+#include <iomanip>
+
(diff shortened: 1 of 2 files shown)


gbrooksby (approved):
This PR looks good to commit.

PR #1418, opened 2021-10-12
Add export specifier to klv_value

daniel-riehm:
Add export specifiers to `klv_value` classes and functions to avoid link errors.


daniel-riehm:
Turns out it was because I didn't also add the export specifiers to the `friend` declarations inside `klv_value`. Fixed, and now the tests pass.

PR #1425, opened 2021-10-19
Add KLV ST1108 parser

daniel-riehm:
Create the read / write tools for the [MISB ST 1108](https://gwg.nga.mil/misb/docs/standards/ST1108.3.pdf) local set. This includes the creation of some standard-specific `enum`s and other types, and tools to handle those. To avoid confusion, the ST1108 local set (this PR) contains one or more of the ST1108 _metric_ local set (#1422); the nomenclature can be confusing, since they are both defined under the same standard.


arrows/klv/klv_1108.cxx
```diff
@@ -0,0 +1,448 @@
...
+  static klv_tag_traits_lookup const lookup = {
+#define ENUM_AND_NAME( X ) X, #X
+    { {},
+      ENUM_AND_NAME( KLV_1108_UNKNOWN ),
+      std::make_shared< klv_blob_format >(),
+      "Unknown",
+      "Unknown tag.",
+      0 },
+    { // "Key" column
+      { 0x060E2B3401010101, 0x0E01050100000000 },
```
gbrooksby, line 369: See comment above on substituting a #define.  Applies to subsequent usage below as well.

gbrooksby (approved):
I recommend seeing my comments on some of the hex constants and changing them to #defines, but otherwise this PR looks okay to merge.

arrows/klv/klv_1108.cxx
daniel-riehm: > The table on page 3 of the MISB standard document gives reasonable names for theses keys.

Yes, and those reasonable names are included a few lines below each key.

Each 16-byte key (composed here of two 8-byte integers) is unique, and will only appear once in the code - here, accompanied by its name, enumeration, format, and description. All future uses of this key will reference the value here (same as the local set key above - that's the reason why there is a function specifically to return that value). It seems to me that `#define`s or `constexpr` variables will simply require adding hundreds more lines of code without any substantial benefit. Their names and where their arbitrary-seeming value came from are already documented in the file. No more information is added by giving each half of each of the 200ish eventual keys its own name, to be used once. Sort of like:
```
// Add 1 for null character
auto const string_length = length + 1;
```
versus
```
constexpr size_t null_character_length = 1;
auto const string_length = length + null_character_length;
```
 The second is not obviously better than the first. Am I missing something here?

gbrooksby: Fair enough.  I understand that adding a few hundred lines of code with no effect on performance has a real cost to the project.  I'll leave it to @mleotta to decide if this is how project $ should be spent.

PR #1442, opened 2021-10-29
Adjust metadata video timestamp csv formatting

daniel-riehm:
When exporting metadata to a CSV, instead of directly writing the video timestamp in microseconds, write a more human-readable format: `HH:MM:SS:SSSSSS`.

### DIFF
--- a/arrows/core/metadata_map_io_csv.cxx
+++ b/arrows/core/metadata_map_io_csv.cxx
@@ -18,6 +18,7 @@
 
 #include <vital/range/iota.h>
 
+#include <iomanip>
 #include <iterator>
 #include <string>
 #include <typeinfo>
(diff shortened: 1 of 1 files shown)


mwoehlke-kitware (approved):
I approve the code. Someone else should perhaps approve the semantics 🙂.

gbrooksby (approved):
This PR looks okay.  It is not a change made in this version of the file, but if you look on line 62, I think there should be a blank line after the `namespace` and before the separator line.

daniel-riehm:
> It is not a change made in this version of the file, but if you look on line 62, I think there should be a blank line after the `namespace` and before the separator line.

Yes, I have addressed some leftover style errors in another PR. Good catch, though!

PR #1446, opened 2021-11-02
Add image height and width metadata fields

daniel-riehm:
Add vital metadata tags for image width and height so that this information is available for export.


gbrooksby (approved):
These changes look okay to me.

PR #1449, opened 2021-11-03
Break up composite data types when exporting to csv

daniel-riehm:
This PR changes how the metadata CSV exporter configuration treats composite types (i.e. `geo_point` and `geo_polygon`). The value for these types comes from only one metadata field, but they result in 3 and 8 CSV columns, respectively. Currently, these sub-values cannot be individually reordered or renamed. With these changes, each vital metadata field can explicitly produce multiple columns, each of which can be requested individually by use of a zero-indexed name. For example, sensor altitude can be requested by the column name `SENSOR_LOCATION.2`. Indices default to `0` if not specified.

Unfortunately, there's no unit tests for this exporter, but I did a check by hand on a test video and it seems to be working correctly.

### DIFF
--- a/arrows/core/metadata_map_io_csv.cxx
+++ b/arrows/core/metadata_map_io_csv.cxx
@@ -20,6 +20,7 @@
 
 #include <iomanip>
 #include <iterator>
+#include <stdexcept>
 #include <string>
 #include <typeinfo>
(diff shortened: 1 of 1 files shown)


gbrooksby (approved):
This PR looks okay to me.

PR #1451, opened 2021-11-04
Add interval_map utility class

daniel-riehm:
Add the `interval_map` class and associated unit tests.

The `interval_map` is a map data structure with non-overlapping `interval`s for keys. It allows logarithmic-time lookup by either single value or `interval`. Erasure and value assignment are done "intelligently": adjacent `interval`s with the same value are automatically merged, empty `interval`s are automatically deleted, and erasing/reassigning a section from the middle of an existing `interval` will split the original entry in two.


daniel-riehm:
jenkins test this please

gbrooksby (approved):
This PR looks okay to me.

PR #1452, opened 2021-11-04
Fix bundle adjust common intrinsics test

mleotta:
The test was checking if all cameras are shared after bundle adjustment
when forcing shared cameras, but the problem was that the test cameras
were shared to begin with, so the test would pass if no sharing change
was made.  This change forces the test cameras to use unique intrincs
from the start to test of the bundle adjustment forces them to be common
again.

### DIFF
--- a/arrows/ceres/options.cxx
+++ b/arrows/ceres/options.cxx
@@ -382,9 +382,10 @@ camera_options
     // - we are forcing common intrinsics and this is the first frame
     // - we are auto detecting shared intrinsics and this is a new sptr
     if( this->camera_intrinsic_share_type == FORCE_UNIQUE_INTRINSICS ||
-        ( this->camera_intrinsic_share_type == FORCE_COMMON_INTRINSICS &&
-          int_params.empty() ) ||
-        camera_intr_map.count( K ) == 0 )
+      ( this->camera_intrinsic_share_type == FORCE_COMMON_INTRINSICS &&
+        int_params.empty() ) ||
+      ( this->camera_intrinsic_share_type == AUTO_SHARE_INTRINSICS &&
+        camera_intr_map.count( K ) == 0 ) )
     {
       this->extract_camera_intrinsics( K, &intrinsic_params[ 0 ] );
       // update the maps with the index of this new parameter vector
--- a/arrows/ceres/options.cxx
+++ b/arrows/ceres/options.cxx
@@ -385,7 +385,7 @@ camera_options
       ( this->camera_intrinsic_share_type == FORCE_COMMON_INTRINSICS &&
         int_params.empty() ) ||
       ( this->camera_intrinsic_share_type == AUTO_SHARE_INTRINSICS &&
-        camera_intr_map.count( K ) == 0 ) )
+        camera_intr_map.find( K ) == camera_intr_map.end() ) )
     {
       this->extract_camera_intrinsics( K, &intrinsic_params[ 0 ] );
       // update the maps with the index of this new parameter vector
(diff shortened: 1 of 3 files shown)


mleotta:
If all goes well, the unit test for shared intrinsics should now properly fail.  I will then add a commit that fixes the code so the test passes again.

arrows/ceres/tests/test_bundle_adjust.cxx
```diff
@@ -384,6 +384,21 @@ test_ba_intrinsic_sharing( camera_map_sptr cameras,
...
+// ----------------------------------------------------------------------------
+// Make sure each camera has unique (not shared) intrinsics
+camera_map_sptr make_intrinsics_unique(camera_map_sptr cameras)
+{
+  camera_map::map_camera_t new_cams;
+  for (auto ci : cameras->cameras())
+  {
+    auto cam = std::dynamic_pointer_cast<camera_perspective>(ci.second);
+    auto new_cam = std::make_shared<simple_camera_perspective>(
+      cam->center(), cam->rotation(), cam->intrinsics()->clone());
```
borovik135, line 396: need to check if cam==nullptr and complain if so instead of risking a core dump

borovik135 (changes requested):
test currently fails, as expected; we need to check for null ptrs for cam casts though

arrows/ceres/tests/test_bundle_adjust.cxx
mleotta: Not really a big deal here because this function is only exposed to unit tests.  The worst that could happen is that it causes a test failure if used on non-perspective cameras.  However, if someone copied this code to use elsewhere it could be a problem, so it's worth added the extra check.

mleotta:
This should be good to go now.  Both the code and the unit test are now fixed and the CI is passing.

borovik135:
> This should be good to go now. Both the code and the unit test are now fixed and the CI is passing.

The unit test on common intrinsics appears to pass now.

Purg:
@borovik135 If you last comment was approval of the current state of the PR, please submit a re-review to change your current "changed requested" state into an "approved" one.

borovik135:
I asked @mleotta  to consider replacing 'camera_intr_map.count( K ) == 0' with 'camera_intr_map.find( K ) == camera_intr_map.end()' to slightly improve performance (line 388), which I think is still pending his response.

mleotta:
I missed that comment about replacing `count` with `find`.  I'm not opposed to that change, but I'm also not convinced that it will make a measurable performance improvement either.  For a `map`, `count` can find at most 1 item and should be about as efficient as finding that one item.

arrows/ceres/options.cxx
```diff
@@ -382,9 +382,10 @@ ::extract_camera_parameters( camera_map::map_camera_t const& cameras,
     // - we are forcing common intrinsics and this is the first frame
     // - we are auto detecting shared intrinsics and this is a new sptr
     if( this->camera_intrinsic_share_type == FORCE_UNIQUE_INTRINSICS ||
-        ( this->camera_intrinsic_share_type == FORCE_COMMON_INTRINSICS &&
-          int_params.empty() ) ||
-        camera_intr_map.count( K ) == 0 )
+      ( this->camera_intrinsic_share_type == FORCE_COMMON_INTRINSICS &&
+        int_params.empty() ) ||
+      ( this->camera_intrinsic_share_type == AUTO_SHARE_INTRINSICS &&
+        camera_intr_map.count( K ) == 0 ) )
```
mleotta, line 388:
```suggestion
        camera_intr_map.find( K ) == camera_intr_map.end() ) )
```

borovik135 (approved):
The changes look good to me now. I think this is ready to be merged.

PR #1454, opened 2021-11-08
Make metadata_map.frames() const

daniel-riehm:
The modified function returns the set of frames for which data is present in a `metadata_map`. As nothing is or should be modified internally, `frames()` should be declared `const`.


gbrooksby (approved):
Looks good to me.

PR #1457, opened 2021-11-10
Add klv_timeline class

daniel-riehm:
Create a class to track KLV data over time. This is the final intended product of KLV parsing, capable of efficiently holding data from any standard(s) over any period of time in a form which can be manipulated relatively easily. Enough information is retained here to validly (re-)encode any supported KLV stream.

Just like in a `klv_set`, of which this is a sort of amalgamation, some tags can have multiple values, though most tags only support one value at a time. This is reflected in the interface, which has options for working under both single- and multi-value paradigms.

### DIFF
--- a/arrows/klv/klv_timeline.h
+++ b/arrows/klv/klv_timeline.h
@@ -1,0 +1,129 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/// \file
+/// \brief Declaration of KLV timeline class.
+
+#include <arrows/klv/klv_packet.h>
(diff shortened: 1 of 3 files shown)


gbrooksby (approved):
This PR looks okay to me.

arrows/klv/klv_timeline.h
```diff
@@ -0,0 +1,102 @@
...
+
+namespace kwiver {
+
+namespace arrows {
+
+namespace klv {
+
+// ----------------------------------------------------------------------------
+/// Tracks the value of KLV fields over time.
+class KWIVER_ALGO_KLV_EXPORT klv_timeline
```
linus-sherrill, line 21: There is such a nice writeup of this class as part of this PR. Would you consider putting that text here also as permanent documentation since it provides high level information?

gbrooksby (approved):
Still looks good to me.

PR #1459, opened 2021-11-10
More unifying of separator line formats.

gbrooksby:
Fixes added per comments by M. Woehlke on the previous iteration.  This version catches the indented separators as well as those written with `===...` rather than `---...`.


mwoehlke-kitware (commented):
I'm somewhat on the fence whether to forbid use of `=` for separator lines. There is an argument made for using `=` to denote "higher level" things (usually, `struct`s / `class`es). OTOH, they are not used consistently right now.

We should be sure we have asked ourselves whether we want to forbid them entirely (easier to maintain, but less informative), or leave them alone for now (aside from fixing lengths) and revisit later.

Looks good otherwise (lengths look correct, several were spot checked).

mwoehlke-kitware (approved):

PR #1461, opened 2021-11-12
Make output file extension override configured output format

daniel-riehm:
For the `dump-klv` tool, make it so that if our output file is named `<something>.json`, we will output in JSON format even if the configuration exporter is set to `csv`.

### DIFF
--- a/arrows/core/applets/dump_klv.cxx
+++ b/arrows/core/applets/dump_klv.cxx
@@ -112,6 +112,28 @@ dump_klv
     config->merge_config( kv::read_config_file( cmd_args["config"].as<std::string>() ) );
   }
 
+  // Output file extension configures exporter
+  if ( cmd_args.count( "log" ) &&
+       !cmd_args.count( "metadata_serializer:type" ) )
+  {
+    const std::string filename = cmd_args[ "log" ].as< std::string >();
+    auto const extension_pos = filename.rfind( '.' );
+    if( extension_pos != filename.npos )
+    {
+      static std::map< std::string, std::string > const extension_map = {
+        { ".JSON", "json" },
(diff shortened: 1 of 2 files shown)


arrows/core/applets/dump_klv.cxx
```diff
@@ -112,6 +112,28 @@ ::run()
     config->merge_config( kv::read_config_file( cmd_args["config"].as<std::string>() ) );
   }
 
+  // Output file extension overrides configured exporter
+  if ( cmd_args.count("log") )
```
mleotta, line 116: We may also want to allow overriding this autoselection behavior with specific implementation.  So I would suggest checking if `metadata_serializer:type` is set in the config and only do this file extension based configuration if it is not specified.  Currently it is specified as `json` in `applets/dump_klv.conf` so we will have to remove that so the default behavior is to check the extension.

daniel-riehm: I have changed it according to your suggestion. Note that there is no "backup backup", such that if there is no config option and no recognized extension, it will crash. This seems like behavior within reason, but I figured I'd point it out just in case we DO want a "backup backup".

gbrooksby: I'm never a huge fan of allowing known opportunities for crashes.  I prefer default behaviors.  For example, the "backup backup" could be to just write a csv file to whatever filename was provided.  We're presuming that a person would _**never**_ possibly want to write a csv file to any filename that ends in anything but `.csv`.  If a person is surprised by what they get in their output file, they can read the documentation or the code.

gbrooksby (approved):
This PR looks good to me.

PR #1462, opened 2021-11-12
Fix metadata CSV exporter configuration

daniel-riehm:
Fix two problems with the metadata CSV exporter which were crippling the `dump-klv` applet. First, give all configuration values common-sense default values. When given an empty configuration file, the exporter will simply export all fields present, using their default names as column headers, instead of crashing. Second, fix a bug in which the configuration that `get_configuration()` produces actually registers as invalid when fed to `check_configuration()`.

### DIFF
--- a/arrows/core/metadata_map_io_csv.cxx
+++ b/arrows/core/metadata_map_io_csv.cxx
@@ -325,15 +325,14 @@ void
 metadata_map_io_csv
 ::set_configuration( vital::config_block_sptr config )
 {
-  d_->write_remaining_columns = config->get_value< bool >(
-    "write_remaining_columns" );
-  d_->write_enum_names = config->get_value< bool >( "write_enum_names" );
+  d_->write_remaining_columns =
+    config->get_value< bool >( "write_remaining_columns", true );
+  d_->write_enum_names =
+    config->get_value< bool >( "write_enum_names", false );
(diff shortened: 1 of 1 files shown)


gbrooksby (approved):
Looks like this is mostly just a change to conform to a different API.  It looks good to me.

mleotta:
Sorry, I'm late to the party on this.  It's okay that this landed in `master`, but generally, if we have branches which are bug fixes that are also relevant to the last release version of KWIVER we should rebase these changes on the `release` branch and target the PR to `release`.  It's also good to add a note in the `release.txt` release notes file indicating what was fixed.  Once merged into `release` we then merge release into `master`, so `master` gets the same changes.

This is already in `master`, but maybe we can rebase to `release` and also put it there?

daniel-riehm:
@mleotta  It looks like `release` is sufficiently behind `master` that these changes don't really make sense with respect to `release`. In fact, as far as I can tell, the `dump-klv` CSV writer seems to work fine in the current `release` branch. It has a more limited set of features than `master`, but it works. The problems addressed in this PR were introduced relatively recently, with the addition of addition config options by myself and others. This is fortunate, then: `release` was never broken.

mleotta:
Got it.  Then this was done correctly.  Sorry for my confusion.

PR #1463, opened 2021-11-15
Add frame dump capability to dump-klv

daniel-riehm:
Add the ability to dump frame images in addition to KLV metadata. Options are included to specify the image file format and output directory. OpenCV is specified as the default image writing implementation because the current `Dockerfile` guarantees it will be built.


gbrooksby (approved):
This PR looks okay to me.

PR #1464, opened 2021-11-17
Add KLV demuxer

daniel-riehm:
This PR adds the de-multiplexer (demuxer) for KLV. This is a coroutine-like object which holds state for the process of turning parsed KLV packets into a KLV timeline. Using this object, packets can be added to the timeline frame by frame as the video is read. The user could then erase old data on the timeline after is has been used, minimizing the memory footprint of the KLV decoding process, or retain all data and efficiently contain metadata for the entire video stream, perhaps for export.

The possibilities of empty, unparsed/invalid, incorrectly duplicated, or missing data complicate some of the logic here, and there's valid questions to be asked weighing the likelihood of these things and how rigid or accommodating our system should be. At the moment, the priority is to parse correctly-encoded packets correctly, and ideally retain enough information about incorrect or unfamiliar packets to replace them in the stream when re-encoding.

The unit tests added here are not intended to be exhaustive, but merely ensure the bulk of the functionality works as intended.

### DIFF
--- a/arrows/klv/klv_data_format.cxx
+++ b/arrows/klv/klv_data_format.cxx
@@ -7,6 +7,13 @@
 
 #include "klv_data_format.h"
 
+#include "klv_0104_new.h"
+#include "klv_0601_new.h"
+#include "klv_1108.h"
+#include "klv_1108_metric_set.h"
+#include "klv_blob.h"
+#include "klv_packet.h"
+
 #include <iomanip>
 
 #include <cfloat>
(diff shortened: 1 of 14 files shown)


arrows/klv/klv_demuxer.cxx
```diff
@@ -0,0 +1,431 @@
...
+      for( auto const tag : { KLV_1108_ASSESSMENT_POINT,
+                              KLV_1108_WINDOW_CORNERS_PACK } )
+      {
+        auto const it = timeline.find( standard, tag, index );
+
+        // All this to check whether the values are different, the complexity
+        // coming from the fact that we don't assume either value exists in the
+        // first place, since WINDOWS_CORNER_PACK is optional
+        if( it != timeline.end() && !it->second.empty() &&
+            ( it->second.begin()->value.empty() != !parent_set.has( tag ) ||
```
gbrooksby, line 69: This is another case of a double negative; i.e. `!= !foo`.  Is the logic
`( it->second.begin()->value.empty() == parent_set.has( tag )` not possible?  It seems more readable if it's possible.

gbrooksby (approved):
I see no real problem with this PR.  I did make a comment about some double-negative logic that I saw, but functionally, my suggestions would make no difference.

gbrooksby:
I'm not sure why the Mac build is failing here.  I checked this branch out and built it on my Mac without a problem.  The test that fails here passes on my Mac.

daniel-riehm:
@gbrooksby @mleotta @mwoehlke-kitware This is finally passing on all platforms, but a few changes were necessary to get there, so possibly a check-up re-review would be in order. The actual `klv_demuxer` code has not really changed, but many KLV templates have been moved from headers to `.cxx` files and explicit instantiations used instead to get it work on all platforms.

Hopefully we can finally get this PR through and move on to the final replacement of the old system by the new.

arrows/klv/klv_data_format.cxx
```diff
@@ -561,6 +795,34 @@ ::description() const
   return ss.str();
 }
 
+#define KLV_INSTANTIATE( T ) \
```
mleotta, line 798: This is not critical, and need not hold up merging, but I have a suggestion on how to organize this code so that all the instantiations of all the types are not all in this one file.  It's a pattern that is used commonly in VXL.  You make a template header file that contains all the template function definitions that you've added in this file and the instantiate macros.  This file often gets a special extension, like `.hxx`, and gets include guards like a header file, but it's not a public header.  We used to use `.txx`, but switched to `.hxx` because it is more commonly recognized as C++ by IDEs.  For each type you want to instantiate you create a `.cxx` file for that type which would look something like this:

File name: `klv_data_format+klv_blob-.cxx`
Contents:
```c++
#include "klv_blob.h"
#include "klv_data_format.hxx"

KLV_INSTANTIATE( klv_blob );
```
In VXL, all of these special instantiation files go in a `templates` subdirectory.  The advantage of this approach is that you keep the template implementations separate from your instantiation for all the various types.  This allows you to compile them in parallel and you don't have one massive file that has to include everything.

arrows/klv/klv_value.cxx
```diff
@@ -178,6 +378,51 @@ operator<<( std::ostream& os, klv_value const& rhs )
   return rhs.empty() ? os << "(empty)" : rhs.m_item->print( os );
 }
 
+// ---------------------------------------------------------------------------
+#define KLV_INSTANTIATE( T )                         \
```
mleotta, line 382: same comment as with `klv_data_format`.  You could put the template definitions and this macro in a `.hxx` file and then include it in several `.cxx` files to instantiate different types so this `klv_value` code doesn't have to be entangled with all the different types you want to instantiate.

PR #1467, opened 2021-11-18
Move internal variant.hpp to the cxx file

dstoup:
Having the internal/variant/variant.hpp include in metadata.h prevents consuming packages from building. Move variant.hpp to metadata.cxx as designed.

### DIFF
--- a/vital/types/metadata.cxx
+++ b/vital/types/metadata.cxx
@@ -11,6 +11,7 @@
 
 #include <vital/types/metadata_traits.h>
 #include <vital/util/demangle.h>
+#include <vital/internal/variant/variant.hpp>
 
 #include <typeindex>
 
--- a/vital/types/metadata.h
+++ b/vital/types/metadata.h
@@ -12,7 +12,6 @@
 
 #include <vital/any.h>
 #include <vital/exceptions/metadata.h>
-#include <vital/internal/variant/variant.hpp>
 #include <vital/types/geo_point.h>
 #include <vital/types/geo_polygon.h>
 #include <vital/types/metadata_tags.h>


dstoup:
There still looks to be a problem with `#include <vital/util/visit.h>`. No specifically set as internal by path but it includes vital/internal/variant/variant.hpp so internal all the same. Of course, that means this PR is pointless since variant.hpp is still coming in. The bigger issue is how to get the private headers out of metadata.h and still use variant. I haven't looked at the changes in c4cc2102c928 too deeply but it seems like it shouldn't be too hard to hide variant from the public interface Thoughts?

daniel-riehm:
If the issue is with compiling `variant.hpp`, I doubt moving it to a `.cxx` will help much. If the issue is that it is not a public header, I would propose the solution of (moving it out of `internal/` and) making it one. We have precedence for third-party >C++11 STL data structure implementations being public in KWIVER with `vital/optional.h`. The alternative would be to write a public wrapper around `variant`, but that seems like needless abstraction and difficult to do properly anyway. Curious to hear @mwoehlke-kitware 's thoughts on this.

dstoup:
@daniel-riehm the issue is that variant.hpp is under internal and should never appear in a .h file, only cxx per the documentation of the kwiver internal headers. Kwiver itself compiles fine with variant.hpp included in metadata.h but vidtk, for example, is not able to compile because variant.hpp isn't available (internal headers aren't installed). This PR was simply a naive attempt to conform to the internal documentation's suggestion.

One solution might be, revisit whether or not variant should be internal. If not, we can install it and problem solved. If it is needed as internal then the classes that include it need to hide the dependency from the API by not exposing variant in the public interface, e.g. variant.hpp included from cxx only. One way to do that is with a private impl class that hides the variant details which probably requires removing a lot of the implementation from the header, etc. But, I haven't really dug and understood the code well enough to know what problems that will introduce.

@mleotta @mwoehlke-kitware  .... is there a good reason to keep variant as 'internal'?

mleotta:
Is this change still needed or has this been resolved by another PR?

daniel-riehm:
Addressed by #1522

PR #1468, opened 2021-11-18
Add KLV-to-vital converter

daniel-riehm:
Add a function to extract `vital::metadata` tags from a `klv_timeline` at a given point in time. All three currently supported standards continue to be supported here.

The `timestamp` here is a frame time, expressed as a MISP Precision Timestamp. This is the unique timestamping method that MISB KLV data uses, and the relevant standards are found [here](https://gwg.nga.mil/misb/docs/standards/ST0603.5.pdf) and [here](https://gwg.nga.mil/misb/docs/standards/ST0604.3.pdf). Videos with KLV data are supposed to come with MISP timestamps embedded for every frame, allowing precise alignment of the metadata with the imagery data. If these timestamps are not present, as is the case in some real-world data, they will have to be estimated by some means, but that is beyond the scope of this function, which only concerns itself with correct behavior given correct inputs.

### DIFF
--- a/arrows/klv/CMakeLists.txt
+++ b/arrows/klv/CMakeLists.txt
@@ -8,6 +8,7 @@ set( sources
   convert_0104_metadata.cxx
   klv_blob.cxx
   klv_checksum.cxx
+  klv_convert_vital.cxx
   klv_data.cxx
   klv_data_format.cxx
   klv_key.cxx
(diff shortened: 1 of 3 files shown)


gbrooksby (approved):
This PR looks okay to me.

PR #1469, opened 2021-11-19
Refactor MISP timestamp parsing

daniel-riehm:
MISP timestamps are key to aligning KLV data with video frames. We therefore want to add MISP timestamp reading to our FFMPEG decoder; however, the current interface for that method requires a `std::vector`, which is not the form FFMPEG provides the packet data in. Thus, calling this function as-is would require copying of the bytes of every frame packet into a temporary dynamic array for no real reason. This PR changes the interface to accept generic iterators instead, avoiding a significant performance hit.

Given the significant style deviations and verbosity of this code, it seemed justified to simply fix all the problems with it at once, hence the total rewrite.

Note that the VIDL FFMPEG reader does current use MISP, and that is also updated here. The plain-old-FFMPEG reader does not use MISP, which will be added in a future PR.


arrows/klv/misp_time.h
```diff
@@ -2,47 +2,92 @@
...
+  for( auto const i :
+       kwiver::vital::range::iota( misp_detail::timestamp_length ) )
+  {
+    // Every third byte is set to 0xFF to avoid the timestamp being read as a
+    // start tag for some other sort of data
+    if( i % 3 != 2 )
+    {
+      result <<= 8;
+      result |= *it;
+    }
```
gbrooksby, line 80: I'm not following either the comment or the logic here.  `i % 3 != 2` will change 2 out of every three bytes, not every third byte as the comment says. If this is tested and works, then great. Maybe your can explain it to me sometime.

gbrooksby (approved):
This looks okay as long as it functions properly.  I only have a question about one small portion, which may just be my own misunderstanding.

arrows/klv/misp_time.h
daniel-riehm: The bytes are formatted like this: `AABAABAABAA`. We want to extract and concatenate the eight `A`s while ignoring the three `B`s. So the body of the `if` statement is executed for the first two out of every three bytes, which is the desired behavior.

PR #1474, opened 2021-11-22
Revert line incorrectly identified as separator

daniel-riehm:
Currently, `master` fails to build due to the automatic separator-fixing tool mistaking this line for a separator.

### DIFF
--- a/vital/plugin_loader/plugin_manager_internal.h
+++ b/vital/plugin_loader/plugin_manager_internal.h
@@ -10,9 +10,8 @@ namespace kwiver {
 namespace vital {
 
 // ----------------------------------------------------------------------------
-//  This class just exposes the protected members of the base class.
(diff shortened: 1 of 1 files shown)


gbrooksby (approved):
Approved.  I thought I got all these.

gbrooksby:
Strange that this didn't build, because I built it locally before I checked in the code.

daniel-riehm:
> Strange that this didn't build, because I built it locally before I checked in the code.

Likely your local `CMakeCache.txt` does not have all the different submodules of KWIVER enabled.

gbrooksby:
I've made a note in the script to check and change this next time around. I changed your suggestion sligtly to
`//\s*([-=])\1+\s*$`.   The script has already been run on everything in kwiver/vital, with a few targeted exceptions.

PR #1476, opened 2021-11-23
Add klv_packet_timestamp

daniel-riehm:
Technically, all frames in videos with KLV data should come with a MISP timestamp that allows precise correlation between KLV data and frame images. Practically, however, the presence or validity of these timestamps is not guaranteed - in fact, our current embedded test video has invalid timestamps. Backup methods of correlating these two streams may find it useful to query the embedded timestamp in each KLV packet to determine the approximate MISP time of a frame. The `klv_packet_timestamp()` function provides this ability.


gbrooksby (approved):
This looks ok to merge.

PR #1477, opened 2021-11-23
Log, don't throw, on wrong checksum

daniel-riehm:
There is a large depository of videos we would like to read the KLV for, except whoever encoded them implemented the checksum algorithm incorrectly. Currently we throw an exception and don't try to parse a packet when the checksums don't match. This PR changes this to a logged error and still attempts to parse the packet anyway. Not ideal, but such concessions need to be made given the apparent state of existing KLV encoders.


gbrooksby (approved):
This looks safe to merge.

PR #1478, opened 2021-11-24
Remove sensor fov name as direct conversion

daniel-riehm:
When converting KLV to vital, the `SENSOR_FOV_NAME` field need special treatment - it must be converted from its `enum` type to `uint64_t`. There is already code for this, but it is preempted by accidentally also being included in the generic converter list, which produces a runtime error due to type mismatch. This PR removes it from the generic conversion list.


gbrooksby (approved):
Look fine to merge.

PR #1480, opened 2021-12-02
Prevent frame number from being printed twice

daniel-riehm:
The frame number is automatically printed as the first column, so no need to automatically insert it in an additional column later on. It can still be re-requested by the configuration if that is truly desired, but double-printing shouldn't be the default behavior.


gbrooksby (approved):
This looks okay to merge.

PR #1481, opened 2021-12-07
Replace old KLV system with new

daniel-riehm:
This PR deletes the old KLV system, removes the `_new` suffix from the files that previously needed that differentiation, and hooks up the new system to the FFMPEG reader. A few other tweaks are necessary here and there to get everything working smoothly, but that's the bulk of it. (The giant whole-file diffs here are just files being deleted and/or renamed, and are not new code)

Our existing tests for the FFMPEG reader's metadata capabilities pass, though admittedly they are hardly comprehensive. Some manual testing was done with variously sourced videos, and there are now no degradations relative to the old system that I could find in the CSV export. All of these tests are only working on the `vital` side of things, which can only ever express a translation of a subset of KLV data. More complete validation will be better provided by the development of the KLV (re)multiplexer in the near future.


arrows/klv/klv_0601.cxx
```diff
@@ -3,430 +3,1432 @@
...
   {
-    return false;
+    return true;
+  }
+  if( lhs.numerator == rhs.numerator )
+  {
+    return lhs.denominator < rhs.denominator;
   }
+  return false;
+}
```
gbrooksby, line 1257: `lhs.numerator < rhs.numberator` only works if `lhs.denominator` == `rhs.denominator`.  But according to line 1252, the denominators are not necessarily equal.  Daniel knows the standard better than I and there may be a reason this can work, but it seems incomplete.

gbrooksby (commented):
All looks good except the one question I have posted relating to klv_0601.cxx.

arrows/klv/klv_0601.cxx
daniel-riehm: This was because I wasn't really thinking of the frame rate as representing an actual number, per se, but as a data structure which needs some arbitrary but consistent ordering mechanism (provided by the `<` operator) for sorting purposes. You can see similar operators implemented for all the other kinds of objects ('klv_0601_control_command') where one being literally 'less than' another is nonsensical. I see your point where it is counterintuitive, and a more literal implementation would probably be better. I might wait for a future PR to change this code, though, since this PR is only copying-pasting these lines, not originating them.

gbrooksby (approved):
My question has been answered so this PR looks okay to merge to me.

PR #1482, opened 2021-12-09
Add direct KLV -> JSON exporter

daniel-riehm:
This PR creates a JSON exporter for a vector of `klv_packet`s using `cereal`. It can handle all types contained in currently supported KLV standards, and consequently will need to be updated as that list expands. The round-trip test guarantees that no data is lost by exporting / importing this format. An example of the resultant JSON structure is included for using in another test and provides a useful example of final output.

I opted to err on the side of verbosity / human-readability for now. Should a use case be storage-limited, a more minimal (but still intercompatible) version could probably be developed on top of this one with less generous whitespace and fewer unnecessary fields. Some edits to `cereal` might be necessary to implement this, however.

The only thing I would still like to change here is the byte sequences, which serialize with full line breaks and tabs between each of the bytes, which is not ideal for either compactness _or_ readability. However, I was unable to identify any way to request inline serialization in `cereal`, so this may have to do for now.

Edit: Something I realized I forgot to implement here is external timestamps - that is, when the KLV packets appear in the video stream relative to the frames and each other. This is critical information for most use cases, however, to avoid bloating this PR, I will add that feature in a new branch once this is merged.


gbrooksby (approved):
This PR looks okay to me.

PR #1484, opened 2021-12-10
Print n, not n squared, commas for empty multicolumns

daniel-riehm:
This PR fixes a bug in which, when a vital tag is being exported to CSV and maps to `n` different columns (e.g. a point mapped to `n=3` columns, one for each dimension), but the value of the tag is empty, `n*n` blank fields will be printed instead of `n`.


mleotta (approved):

gbrooksby (commented):
This looks fine to me.

PR #1485, opened 2021-12-13
Round frame number instead of truncating

daniel-riehm:
If the frame number presentation time stamps 'wobble' a little bit  (e.g. 500, 1000, 1499, 2000, 2501), then the calculated frame number can be one less or more than it should be, resulting in merged or dropped frames. We fix this by rounding fractional frame numbers to the nearest integer instead of truncating them, compensating for 'wobble' in either direction.


mleotta (approved):

mleotta (changes requested):
The change looks good, but this is a bug fix that most certainly applies to the last release.  Please rebase this change on the `release` branch and retarget this PR to the `release` branch.  It would also be good to make a note in `doc/release_notes/release.txt` briefly describing the issue that was fixed.

gbrooksby (approved):
This PR looks okay to me, although there appears to be a problem with the Windows build.

mleotta:
It looks like the Windows CI machine has other unrelated issues.  This one line change can't possibly be problematic.  I'm going to overrule the CI and merge it.

PR #1486, opened 2021-12-14
Change GSD to center pixel instead of average

daniel-riehm:
Use the MISB standard equation for calculating GSD. If you use trigonometric identities, namely `1 + cos(2x) = cos^2(x)` and `cos(pi/2 - x) = sin(x)`, you can simplify our existing vertical formula down to `slant_range * sin( fov ) / sin( -pitch ) / image_height`. This has two main differences from the new formula: `sin( fov )` is replaced with `tan( fov )` in the new version and the **angle** is divided by `image_height` instead of the whole quantity. The `sin/tan` mixup I think was just a mistake which got through our test because `sin` and `tan` are approximately equal for small angles - and most drone footage has very small field of view due to the distances involved. The decision of where to divide by image height is a change between estimating the average GSD over the whole image and estimating the GSD of the central pixel of the image. This difference also varies with FOV; at typical angles (like in our unit test) it doesn't show up until after the sixth decimal place. But the closer you get to a fisheye-lens sort of situation, the bigger the difference is between the center and edges of the image.

Again, the changes are small enough at typical FOVs that we didn't even have to change our reference GSD value, and only changed the sixth digit of VNIIRS, which is a nonlinear function of GSD. But, it's probably best to used standardized and technically more correct methods when possible.

### DIFF
--- a/arrows/core/derive_metadata.cxx
+++ b/arrows/core/derive_metadata.cxx
@@ -231,7 +231,8 @@ double
 compute_horizontal_gsd( double slant_range, double sensor_horizontal_fov,
                         double frame_width )
 {
-  return 2.0 * slant_range * tan( sensor_horizontal_fov / 2.0 ) / frame_width;
+  return 2.0 * slant_range *
+         std::tan( sensor_horizontal_fov / frame_width / 2.0 );
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mleotta (changes requested):
I would also classify this as a bug fix.  I assume these calculations were in the last KWIVER release.  If so, we should apply this change to `release` and make a release note.

mleotta:
It looks like you'll need to cherry pick just the changes you made when rebasing onto release.  I usually use `git rebase --onto` for this.  You can specify the start and end commits of what you want to move and then the location to move it to.

daniel-riehm:
Oh oops. Looks like `derive_metadata` isn't in `release` yet, so we can't move the changes there.

mleotta:
@daniel-riehm if you can remove the release note that is not longer relevant, I will merge this to master.  Again, sorry about the bad suggestion.  This was ready to go before I made my original comments.

PR #1487, opened 2021-12-14
Create and export klv_timed_packet

daniel-riehm:
Add a struct which contains a timestamp and a packet and write the exporter for it. The timestamp / frame number, which was not previously exported, is key information if we want to identically re-encode the KLV data later.


gbrooksby (approved):
This PR looks okay to merge.

PR #1491, opened 2021-12-15
Change headers to point to new KLV system

daniel-riehm:
`master` currently doesn't build because the JSON exporter branch which was just merged points at the old names for the new KLV system. This fixes that problem.


gbrooksby (commented):
Looks good to me.

PR #1494, opened 2021-12-15
Standardizing comment and separator line formats in ocv, proj, and qt directories.

gbrooksby:


arrows/ocv/bounding_box.h
```diff
@@ -19,30 +17,26 @@ namespace kwiver {
...
-/**
- * @brief Convert CvRect to bounding_box
- *
- * This operator converts a CvRect to a kwiver bounding box.
- *
- * @param vbox CvRect to convert
- *
- * @return Equivalent bounding box.
- */
+/// @brief Convert CvRect to bounding_box
```
hdefazio, line 20: Is there a difference between @ or \ for these?

gbrooksby: Either symbol works with Doxygen.  [See the documentation here.](https://www.doxygen.nl/manual/commands.html)

mleotta: Both work, but I believe we also decided to favor `\` over `@` as part of our KWIVER style.  We also decided to use the auto `\brief` where the first line of documentation, separated by a blank comment line, is automatically treated as the `\brief` comment.

However, neither of these changes is necessary in this PR.  I'd rather get all the `/**` converted to `///` first and then revisit the `\` vs `@` and auto `\brief` in a second round.

PR #1499, opened 2021-12-17
Implement ST0601 tag 122: country codes

daniel-riehm:
An implementation of this tag, one of perhaps 20 in ST0601 which are not yet implemented, was requested. This PR implements it according to the standard. One thing this PR doesn't do is decode all the various country coding methods (i.e. `FRA` means `France` in the GENC Three-Letter system), since there are ~15 of those, and 15 times 200+ countries equals a lot of typing. For now, the users of this field or the three similar fields in ST0102 will have to wrangle that translation themselves. This could possibly change in the future should the need be sufficiently great.

### DIFF
--- a/arrows/klv/klv_0601.cxx
+++ b/arrows/klv/klv_0601.cxx
@@ -4,7 +4,6 @@
 
 #include "klv_0601.h"
 
-#include "klv_0102.h"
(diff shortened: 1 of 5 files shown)


gbrooksby (approved):
This PR looks okay to me.

PR #1501, opened 2022-01-06
Add video_settings

daniel-riehm:
This PR is meant to address two design problems relating to video writers.

First, each video writer implementation may **require** different parameters to start encoding a video. For example, FFmpeg needs to know the video size and frame rate before you give it the first image. In contrast, an image list writer doesn't necessarily need the images to be the same size all the way through, and giving it a frame rate doesn't even make sense. Some hypothetical future encoder may require additional implementation-specific information, like which encoding format to use.

Second, video writer implementations may **allow** for any number of quality/format parameters to be tweaked, even if they're not required to be set by the user. If we just go with "reasonable defaults", we may end up re-encoding a high-fidelity original video with much lower compression, or a very-compressed original video with a much larger file size than necessary. Ideally, we would be able to transfer all the fiddly settings from a video input to a video output, but it's not easy do that without knowing implementation details.

These problems make setting the parameters for `video_input.open()` tricky - do we have a bunch of parameters, most of which may go unused, or very few parameters, allowing little flexibility? So here we decide to use RTTI. `video_settings` is a base class whose only purpose right now is to be specialized into an implementation-specific version (`ffmpeg_video_settings`). A given `video_input` can export its settings into its respective derived class, cast it to a generic `video_settings`, and these settings can be handed off to a `video_output` in the future. If that `video_output` can read the implementation-specific settings, it will do so. Otherwise, it will use reasonable defaults.

In the future, we may add virtual methods to `video_input` so the user can generically get/set common parameters like frame dimensions and frame rate. But not in this PR.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_input.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_input.cxx
@@ -7,6 +7,7 @@
 
 #include "ffmpeg_init.h"
 #include "ffmpeg_video_input.h"
+#include "ffmpeg_video_settings.h"
 
 #include <arrows/klv/klv_convert_vital.h>
 #include <arrows/klv/klv_demuxer.h>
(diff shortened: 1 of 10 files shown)


arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -7,6 +7,7 @@
 
 #include "ffmpeg_init.h"
 #include "ffmpeg_video_input.h"
```
gbrooksby, line 9: This is not a change for this PR, but shouldn't `ffmpeg_video_input.h` be the first include since this is `ffmpeg_video_input.h`?

gbrooksby (approved):
LGTM.  I made a comment in one place that is unrelated to the changes in this PR.

mleotta (approved):
Looks good to me (only skimmed at a high level)

arrows/ffmpeg/ffmpeg_video_input.cxx
daniel-riehm: Probably? The reason for this order was probably alphabetization, but I'm not 100% sure which takes priority in KWIVER style.

PR #1504, opened 2022-01-10
Ensure FFmpeg drains buffered frames

daniel-riehm:
Our current FFmpeg video input treats the failure of `av_read_frame()` as the end of the video. However, there still may be frames buffered in the FFmpeg decoder, e.g. if the last frame of the input is not a keyframe, the decoder may be waiting in vain for the next keyframe packet before emitting another frame image. The correct behavior when the last packet has been read from file is to then call `av_send_packet()` with a `NULL` packet, which signals the decoder to enter 'draining mode'. Any remaining decodable frames can then be read out, preventing the previously observed loss of a few frames at the end of some videos.


gbrooksby (approved):
LGTM.

PR #1505, opened 2022-01-11
Add FFmpeg video writer

daniel-riehm:
Create a proof-of-concept FFmpeg video writer. There are several features/functions/cases that are unimplemented, but the basic structure and core functionality is there. It writes a sequence of images into a single video file, but cannot handle metadata yet. Credit to Matt Woehlke for the `video_output` algorithm interface.

This PR comes with a round-trip unit test which reads our test video, writes it back out, and reads it back in again. Each frame is then compared with the original to verify it is approximately the same. Note that, even with the copied-over compression settings, the resulting video is of noticeably lower quality, since it has now been compressed twice. If we have a use case where we want to exactly re-encode the video images without losing quality (only editing the metadata), some system of transferring the raw video packets from input to output will have to be devised - perhaps bundling them with the frame images?


arrows/ffmpeg/ffmpeg_video_output.cxx
```diff
@@ -0,0 +1,498 @@
...
+::write_remaining_packets()
+{
+  // Enter "draining mode" - i.e. signal end of file
+  avcodec_send_frame( codec_context, nullptr );
+
+  while( write_next_packet() ) {}
+}
+
+ffmpeg_video_output
+::ffmpeg_video_output() : d{ new impl{} } {
```
gbrooksby, line 117: Should the final `{` on this line be on the next line by itself?

gbrooksby (approved):
This PR look okay to me.  I made one comment on formatting.

arrows/ffmpeg/ffmpeg_video_output.cxx
daniel-riehm: Good catch! I must have forgotten to re-run uncrustify. Fixed.

PR #1510, opened 2022-01-20
Make klv_to_vital_metadata() return sptr

daniel-riehm:
Change the klv-to-vital conversion function so that it returns a shared pointer to a metadata object instead of by value. This is required by future changes which may make the `metadata` class `virtual`.


gbrooksby (approved):
LGTM.

PR #1512, opened 2022-01-24
Refactor and expand MISP timestamp code

daniel-riehm:
Refactor the MISP timestamp functions to look more like the KLV read/write functions. Read (and newly write) the timestamp status byte as well as the raw timestamps. Relevant standard: [ST0603](https://nsgreg.nga.mil/doc/view?i=4501)


gbrooksby (approved):
LGTM

PR #1513, opened 2022-01-25
Add klv_set::print_typed()

daniel-riehm:
Add a function which prints `klv_set`s in a more friendly manner, for example printing the name of each tag instead of just the number. This print function is purely for internal/debugging purposes (e.g. error messages on unit test failure) and shouldn't affect the functionality of anything.


gbrooksby (approved):
LGTM

PR #1514, opened 2022-01-29
Fixes setting the flags for code coverage

chetnieter:
The KWIVER_CPP_COVERAGE option was not getting parsed correctly and it was not setting the flags correctly to enable testing code coverage with gcov. This commit fixes those issues. Prior to this to get the code coverage testing to work you had to set the correct flags manually.


mleotta (approved):
Looks okay to me, but I don't know much about `kwiver_check_compiler_flag` or why that wouldn't work.

chetnieter:
> 



> Looks okay to me, but I don't know much about `kwiver_check_compiler_flag` or why that wouldn't work.

It is apparently a known limitation of the the underlying `check_cxx_compiler_flag` macro in cmake related to how `-fprofile-arcs` is both a compile and link flag.

https://cmake.org/Bug/view.php?id=7590

PR #1515, opened 2022-02-09
Fix 0104 date parsing crash

daniel-riehm:
Change behavior when a ST0104 date fails to parse from crashing via uncaught exception to printing an error message and continuing on.

### DIFF
--- a/arrows/klv/klv_convert_vital.cxx
+++ b/arrows/klv/klv_convert_vital.cxx
@@ -138,6 +138,31 @@ parse_geo_point( klv_timeline const& klv_data,
 
 // ----------------------------------------------------------------------------
 void
+klv_0104_parse_datetime_to_unix(
+  klv_timeline const& klv_data, uint64_t timestamp, kv::metadata& vital_data,
+  klv_lds_key klv_tag, kv::vital_metadata_tag vital_tag )
+{
+  constexpr auto standard = KLV_PACKET_MISB_0104_UNIVERSAL_SET;
+
+  auto const datetime = klv_data.at( standard, klv_tag, timestamp );
+  if( datetime.valid() )
+  {
+    try
+    {
+      auto const value = datetime.get< std::string >();
+      vital_data.add(
+        vital_tag, kv::std_0104_datetime_to_unix_timestamp( value ) );
+    }
+    catch( kv::metadata_exception const& e )
+    {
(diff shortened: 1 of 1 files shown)


daniel-riehm:
@hdefazio

arrows/klv/klv_convert_vital.cxx
```diff
@@ -241,19 +241,33 @@ klv_0104_to_vital_metadata( klv_timeline const& klv_data, uint64_t timestamp,
...
+    {
+      auto const value = event_start_datetime.get< std::string >();
+      vital_data.add< kv::VITAL_META_EVENT_START_TIMESTAMP >(
+          kv::std_0104_datetime_to_unix_timestamp( value ) );
+    }
+    catch( kv::metadata_exception const& e )
+    {
+      LOG_ERROR( kv::get_logger( "klv" ), e.what() );
+    }
   }
```
hdefazio, line 271: This seems repetitive with the if statement above - possibly make into a function?

arrows/klv/klv_convert_vital.cxx
```diff
@@ -236,25 +261,13 @@ klv_0104_to_vital_metadata( klv_timeline const& klv_data, uint64_t timestamp,
...
-  // Parse the datetime string into a UNIX microsecond timestamp
-  auto const event_start_datetime =
-    klv_data.at( standard, KLV_0104_START_DATETIME, timestamp );
-  if( start_datetime.valid() )
-  {
-    auto const value = event_start_datetime.get< std::string >();
-    vital_data.add< kv::VITAL_META_EVENT_START_TIMESTAMP >(
-        kv::std_0104_datetime_to_unix_timestamp( value ) );
-  }
+  // Parse the datetime strings into UNIX microsecond timestamps
```
hdefazio, line 264: This comment might be better as the function description. And maybe rename the function to `klv_0104_parse_datetime_to_unix` so the comment isn't needed here?

daniel-riehm: I use these comments to break up this large function into "sections" to improve readability, sort of like the `//---` function separators. So I would want to keep a comment here, even if not strictly necessary for its own sake. But I can change the function name if you think it improves comprehension at function definition.

hdefazio: yep, that's fine. This looks good to me now.

PR #1516, opened 2022-02-10
Created KWIVER_ENABLE_DBOW2 option

willdunklin:
DBoW2 was automatically included with the inclusion of OpenCV. Created KWIVER_ENABLE_DBOW2 option which defaults to ON.


CMake/kwiver-depends-DBoW2.cmake
```diff
@@ -0,0 +1,6 @@
+# Optional confgure DBoW2 dependency
+
+option( KWIVER_ENABLE_DBOW2
+  "Enable DBoW2 dependent code and plugins"
+  ON
```
chetnieter, line 5: Suggest changing `ON` to `${fletch_ENABLED_OpenCV}`.

CMake/kwiver-depends-DBoW2.cmake
```diff
@@ -0,0 +1,6 @@
...
+  )
```
chetnieter, line 6: Please add a section that includes `find_package( OpenCV REQUIRED )` if dbow2 is enabled. It should be similar to https://github.com/Kitware/kwiver/blob/master/CMake/kwiver-depends-OpenCV.cmake#L9-L10.

chetnieter (changes requested):
Other than managing the case where OpenCV is not available things look good.

chetnieter:
jenkins test this please

chetnieter:
jenkins test this please

PR #1517, opened 2022-02-10
Ensure NaN KLV values compare equal

daniel-riehm:
While C++ treats all `NaN` values as not equal to each other, for the purposes of KLV it is useful for `klv_value`s with `NaN`s to compare equal; else two copies of the same KLV packet with some `NaN` value buried somewhere can never compare as equal. This change makes testing/validation much easier in such cases.

### DIFF
--- a/arrows/klv/klv_value.cxx
+++ b/arrows/klv/klv_value.cxx
@@ -24,6 +24,30 @@ namespace arrows {
 
 namespace klv {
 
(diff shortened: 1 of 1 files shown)


daniel-riehm:
@hdefazio

hdefazio:
@daniel-riehm LGTM

PR #1518, opened 2022-02-14
Expand python api bindings for mesh objects 

chetnieter:
Expands the python api bindings for mesh objects and mesh related utilities.


johnwparent (approved):

chetnieter:
> After a quick pass this generally looks good. I don't believe this is the case here, but in general we need to try to avoid calling the Pybind11 API within the scope of the OMP pragma.
> 
> Approval pending CI passing.

I am going to hold off on merging. I don't need the method that uses OpenMP so I am going to put off those commits until I can refactor them so OMP pragma can be moved out of the Pybind11 API.

PR #1520, opened 2022-02-23
Fix darknet arrow build

dstoup:
The following patches are required to get the darknet arrow to build. The first commit assumes Darknet_DIR is being used and points to a valid DarknetConfig.cmake.

### DIFF
--- a/arrows/darknet/CMakeLists.txt
+++ b/arrows/darknet/CMakeLists.txt
@@ -14,7 +14,7 @@ set( sources
   darknet_trainer.cxx )
 
 set( darknet_linked_libs
-  darknet
+  Darknet::dark
   kwiversys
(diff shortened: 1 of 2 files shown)


neal-siekierski:
Thanks! Would you be able to check if this causes any issues building or running the Darknet-related parts of [VIAME](https://github.com/VIAME/VIAME#quick-build-instructions)? I would see dealing with any such issues as the main thing needed to land this MR.

dstoup:
@neal-siekierski I can do that once we make some decisions.

I was about to pull the PR down since I'm not sure it will be needed. It all came up because of project build changes that I don't think we're going to pursue. I will pull it for now, keep the branch and do the testing if it's something we end up needing.

Issue #1521, opened 2022-02-28
KWIVER consumers can't build against an install. Variant.hpp publicly exposed


dstoup:
Back in November a PR landed that has variant.hpp included in the header, visit.h. Visit.h is installed but variant.hpp is not which prevents any external build from succeeding against an installed KWIVER. We either need to remove variant.hpp from 'internal' or fix visit so variant.hpp is kept internal.

daniel-riehm:
Tentatively closing as issue seems to have been resolved by #1522

PR #1522, opened 2022-02-28
Make MPark variant public

daniel-riehm:
Possible fix for #1521, making `mpark::variant` public and moving it to `vital/util/`.


dstoup:
My hope is that passing CI runs indicates we shouldn't expect conflicts. I will merge now so I can get the vidtk dashboards working again.

PR #1531, opened 2022-03-08
Fix 16-bit sum parity bug

daniel-riehm:
When concatenating `klv_running_sum_16()` results, we need to take into account whether the previous checksum ended on an even or odd byte, since the checksum operates on two-byte chunks instead of one-byte chunks like a sensible checksum.


hdefazio:
LGTM

PR #1534, opened 2022-03-09
Remove points

danlipsa:


danlipsa:
Needs https://github.com/Kitware/fletch/pull/699

danlipsa:
@mleotta Please review. How do I get a user name for Jenkins to see the issue with the linux build?

mleotta:
I think the linux CI machine might just be down.  I'm seeing a similar failure on other PRs.  @dstoup or @Cookt2 can you confirm?

config/applets/color_mesh.conf
```diff
@@ -13,11 +14,14 @@ all_frames = false
...
-# Remove occluded points if parameter is true.
+# Do not color occluded points if parameter is true.
 remove_occluded = true
 
-# Remove masked points if parameter is true.
+# Do not color masked points if parameter is true.
 remove_masked = true
+
+# Remove points that are not colored.
+remove_not_colored = false
```
mleotta, line 27: @danlipsa I'm wondering if we should make this a threshold instead of just a binary option.  If that value is `0` you get the current remove if not colored behavior.  If _N_ > 0 you remove frames with _N_ or less observations.  Then maybe `-1` could be a special case meaning don't remove anything.

config/applets/color_mesh.conf
```diff
@@ -13,11 +14,14 @@ all_frames = false
...
 # We compare the depth buffer value with the depth of the mesh point.
 # We use threshold >= 0 to fix floating point inaccuracies
-# Default value is 0, bigger values will remove more points.
-occlusion_threshold = 0.0
+# Default value is 1, bigger values will remove more points.
+occlusion_threshold = 1
 
-# Remove occluded points if parameter is true.
+# Do not color occluded points if parameter is true.
 remove_occluded = true
```
mleotta, line 21: I wonder if this should now be called `color_occluded` and have the opposite meaning as it currently does.  That is
`remove_occluded = true` becomes `color_occluded = false`

config/applets/color_mesh.conf
```diff
@@ -13,11 +14,14 @@ all_frames = false
...
 
-# Remove masked points if parameter is true.
+# Do not color masked points if parameter is true.
 remove_masked = true
```
mleotta, line 24: Likewise this should probably be `color_masked` and take on the inverse to its current meaning.

mleotta (changes requested):

config/applets/color_mesh.conf
danlipsa: Yes, I was thinking exactly that.

config/applets/color_mesh.conf
danlipsa: Sounds good

mleotta:
@danlipsa this is failing because VTK 9.1 is not the default on the CI machines.  Either I need to get consensus to make it the default or we need to conditionally disable these new features if building with older VTK.

danlipsa:
I see. Yes, the filter I use is not available for VTK < 9.1

mleotta:
It seems that we need to update this file:
https://github.com/Kitware/kwiver/blob/master/CMake/kwiver-depends-VTK.cmake
to enforce that KWIVER requires at least VTK 9.1.  Currently we can build against > 8.0.

The alternative is to find away to disable the new features you added when using older VTK or to use an alternate implementation that does not need VTK 9.1.

danlipsa:
@mleotta Are the CIs updated?

danlipsa:
jenkins test this please

danlipsa:
@mleotta So now I compile color_mesh only if we have VTK >= 9.1. This way we keep visgui working for older versions of VTK. Would that work?

arrows/vtk/applets/color_mesh.cxx
```diff
@@ -299,7 +305,7 @@ class color_mesh::priv : public kwiver::arrows::vtk::mesh_coloration
       "We use threshold >= 0 to fix floating point inaccuracies "
       "Default value is 0, bigger values will remove more points.");
     config->set_value(
-      "remove_occluded", true,
+      "color_occluded", false,
       "Remove occluded points if parameter is true.");
```
mleotta, line 309: update description string "Remove" --> "Color"

arrows/vtk/applets/color_mesh.cxx
```diff
@@ -238,6 +235,15 @@ class color_mesh::priv : public kwiver::arrows::vtk::mesh_coloration
...
     }
 
+    // set variables from the config
+    frame_sampling_ = config_->get_value("frame_sampling", 1);
+    frame_ = config_->get_value("frame", -1);
+    all_frames_ = config_->get_value("all_frames", false);
+    occlusion_threshold_ = config_->get_value("occlusion_threshold", 0.0);
+    color_occluded_ = config_->get_value("color_occluded", false);
+    color_masked_ = config_->get_value("color_masked", false);
+    remove_color_count_less_equal_ = config_->get_value("remove_color_count_less_equal", true);
```
mleotta, line 245: The default value needs to be update since this is no longer boolean.

More generally, the way we usually get config values is to use the variable itself as a default.  This way you don't need to keep these default values in sync in two places in the code.  So do this instead
```suggestion
    remove_color_count_less_equal_ = config_->get_value("remove_color_count_less_equal", 
                                                        remove_color_count_less_equal_);
```

and likewise for the other config settings

arrows/vtk/mesh_coloration.cxx
```diff
@@ -560,12 +584,13 @@ ::initialize_data_list( int frame_id )
...
     for( auto const& cam_itr : cam_map )
     {
-      if( ( counter++ ) % sampling_ != 0 )
+      if( ( counter++ ) % frame_sampling_ != 0 )
       {
         continue;
       }
       push_data( cam_itr, ts, has_mask );
     }
+    LOG_INFO( logger_, "Camera and image list size: " << data_list_.size() );
```
mleotta, line 593: This might be better as a `LOG_DEBUG`

mleotta (changes requested):
See inline comments.

I think disabling color-mesh for older VTK is a reasonable compromise.  Disabling only the use of `vtkRemovePolyData` within color-mesh would be better, but also a lot messier.

chetnieter:
Jenkins test this please

chetnieter:
All the build failures say 'Build step 'Execute shell' marked build as failure' at the very end despite the build succeeding and the tests passing.

mleotta:
Jenkins test this please

PR #1535, opened 2022-03-09
Add error checking to local_geo_cs IO functions

willdunklin:
Changed the return value of `read_local_geo_cs_from_file` and `write_local_geo_cs_to_file` from `void` to `bool` which signifies if the file stream is valid after execution.


vital/types/local_geo_cs.cxx
```diff
@@ -59,8 +65,16 @@ write_local_geo_cs_to_file(local_geo_cs const& lgcs,
...
+                                 << " " << lon_lat_alt[2]
+                                 << std::endl;
+    return ofs.good();
+  }
+  else
+  {
+    throw vital::file_write_exception(file_path,
+                                      "Unable to open file for writing.");
   }
+  return false;
```
mleotta, line 77: You'll never actually get here because you'll throw the exception first.  I'm not sure that this function should both throw an exception or return false on different types of errors.  I suggestion you just log an error message and return false.

mleotta (approved):

mleotta:
jenkins please test this

chetnieter:
jenkins test this please

PR #1536, opened 2022-03-09
Reorganize test data into a central location

OwenMcGee:
all "data" folders previously pointed to by an instance of the "kwiver_test_data_directory" CMake variable have been moved into a new folder "test_data". The data is still fully separated within this new folder, and repeated uses of the same or similar data still needs to be coalesced.

The file structure for shared data isn't yet decided, and there are also a few other folders containing data that should maybe also be moved into this central folder (namely "arrows/test_data")


mleotta:
jenkins test this please

mleotta:
jenkins test this please

PR #1533, opened 2022-03-09
WIP: Adding test for command line usage of mvg applet init-cameras-landmarks

OwenMcGee:
init-cameras-landmarks still takes about 70 seconds, which is very slow (might time out).

I also set the tests for both track-features and init-cameras-landmarks to point to custom config files in the data folder. Currently these are identical to the default config files, they're just there to make it easier to set custom configurations in the future.

PR #1537, opened 2022-03-10
Adding test for estimate-depth command

OwenMcGee:
Tests added for the kwiver estimate-depth command


mleotta:
For estimate-depth, try the following settings for a shorter run:
```
# Compute only two depth maps in the test
batch_depth:num_depth = 2

# Use fewer support frames 
compute_depth:num_support = 4

# Use fewer optimization iterations
compute_depth:super3d:iterations = 100

# Use lower resolution volume
compute_depth:super3d:depth_sample_rate = 2.0
```
We may not need all of these, but these should be the key parameters to make it run faster (but less accurate) which should be fine for the purposes of the applet test.

OwenMcGee:
@mleotta using those configs the test runs in ~10 seconds for me now, so I pushed a commit with all of those parameter changes.

OwenMcGee:
This branch originally also included the mvg applet tests, but they're now removed so that this PR and PR #1533 can be treated as separate

mleotta:
jenkins test this please

arrows/vtk/applets/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+include(kwiver-test-setup)
+
+add_test(NAME vtk:applets:estimate_depth
+  COMMAND kwiver estimate-depth
+            --config "${kwiver_test_data_directory}/estimate_depth_testing.conf"
+            --video-source "${kwiver_test_data_directory}/aphill_240_1fps_crf32.mp4"
+            --input-cameras-dir "${kwiver_test_data_directory}/krtd"
+            --input-landmarks-file "${kwiver_test_data_directory}/landmarks.ply")
+add_test(NAME vtk:applets:fuse_depth
```
mleotta, line 9: This test is failing on some CI machines that don't have CUDA because the default config uses CUDA.  We need to borrow some more code from TeleSculptor and update `fuse-depth` to work around this.  TeleSculptor has two different configs, [one for GPU](https://github.com/Kitware/TeleSculptor/blob/master/config/gui_integrate_depth_maps_gpu.conf) and [one for CPU](https://github.com/Kitware/TeleSculptor/blob/master/config/gui_integrate_depth_maps.conf).  They both include a [common set of parameters](https://github.com/Kitware/TeleSculptor/blob/master/config/integrate_depth_maps_common.conf) that works for both cases.  Then there is a [config option to use GPU](https://github.com/Kitware/TeleSculptor/blob/master/gui/tools/FuseDepthTool.cxx#L98-L102) that selects which config to load.  We can also [check if the cuda plugin](https://github.com/Kitware/TeleSculptor/blob/c7059fd18adad9246f21a8a3f75f835f857588cf/gui/MainWindow.cxx#L1731) is available and use that to set the default option to use GPU only if cuda is found.

arrows/vtk/applets/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+  COMMAND kwiver fuse-depth
+            --config "${kwiver_test_data_directory}/fuse_depth_testing.conf"
+            --input-cameras-dir "${kwiver_test_data_directory}/krtd"
+            --input-landmarks-file "${kwiver_test_data_directory}/landmarks.ply"
+            --input-geo-origin-file "${kwiver_test_data_directory}/geo_origin.txt"
+            --input-depths-dir "${kwiver_test_data_directory}/depths")
+add_test(NAME vtk:applets:color_mesh
```
mleotta, line 16: This test is failing on some machines because of "bad X server connection".  I think this means we need to change how VTK is built on those systems.  Probably no change to this branch but maybe a change in Fletch or just a change in packages we install on the CI machines.  @chetnieter and @danlipsa know more about this.

mleotta:
jenkins test this please

PR #1538, opened 2022-03-14
Create abstract interface for pointcloud_io

willdunklin:
Added abstract interface in vital with implementation in PDAL arrow.


vital/algo/pointcloud_io.cxx
```diff
@@ -0,0 +1,69 @@
...
+
+pointcloud_io
+::pointcloud_io()
+{
+  attach_logger( "algo.pointcloud_io" );
+}
+
+void
+pointcloud_io::
+save(vital::path_t const& filename,
```
willdunklin, line 30: The original standalone function had two signatures and I've copied them into this interface as well. I felt like this might need some refactoring so I'd appreciate feedback on whether to remove it.
The first signature here isn't used in any KWIVER applets and is solely used on [this line in TeleSculptor](https://github.com/Kitware/TeleSculptor/blob/master/gui/MainWindow.cxx#L2398).

willdunklin: Answered my own question, I removed the signature used in TeleSculptor. It can be explicitly implemented in TeleSculptor where it needs to be added.

mleotta: was that the signature for saving landmarks?  If so, I would keep it in here.

arrows/vtk/applets/color_mesh.cxx
```diff
@@ -480,7 +480,9 @@ class color_mesh::priv : public kwiver::arrows::vtk::mesh_coloration
       }
     }
 
-    kwiver::arrows::pdal::save_point_cloud_las(output_path, lgcs, points, colors);
+    kwiver::arrows::pdal::pointcloud_io pc_io;
```
mleotta, line 483: This is perfectly fine for this PR, but the ultimate goal of this refactoring is to dynamically construct the `pc_io` object from a config file that contains the string "pdal" and thereby remove any direct dependency of vtk applets on pdal.  That change can come in a follow-up PR.

arrows/vtk/applets/fuse_depth.cxx
```diff
@@ -418,7 +418,9 @@ class fuse_depth::priv
           inPts->GetPoint(i, points[i].data());
         }
 
-        kwiver::arrows::pdal::save_point_cloud_las(output_mesh_file, lgcs, points, colors);
+        kwiver::arrows::pdal::pointcloud_io pc_io;
```
mleotta, line 421: same comment as in color_mesh.cxx

arrows/pdal/register_algorithms.cxx
```diff
@@ -0,0 +1,42 @@
...
+                    "Write pointclouds to .las file using PDAL." )
+    .add_attribute( kwiver::vital::plugin_factory::PLUGIN_MODULE_NAME, module_name )
+    .add_attribute( kwiver::vital::plugin_factory::PLUGIN_VERSION, "1.0" )
+    .add_attribute( kwiver::vital::plugin_factory::PLUGIN_ORGANIZATION, "Kitware Inc." )
+    ;
+
+  vpm.mark_module_as_loaded( module_name );
+}
+
+} // end namespace gdal
```
mleotta, line 40: gdal --> pdal

arrows/pdal/register_algorithms.cxx
```diff
@@ -0,0 +1,42 @@
...
+register_factories( kwiver::vital::plugin_loader& vpm )
+{
+  static auto const module_name = std::string( "arrows.pdal" );
+  if (vpm.is_module_loaded( module_name ) )
+  {
+    return;
+  }
+
+  // add factory               implementation-name       type-to-create
+  auto fact = vpm.ADD_ALGORITHM( "pdal", kwiver::arrows::pdal::pointcloud_io );
```
mleotta, line 29: This is the old deprecated way of registering algorithms.  The new way pulls this information from that `PLUGIN_INFO` call in the class itself and has simpler syntax.  See example here:
https://github.com/Kitware/kwiver/blob/master/arrows/mvg/register_algorithms.cxx#L55

arrows/pdal/pointcloud_io.h
```diff
@@ -12,25 +12,30 @@
...
-save_point_cloud_las(vital::path_t const& filename,
-                     vital::local_geo_cs const& lgcs,
-                     std::vector<vital::vector_3d> const& points,
-                     std::vector<vital::rgb_color> const& colors = {});
+class KWIVER_ALGO_PDAL_EXPORT pointcloud_io
+  : public vital::algo::pointcloud_io
+{
+public:
+  PLUGIN_INFO( "pdal",
+               "Use PDAL to write pointclouds in .las format." );
```
mleotta, line 26: We might want to just want to say "Use PDAL to write point clouds" in anticipation of generalizing to other formats, which should be really easy, but not needed in this PR.

mleotta: FYI I think it is as simple as creating a kwiver config variable on this algorithm that takes a string and uses it instead of "writers.las" in 
```c++
::pdal::Stage *writer = factory.createStage("writers.las");
```
some additional error checking will be needed as well to make sure the writer is valid.  Options are here: https://pdal.io/stages/writers.html

Again, let's not do this in this PR.

mleotta (changes requested):
See inline comments.

vital/algo/pointcloud_io.h
```diff
@@ -0,0 +1,71 @@
...
+  /// \throws kwiver::vital::path_not_a_file Thrown when the given path does
+  ///    not point to a file (i.e. it points to a directory).
+  ///
+  /// \param filename the path to the file to save
+  /// \param points the vector of points to write
+  /// \param colors optional vector of colors corresponding to the points
+  void save(std::string const& filename,
+            std::vector<vital::vector_3d> const& points,
+            std::vector<vital::rgb_color> const& colors = {});
+
```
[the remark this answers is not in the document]
willdunklin: Any suggestion on naming the function here?

arrows/pdal/pointcloud_io.cxx
```diff
@@ -22,38 +23,20 @@ namespace kwiver {
...
 {
   namespace kv = kwiver::vital;
-  kv::logger_handle_t logger( kv::get_logger( "save_point_cloud_las" ) );
+  kv::logger_handle_t logger( kv::get_logger( "pdal_pointcloud_io_save_" ) );
 
   if( !colors.empty() && colors.size() != points.size() )
   {
-    throw vital::invalid_value("save_point_cloud_las: number of colors provided does "
-                               "not match the number of points");
+    throw vital::invalid_value("pdal_pointcloud_io_save_: number of colors "
```
mleotta, line 38: `pdal_pointcloud_io_save_` --> `pdal::pointcloud_io::save_`

vital/algo/pointcloud_io.h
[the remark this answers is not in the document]
mleotta: You can also call the other function `save`.  It won't conflict if it takes different data structures.

arrows/pdal/pointcloud_io.h
```diff
@@ -12,25 +12,34 @@
 
 #include <vital/types/local_geo_cs.h>
```
mleotta, line 13: We should probably move `pointcloud_io.h` and `pointcloud_io.cxx` into an `algo` subdirectory under `pdal` now that these files are an implementation of an algorithm in `vital/algo`.  This is the new pattern we are moving to help keep things tidy.  See the `mvg` arrow as an example.

mleotta (changes requested):
Looking much improved.  See a few more inline comments.

arrows/pdal/algo/pointcloud_io.cxx
```diff
@@ -31,11 +31,11 @@ save_(vital::path_t const& filename,
       std::vector<vital::rgb_color> const& colors) const
 {
   namespace kv = kwiver::vital;
-  kv::logger_handle_t logger( kv::get_logger( "pdal_pointcloud_io_save_" ) );
+  kv::logger_handle_t logger( kv::get_logger( "pdal::pointcloud_io::save_" ) );
```
mleotta, line 34: I missed this before.  It should be "arrows.pdal.pointcloud_io".

It's not required, but typically we use `attach_logger("arrows.pdal.pointcloud_io")` in the constructor of this class and then access the logger latter with `logger()` as you did in the base class.

arrows/pdal/tests/test_pointcloud_io.cxx
```diff
@@ -34,19 +36,72 @@ class pointcloud_io : public ::testing::Test
 };
```
mleotta, line 37: You should also add a "create" test to confirm that you can construct this object from a string using the plugins.  See example here: https://github.com/Kitware/kwiver/blob/master/arrows/mvg/tests/test_integrate_depth_maps.cxx#L29-L37

mleotta (changes requested):
Almost there.  I found a couple more things.

arrows/pdal/tests/test_pointcloud_io.cxx
```diff
@@ -34,19 +36,79 @@ class pointcloud_io : public ::testing::Test
...
 
 // ----------------------------------------------------------------------------
-TEST_F(pointcloud_io, save_geo_origin) {
+TEST_F(pointcloud_io, create)
+{
+  auto pc_io = new kwiver::arrows::pdal::pointcloud_io();
```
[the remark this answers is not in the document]
mleotta: The idea is to test that you can construct the "pdal" object from the base class `create` function using the plugins.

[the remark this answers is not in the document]
willdunklin: My bad, I'll switch it over.

mleotta:
jenkins test this please

mleotta (approved):

PR #1539, opened 2022-03-14
Bugfix: set_origin referencing geo_point with wrong crs

willdunklin:
Background: `local_geo_cs::set_origin(const geo_point& origin)` is a function that modifies the `geo_origin_` to use a different coordinate referencing system (`crs`). Calling `geo_point::location(int crs)` returns a geopoint indexed by the `crs` variable. 

The existing code for `set_origin` creates a new `crs` to change the coordinate referencing system, but mistakenly uses that new `crs` to index the prexisting point. This leads to unexpected data (most likely garbage) being put in the `geo_point()` constructor.

### DIFF
--- a/vital/tests/test_local_geo_cs.cxx
+++ b/vital/tests/test_local_geo_cs.cxx
@@ -50,7 +50,7 @@ TEST_F(local_geo_cs, read_local_geo_cs_from_file)
   valid = read_local_geo_cs_from_file(lgcs, geo_origin_valid);
   EXPECT_TRUE(valid);
   // Expect the local geo cs to be valid
-  kv::vector_3d origin = lgcs.origin().location(crs);
+  kv::vector_3d origin = lgcs.origin().location();
   EXPECT_NEAR(origin[0], -77.3578172263, 1e-8);
   EXPECT_NEAR(origin[1], 38.1903504278, 1e-8);
   EXPECT_NEAR(origin[2], -68.0169758322, 1e-8);
@@ -60,7 +60,7 @@ TEST_F(local_geo_cs, read_local_geo_cs_from_file)
   valid = kv::read_local_geo_cs_from_file(lgcs, geo_origin_invalid);
   EXPECT_FALSE(valid);
   // Expect the origin to be unchanged
-  origin = lgcs.origin().location(crs);
+  origin = lgcs.origin().location();
   EXPECT_NEAR(origin[0], 0.0, 1e-8);
   EXPECT_NEAR(origin[1], 0.0, 1e-8);
   EXPECT_NEAR(origin[2], 0.0, 1e-8);
--- a/vital/types/local_geo_cs.cxx
+++ b/vital/types/local_geo_cs.cxx
@@ -35,7 +35,7 @@ local_geo_cs
   vector_3d lon_lat_alt = origin.location( SRID::lat_lon_WGS84 );
   auto zone = utm_ups_zone( lon_lat_alt );
   int crs = (zone.north ? SRID::UTM_WGS84_north : SRID::UTM_WGS84_south) + zone.number;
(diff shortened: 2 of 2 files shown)


vital/types/local_geo_cs.cxx
```diff
@@ -62,7 +62,7 @@ write_local_geo_cs_to_file(local_geo_cs const& lgcs,
   kwiver::vital::logger_handle_t logger(kwiver::vital::get_logger(
     "write_local_geo_cs_to_file"));
   // write out the origin of the local coordinate system
-  auto lon_lat_alt = lgcs.origin().location( SRID::lat_lon_WGS84 );
```
mleotta, line 65: This explicit ` SRID::lat_lon_WGS84` is intentional.  It says "give me the location of the origin in WGS84 coordinate".  This is to ensure that the local_geo_cs file is always in lat, lon, alt (WGS84).  By default the origin often is in WGS84, so your change will still work as expected.  However, it is possible to create an origin point in a different coordinate system, so this conversion is necessary.

vital/types/local_geo_cs.cxx
```diff
@@ -35,7 +35,7 @@ ::set_origin(const geo_point& origin)
   vector_3d lon_lat_alt = origin.location( SRID::lat_lon_WGS84 );
   auto zone = utm_ups_zone( lon_lat_alt );
   int crs = (zone.north ? SRID::UTM_WGS84_north : SRID::UTM_WGS84_south) + zone.number;
-  geo_origin_ = geo_point(origin.location(crs), crs);
```
mleotta, line 38: passing `crs` is required for this to work in the general case.  You must first convert the origin from whatever space it was in to this new `crs`, then you construction a new `geo_point` where this `crs` is the default representation.

vital/tests/test_local_geo_cs.cxx
```diff
@@ -50,7 +50,7 @@ TEST_F(local_geo_cs, read_local_geo_cs_from_file)
   valid = read_local_geo_cs_from_file(lgcs, geo_origin_valid);
   EXPECT_TRUE(valid);
   // Expect the local geo cs to be valid
-  kv::vector_3d origin = lgcs.origin().location(crs);
+  kv::vector_3d origin = lgcs.origin().location();
```
mleotta, line 53: This change only works because the default `crs` happens to be `kv::SRID::lat_lon_WGS84`

mleotta (changes requested):
I don't understand what the bug is here.  The existing code looks correct to me and the change looks wrong.  Maybe the right thing to do is to create a unit test that illustrates the bug and is currently failing.

willdunklin:
Ah, I critically missed the behavior of `location(crs)` if the `crs` parameter is [not in the `geo_point` map](https://github.com/Kitware/kwiver/blob/master/vital/types/geo_point.cxx#L72). I thought that `set_origin` wouldn't copy undefined `crs` indexed data but it does, which means you can do something like:
```c++
auto source = vital::geo_point([0, 0, 0], 1);
auto dest = vital::geo_point(source.location(2), 2);

cout << dest.crs() << " " << dest.location(dest.crs()) << endl; // dest.location() is equivalent 
cout << source.crs() << " " << dest.location(source.crs()) << endl;
```
which will produce the output:
```
2 [12.34, -2.02, 100] <- where this array is some garbage data
1 [0, 0, 0]
```

My previous assumption was that `dest.location(source.crs())` would be undefined, so the criticality of this "bugfix" is much less severe than I thought. 

The point of this function is still a bit unclear to me as it modifies the default `crs` (the one getting called by `geo_point_.crs()`) while that index is still pointing at undefined data. I may be completely wrong, but it feels like `set_origin` should probably copy the data from the default `origin.location()` to the new `geo_point_` meaning:
```
dest.location(dest.crs()) -> [0, 0, 0]
```

PR #1541, opened 2022-03-15
Add interval_map::weak_set()

daniel-riehm:
Add a `weak_set()` method for the `interval_map` class. This is a variation on the existing `set()` method, which sets the map's value for some period of time, replacing any value(s) already present during that time. `weak_set()` similarly sets the maps' value, but does not replace any existing values, instead only setting the new value during any gaps where no value is currently set.

The immediate motivation for this change is that some of our test videos present KLV packets in the wrong order. Currently we just drop any packets that "jump backwards in time", but better handling of such situations is possible with `weak_set()`.

### DIFF
--- a/vital/util/tests/test_interval_map.cxx
+++ b/vital/util/tests/test_interval_map.cxx
@@ -377,6 +377,157 @@ TEST ( interval_map, set )
 }
 
 // ----------------------------------------------------------------------------
+TEST ( interval_map, weak_set )
+{
+  using map_type = interval_map< int, int >;
+  auto const basis_map = map_type{
+    { { 0, 5 }, 0 },
+    { { 5, 10 }, 1 },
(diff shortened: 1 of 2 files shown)


vital/util/tests/test_interval_map.cxx
```diff
@@ -376,6 +376,157 @@ TEST ( interval_map, set )
...
+  {
+    auto test_map = basis_map;
+    test_map.weak_set( { 11, 11 }, 20 );
+    EXPECT_EQ( basis_map, test_map );
+  }
+
+  // Point interval - outside
+  {
+    auto test_map = basis_map;
+    test_map.set( { -1, -1 }, 20 );
```
mleotta, line 400: Should this be `weak_set`?

mleotta (approved):

PR #1543, opened 2022-03-17
Adding test for command line usage of core applet dump-klv

OwenMcGee:
Adding tests for the `kwiver dump-klv` command.

It appears that the dump-klv default config is missing a definition for the `metadata_serializer:type`, making it impossible to run without a custom config file. There is also no command line option to set the value.


mleotta:
We should update the default config for `dump-klv` to have a valid `metadata_serializer:type` and/or make sure that it works without a serializer (e.g. just dumps messages to stdout).  A command line switch or detection of serializer type based on output file extension would also be helpful.  I nominate @daniel-riehm to make these improvements in a separate PR since these are related to his work with KLV.

mleotta:
@OwenMcGee does render-mesh not work because you don't have appropriate test data?  Or is it something else?

OwenMcGee:
@mleotta right now I believe I just don't have the appropriate test data, I'm still somewhat figuring out how the commands work

mleotta:
jenkins test this please

PR #1545, opened 2022-03-18
Adding test for render-mesh command

OwenMcGee:
Currently the command seems to hang on the "reading mesh" step, even with proper input.

Also, providing a config file input seems to result in an error: `std::bad_cast`


mleotta:
@OwenMcGee I'm not sure what these kwrobot-v1 errors are all about, but maybe it's related to Windows line endings again.  Can you try converting the ASCII PLY file to use unix line endings and see if that makes a difference?

mleotta:
jenkins please test this

mleotta:
jenkins please test this

OwenMcGee:
@mleotta the file endings were already fine, and I believe you're using the command wrong (it should be "jenkins test this please")

mleotta:
Ah, I can never remember which way it is.

mleotta:
jenkins test this please

PR #1551, opened 2022-03-22
Disable offending tests on Windows

johnwparent:
Currently a subset of tests are failing on Windows. For now we should disable them, so that is does not inhibit the CI moving forward.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,18 @@ OPTION( KWIVER_ENABLE_C_BINDINGS               "Enable C bindings libraries" OFF
 CMAKE_DEPENDENT_OPTION( KWIVER_ENABLE_PYTHON   "Enable python code" OFF
   KWIVER_ENABLE_C_BINDINGS OFF )
 
+if (WIN32)
+  # Set default for python tests to OFF on windows for now
+  CMAKE_DEPENDENT_OPTION( KWIVER_ENABLE_PYTHON_TESTS "Enable python tests" OFF
+    "KWIVER_ENABLE_PYTHON; KWIVER_ENABLE_TESTS" OFF )
+  if (KWIVER_ENABLE_PYTHON_TESTS)
+    message( WARNING "Python tests are not currently supported for Windows." )
+  endif()
+else()
+  CMAKE_DEPENDENT_OPTION( KWIVER_ENABLE_PYTHON_TESTS "Enable python tests" ON
+    "KWIVER_ENABLE_PYTHON; KWIVER_ENABLE_TESTS" OFF )
+endif()
+
 if (KWIVER_ENABLE_SERIALIZE_PROTOBUF)
   if (WIN32)
     message( WARNING "Protobuf serialization not supported for windows. Being disabled." )
--- a/python/kwiver/arrows/CMakeLists.txt
+++ b/python/kwiver/arrows/CMakeLists.txt
@@ -1,7 +1,7 @@
 add_subdirectory(serialize)
 add_subdirectory(core)
 
-if(KWIVER_ENABLE_TESTS)
+if(PYTHON_TEST AND NOT WIN32)
   add_subdirectory( tests )
 endif()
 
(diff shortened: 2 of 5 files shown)


mleotta (changes requested):
I think it would be better to make a `KWIVER_ENABLE_PYTHON_TESTS` CMake option that defaults to true when `KWIVER_ENABLE_TESTS` is turned on but can be turned off independently of the general `KWIVER_ENABLE_TESTS`.  Then we can manually set this option on the dashboard or turn it off later if we want to work on fixing these.

Are there some Python tests that do work on Windows and we are turning them all off?  Or are all Python tests failing on Windows?

johnwparent:
All Python tests are failing on Windows now. Additionally it appears as though one of the C++ sprokit module tests is failing because it's finding some Python modules (my best intuition from a quick glance), so we'll need to optionally pull those out as well. 
I'll refactor to use the CMake option. Wasn't sure if we wanted to go that route, as I assumed if we're enabling Python, we'd want to be testing it if we're testing the rest of Spack (outside of this use case). Alternatively if we want to go this route where we're not testing Python at all, we can always disable Python altogether in the Windows dashboard for the time being.

mleotta:
I suppose this depends on whether Python is really broken on Windows or there is just some issue getting the tests to run properly.

dstoup:
Unless thing have really changed, my look showed all Python tests were using the old testing infrastructure and hadn't been ported to GTEST

mleotta:
Is GTEST a thing for Python?  I thought that was just for C/C++ tests?

It seems these tests work on everything but Windows.

willdunklin:
@mleotta I created an option `KWIVER_ENABLE_PYTHON_TESTS` as you described which is OFF for Windows by default but can be manually overridden.
I also added a condition to disable sprokit's [test_modules.cxx](https://github.com/Kitware/kwiver/blob/master/sprokit/tests/sprokit/pipeline/test_modules.cxx) which contained the last offending test reliant on python (envvar), as John mentioned previously.

The Windows build now passes, although there is one test that is failing to run. It's in `examples` [here](https://github.com/Kitware/kwiver/blob/master/examples/external/darknet.cmake), which is notable because every other machine other than WindowsPR has `KWIVER_ENABLE_EXAMPLES=OFF`. We should probably change that cmake config on the Jenkins side of things.

CMakeLists.txt
```diff
@@ -126,6 +126,18 @@ OPTION( KWIVER_ENABLE_C_BINDINGS               "Enable C bindings libraries" OFF
 CMAKE_DEPENDENT_OPTION( KWIVER_ENABLE_PYTHON   "Enable python code" OFF
   KWIVER_ENABLE_C_BINDINGS OFF )
 
+if (WIN32)
+  # Set default for python tests to OFF on windows for now
+  CMAKE_DEPENDENT_OPTION( KWIVER_ENABLE_PYTHON_TESTS "Enable python tests" OFF
+    "KWIVER_ENABLE_PYTHON; KWIVER_ENABLE_TESTS" OFF )
```
mleotta, line 132: This `;` may need to be `AND` instead since I think this is a logical expression that need to be evaluated.  I'm not sure what happens with the `;`.  Maybe it only evaluates or the other?

CMakeLists.txt
```diff
@@ -126,6 +126,18 @@ OPTION( KWIVER_ENABLE_C_BINDINGS               "Enable C bindings libraries" OFF
...
+  if (KWIVER_ENABLE_PYTHON_TESTS)
+    message( WARNING "Python tests are not currently supported for Windows." )
+  endif()
+else()
+  CMAKE_DEPENDENT_OPTION( KWIVER_ENABLE_PYTHON_TESTS "Enable python tests" ON
+    "KWIVER_ENABLE_PYTHON; KWIVER_ENABLE_TESTS" OFF )
```
mleotta, line 138: See above about `AND` instead of `;`

mleotta (changes requested):
See comments for one minor change suggestion

CMakeLists.txt
willdunklin: The CMake documentation suggests that the `;` is sufficient https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

mleotta: Interesting.  Okay, as long as that is supported I have not objections.

mleotta (approved):

PR #1558, opened 2022-03-29
Adding tests to mesh_io

OwenMcGee:
Adding coverage to vital/io/mesh.cxx. By extension this is also adding coverage to vital/types/mesh.cxx


vital/tests/test_mesh_io.cxx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// ----------------------------------------------------------------------------
+class mesh_io : public ::testing::Test
+{
+  TEST_ARG(data_dir);
+
+  void compare_meshes(const kwiver::vital::mesh_sptr& first,
+                      const kwiver::vital::mesh_sptr& second)
+  {
```
OwenMcGee, line 34: I'm not totally sure if this is how I should be comparing meshes to check the data is the same. I believe this is exhaustive enough (at least for the data that gets set in the cube_mesh function, I'm not checking edges or normals here), but I'm not sure if there's a better way to do this

vital/tests/test_mesh_io.cxx
```diff
@@ -0,0 +1,184 @@
...
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/// \file
+/// \brief test core mesh_io functionality
+
+#include <kwiversys/SystemTools.hxx>
+#include <tests/test_gtest.h>
+#include <tests/test_scene.h>
+#include <arrows/core/mesh_operations.cxx>
```
mleotta, line 11: Probably best not to include .cxx files if we can avoid it, especially files from arrows in vital.  Can we just save a version of the triangulated cube mesh to a file and use that data?

vital/tests/test_mesh_io.cxx
```diff
@@ -0,0 +1,184 @@
...
+#include <vital/exceptions.h>
+
+#include <vital/io/mesh_io.cxx>
```
mleotta, line 14: What do we need this for?  Normally we should only include the header files.

vital/tests/test_mesh_io.cxx
```diff
@@ -0,0 +1,184 @@
...
+    kwiver::vital::file_write_exception );
+  kwiversys::SystemTools::RemoveADirectory( "temp" );
+}
+
+// ----------------------------------------------------------------------------
+TEST_F(mesh_io, invalid_input_file)
+{
+  EXPECT_THROW( kwiver::vital::check_input_file( data_dir + "/nonexistant" ),
+    kwiver::vital::file_not_found_exception );
+  EXPECT_THROW( kwiver::vital::check_input_file( data_dir + "/videos" ),
```
mleotta, line 82: Is this second `EXPECT_THROW` testing something different from the first?

vital/tests/test_mesh_io.cxx
```diff
@@ -0,0 +1,184 @@
...
+TEST_F(mesh_io, read_write_obj)
+{
+  kwiver::vital::mesh_sptr original = kwiver::testing::cube_mesh( 1.0 );
+  original->compute_vertex_normals();
+
+  std::string path = "temp/cube_mesh.obj";
+  kwiver::vital::write_obj( path, *original );
+  kwiver::vital::mesh_sptr copy = kwiver::vital::read_obj( path );
+
+  compare_meshes( original, copy );
```
mleotta, line 129: If we are testing that we can write normals then maybe the comparison function should also check if normals are found in each mesh and if so that they are the same.

mleotta (changes requested):
see inline comments.  Mostly avoid including `.cxx` files.

vital/tests/test_mesh_io.cxx
OwenMcGee: yes, one tests the exception for passing in a nonexistant file, the other tests the exception for passing in a directory

mleotta (changes requested):
This PR seems to be relocating several test data files that have nothing to do with mesh_io, one of  which is causing a conflict.  I think this renaming has already happened on `master` so I'm not sure why it is on this branch too.  Please remove unnecessary changes.  Rebase on `master` if needed to make the PR clean.

chetnieter (dismissed):
Looks good. Just need to remove the included cxx file.

vital/types/mesh.cxx
```diff
@@ -41,15 +41,16 @@ ::operator==( mesh_vertex_array_base const& other ) const
 
   const std::vector< vector_3d > these_normals = normals();
   const std::vector< vector_3d > other_normals = other.normals();
+  const double limit = 0.000001;
```
mleotta, line 44: I'm not a big fan of `operator==` being defined as an approximate equality test and having hard coded thresholds like this.  I think it would be better to redefine these functions as
```c++
bool approx_equal(mesh_vertex_array_base const& array1,
                  mesh_vertex_array_base const& array2,
                  double epsilon = 1e-6)
```
and a similar pattern for the other `operator==`.

vital/types/mesh.cxx
```diff
@@ -111,22 +112,27 @@ bool
...
 {
-  if( regularity() != other.regularity() || size() != other.size() )
+  if( size() != other.size() )
+  {
+    return false;
+  }
+  if( regularity() != other.regularity() )
   {
     return false;
   }
```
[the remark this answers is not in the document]
mleotta: I would convert the mesh with regular face array to a `mesh_face_array`.  There is a constructor from base class that that can turn any face array into a generic `mesh_face_array`:

https://github.com/Kitware/kwiver/blob/master/vital/types/mesh.h#L405

mleotta (changes requested):

chetnieter:
@OwenMcGee @mleotta Is this ready to be merged?

OwenMcGee:
@chetnieter I believe so, if everything looks good to you and Matt. I think I've addressed all the issues that have been pointed out

vital/tests/test_mesh_io.cxx
```diff
@@ -0,0 +1,141 @@
...
+
+  kwiversys::SystemTools::RemoveADirectory( "temp" );
+}
+
+// ----------------------------------------------------------------------------
+TEST_F(mesh_io, invalid_input_file)
+{
+  EXPECT_THROW( read_ply2( data_dir + "/nonexistant" ),
+    file_not_found_exception );
+  EXPECT_THROW( read_ply2( data_dir + "/videos" ), file_not_found_exception );
```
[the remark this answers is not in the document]
OwenMcGee: It's as you say, line 51 is testing for a nonexistent _file_, triggering https://github.com/Kitware/kwiver/blob/24a51941a07724b8002779d3238d60a4c3217903/vital/io/mesh_io.cxx#L56
While line 53 is testing for a real _directory_ (as "videos" is actually does exist), triggering https://github.com/Kitware/kwiver/blob/24a51941a07724b8002779d3238d60a4c3217903/vital/io/mesh_io.cxx#L61

PR #1563, opened 2022-03-29
Adding tests to mesh

OwenMcGee:
Adding coverage to vital/types/mesh.cxx


vital/tests/test_mesh.cxx
```diff
@@ -0,0 +1,236 @@
...
+    EXPECT_EQ( first_faces[i].size(), second_faces[i].size() );
+    for(unsigned int j=0; j<first_faces[i].size(); j++)
+    {
+      EXPECT_EQ( first_faces[i][j], second_faces[i][j] );
+    }
+  }
+}
+
+// ----------------------------------------------------------------------------
+TEST(mesh, group_names)
```
OwenMcGee, line 61: Since this is really targeted to test the mesh_face_array class, should the test name be "TEST( mesh_face_array, group_names )" instead, or should it stay with just "mesh", as that's the name of the file?

vital/tests/test_mesh.cxx
```diff
@@ -0,0 +1,236 @@
...
+{
+  kwiver::vital::mesh_sptr cube_mesh = kwiver::testing::cube_mesh( 1.0 );
+  std::shared_ptr<kwiver::vital::mesh_face_array> faces =
+    std::make_shared<kwiver::vital::mesh_face_array>(cube_mesh->faces());
+
+  for(unsigned int i=0; i<faces->size(); i++){
+    EXPECT_EQ( faces->group_name(i), "" );
+  }
+
+  EXPECT_EQ( faces->make_group( "testing name" ), faces->size() );
```
mleotta, line 71: This test code works and provides code coverage, but does not fully test that the groups are working.  The idea behind groups is that you first make a group, then add some faces which become part of that group, then make a new group, then add more faces that now belong to the second group, and so on.  Then you should be able to access the indices of the faces that were assigned to each group.  The current API has no easy way to partition an existing set of faces into multiple group without copying those faces into a new face array.

To test you could make a group, add 4 faces, make a second group, add 2 more faces.  Then test that the first group has the 4 expected faces and the second groups and the 2 expected faces.  Or something like that.

vital/tests/test_mesh.cxx
```diff
@@ -0,0 +1,236 @@
...
+  first_mesh->compute_face_normals();
+  std::shared_ptr<kwiver::vital::mesh_face_array> first_faces =
+    std::make_shared<kwiver::vital::mesh_face_array>(first_mesh->faces());
+  first_faces->make_group( "first name" );
+
+  kwiver::vital::mesh_sptr second_mesh = kwiver::testing::grid_mesh( 1, 1 );
+  second_mesh->compute_face_normals();
+  std::shared_ptr<kwiver::vital::mesh_face_array> second_faces =
+    std::make_shared<kwiver::vital::mesh_face_array>(second_mesh->faces());
+  second_faces->make_group( "second name" );
```
mleotta, line 99: Making groups on each face array isn't doing anything meaningful.  However, if you make a group on the first face array before merging in more faces, that's another good way to test the groups.  You should be able to construct `first_faces` such that it contains three groups that contain the faces originally found in `first_faces`, `second_faces`, and `third_faces`

vital/tests/test_mesh.cxx
```diff
@@ -0,0 +1,236 @@
...
+  EXPECT_TRUE( copy->is_init() );
+}
+
+// ----------------------------------------------------------------------------
+TEST(mesh, assignment_operator)
+{
+  kwiver::vital::mesh_sptr original = kwiver::testing::cube_mesh( 1.0 );
+  kwiver::vital::mesh_sptr copy =
+    std::make_shared<kwiver::vital::mesh>();
+  *copy = kwiver::vital::mesh( *original );
```
mleotta, line 201: Is this a copy constructor and an assignment operator?  Can you just do `*copy = *original;` or does that not work?

vital/tests/test_mesh.cxx
```diff
@@ -0,0 +1,236 @@
...
+  EXPECT_TRUE( copy->is_init() );
+  *original = kwiver::vital::mesh();
+  EXPECT_FALSE( original->is_init() );
+  EXPECT_TRUE( copy->is_init() );
+}
+
+// ----------------------------------------------------------------------------
+TEST(mesh, compute_vertex_normals)
+{
+  kwiver::vital::mesh_sptr grid_mesh = kwiver::testing::grid_mesh( 1, 1 );
```
mleotta, line 215: Maybe make this grid mesh at least (2,3).  A (1,1) mesh is just a single face.  There might be cases not covered for such a basic case.

Actually cube might be better here since different faces will have different normals.

Maybe even better is a mesh that appends a cube to a grid mesh.  The cube is a closed shape with no boundary edges which may have different special cases from the grid mesh which does have a boundary and all faces have the same normal.  It might be best to cover both.

vital/tests/test_mesh.cxx
```diff
@@ -0,0 +1,236 @@
...
+  EXPECT_FALSE( grid_mesh->has_half_edges() );
+  grid_mesh->compute_vertex_normals();
+  EXPECT_TRUE( grid_mesh->vertices().has_normals() );
+  EXPECT_TRUE( grid_mesh->has_half_edges() );
+}
+
+// ----------------------------------------------------------------------------
+TEST(mesh, compute_vertex_normals_from_faces)
+{
+  kwiver::vital::mesh_sptr grid_mesh = kwiver::testing::grid_mesh( 1, 1 );
```
mleotta, line 227: same comment as above.

mleotta (dismissed):
Overall this looks pretty good, but see inline comments with suggestions for improvement.  These tests are probably sufficient to get code coverage but some enhancements could make the unit tests better at checking expected functionality.

vital/tests/test_mesh.cxx
```diff
@@ -0,0 +1,236 @@
...
+
+// ----------------------------------------------------------------------------
+TEST(mesh, compute_vertex_normals)
+{
+  kwiver::vital::mesh_sptr grid_mesh = kwiver::testing::grid_mesh( 1, 1 );
+
+  EXPECT_FALSE( grid_mesh->vertices().has_normals() );
+  EXPECT_FALSE( grid_mesh->has_half_edges() );
+  grid_mesh->compute_vertex_normals();
+  EXPECT_TRUE( grid_mesh->vertices().has_normals() );
```
mleotta, line 220: Spot checking the value of some normal vectors might be good as well.  For the grid mesh I think they are all (0,0,1).  For the cube you should have one each of (1,0,0), (-1,0,0), (0,1,0), (0,-1,0), (0,0,1), (0,0,-1).

vital/tests/test_mesh.cxx
OwenMcGee: I expanded the current test to try and capture what you're saying here, by adding faces to a new group and verifying both groups kept the correct sizes. Let me know if it needs any more changes

chetnieter (approved):
@OwenMcGee Looks good but before we merge could you run uncrustify on the files you changed/added.

vital/types/mesh.h
```diff
@@ -261,268 +325,349 @@ class VITAL_EXPORT mesh_face_array_base
...
-  virtual unsigned int num_verts(unsigned int f) const { return static_cast<unsigned int>(faces_[f].size()); }
+  virtual unsigned int
+  num_verts( unsigned int f ) const { return static_cast< unsigned int >( 
+                                                                                         faces_
+                                                                                         [
+                                                                                           f
+                                                                                         ]
+                                                                                           .
+                                                                                           size() );
+  }
```
OwenMcGee, line 441: I've got no idea what uncrustify was thinking about with this

PR #1566, opened 2022-03-30
Prevent crash when ST0102 tag not found

daniel-riehm:
Prevent a crash in the unlikely scenario that a security metadata set is found which does not contain a security classification entry.

@hdefazio


daniel-riehm:
(Windows failures appear to be unrelated)

PR #1572, opened 2022-04-01
Import previously-private KLV development

daniel-riehm:
Add several months of commits expanding KWIVER's KLV parsing abilities, which were being developed in a private repository but now have been given permission to be released to the public.


daniel-riehm:
jenkins test this please

mleotta (approved):
This has all been previously reviewed.  I skimmed it and it seems reasonable.  Heavily covered by unit tests which are all passing, so I think it's good to go.

mleotta:
One this that is missing is a release note update in `master.txt`, but we can add this in a new PR to cover all changes to KLV that have not been documented.

PR #1574, opened 2022-04-04
WIP: Adding more tests to image.cxx

OwenMcGee:
Adding tests for the file vital/types/image.cxx

### DIFF
--- a/vital/tests/test_image.cxx
+++ b/vital/tests/test_image.cxx
@@ -63,6 +63,42 @@ struct val_incr_op
 }
 
 // ----------------------------------------------------------------------------
+TEST(image, print_pixel_traits)
+{
+  image_pixel_traits traits;
+  std::stringstream output;
+
+  traits.type = image_pixel_traits::UNKNOWN;
+  traits.num_bytes = 1;
+  output << traits;
+  EXPECT_EQ( output.str(), "Unknown_1" );
+
+  output.str(std::string());
+  traits.type = image_pixel_traits::UNSIGNED;
+  traits.num_bytes = 2;
+  output << traits;
+  EXPECT_EQ( output.str(), "Unsigned_2" );
+
+  output.str(std::string());
+  traits.type = image_pixel_traits::SIGNED;
+  traits.num_bytes = 3;
+  output << traits;
+  EXPECT_EQ( output.str(), "Signed_3" );
+
+  output.str(std::string());
+  traits.type = image_pixel_traits::FLOAT;
+  traits.num_bytes = 4;
+  output << traits;
+  EXPECT_EQ( output.str(), "Float_4" );
+
+  output.str(std::string());
+  traits.type = image_pixel_traits::BOOL;
+  traits.num_bytes = 5;
+  output << traits;
+  EXPECT_EQ( output.str(), "Bool_5" );
+}
+
+// ----------------------------------------------------------------------------
 TEST(image, default_constructor)
 {
   image img;
--- a/vital/types/image.cxx
+++ b/vital/types/image.cxx
@@ -246,8 +246,8 @@ image
 }
 
 /// Equality operator
-bool image::
-operator==( image const& other ) const
+bool image
+::operator==( image const& other ) const
 {
   return  data_         == other.data_ &&
           first_pixel_  == other.first_pixel_ &&


vital/tests/test_image.cxx
```diff
@@ -36,6 +37,33 @@ value_at( unsigned int i, unsigned int j, unsigned int k )
...
 }
 
+// ----------------------------------------------------------------------------
+void check_equal( image_memory mem_1, image_memory mem_2 )
+{
+  char* mem_1_data = (char*) mem_1.data();
+  char* mem_2_data = (char*) mem_2.data();
+
+  EXPECT_EQ( mem_1.size(), mem_2.size() );
+  for(unsigned int i=0; mem_1_data[i]!='\0'; i++)
```
mleotta, line 47: These are not actually null terminated strings, so you don't want to loop until you hit a null character you want to iterate through `mem_1.size()` bytes.

vital/tests/test_image.cxx
```diff
@@ -36,6 +37,33 @@ value_at( unsigned int i, unsigned int j, unsigned int k )
...
+  {
+    EXPECT_EQ( mem_1_data[i], mem_2_data[i] );
+  }
+}
+
+// ----------------------------------------------------------------------------
+void check_equal( image img_1, image img_2 )
```
mleotta, line 54: This should be 
```c++
template <typename T>
void check_equal( image_of<T> const& img_1, image const& img_2 )
```
to match what it is replacing below.  It is testing that these accessors for `image_of<T>` and `image` return the same thing after converting one to the other.

vital/tests/test_image.cxx
```diff
@@ -36,6 +37,33 @@ value_at( unsigned int i, unsigned int j, unsigned int k )
   return static_cast<byte>( ( ( ks * k ) + ( js * j ) + ( is * i ) ) % 255 );
 }
 
+// ----------------------------------------------------------------------------
+void check_equal( image_memory mem_1, image_memory mem_2 )
```
mleotta, line 41: You probably should pass by `const &`

vital/tests/test_image.cxx
```diff
@@ -62,6 +62,42 @@ struct val_incr_op
 
 }
 
+// ----------------------------------------------------------------------------
+TEST(image, print_pixel_traits)
+{
+  image_pixel_traits traits;
```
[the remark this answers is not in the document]
mleotta: looks okay

vital/tests/test_image.cxx
```diff
@@ -62,6 +90,79 @@ struct val_incr_op
...
+  traits.type = image_pixel_traits::BOOL;
+  traits.num_bytes = 5;
+  output << traits;
+  EXPECT_EQ( output.str(), "Bool_5" );
+}
+
+// ----------------------------------------------------------------------------
+TEST(image_memory, copy_constructor)
+{
+  std::string set_data = "test";
```
mleotta, line 132: I suppose this works as a test, but normally image memory is not a null terminated string.  It is common to have zero in the middle or even be all zeros.

vital/tests/test_image.cxx
```diff
@@ -62,6 +90,79 @@ struct val_incr_op
...
+  image_memory mem{ set_data.size()*sizeof( char ) };
+
+  char* data = (char*) mem.data();
+  for(unsigned int i=0; i<set_data.size(); i++)
+  {
+    data[i] = set_data[i];
```
mleotta, line 138: Maybe it's better to just  use `data[i] = static_cast<char>(i);` instead.  Or some other function of `i`.  That way we don't falsely assume the data is a null terminated string.

vital/tests/test_image.cxx
```diff
@@ -62,6 +90,79 @@ struct val_incr_op
...
+  }
+
+  image_memory mem_copy{ mem };
+  check_equal( mem, mem_copy );
+}
+
+// ----------------------------------------------------------------------------
+TEST(image_memory, assignment_operator)
+{
+  std::string set_data = "test";
```
mleotta, line 148: same comment above about using strings as test data.

vital/tests/test_image.cxx
```diff
@@ -297,6 +370,12 @@ TEST(image, equal_content)
     }
   }
   EXPECT_TRUE( equal_content( img1, img2 ) );
+
+  image img3( w, h, d, false, image_pixel_traits{ image_pixel_traits::BOOL } );
+  EXPECT_FALSE( equal_content( img1, img3 ) );
+
+  *img2.first_pixel()=1;
```
mleotta, line 377: better syntax would be `img2.at<bool>(0,0) = 1;`.  `first_pixel()` returns a `void*`.  I'm surprised this works without the compiler complaining.

mleotta (changes requested):
see inline comments

vital/tests/test_image.cxx
OwenMcGee: I believe this is actually calling the `image_of` class's `first_pixel()` function, which returns T* not void*, which is why the compiler is okay with it.

vital/tests/test_image.cxx
OwenMcGee: The most recent commit addresses this, but I needed to make the `data()` function const, which included changing it in the ocv and vxl arrows. Let me know if that wasn't the right way to do things

chetnieter (approved):
Looks good to me. I am going to re-run the dashboards to see if that test failure was a fluke.

chetnieter:
Jenkins test this please

vital/tests/test_image.cxx
OwenMcGee: These changes were determined to be out of the scope of this PR, so for now we're just leaving this as is. (that "most recent commit" I referenced has been removed)

PR #1577, opened 2022-04-08
Update release notes

daniel-riehm:


doc/release-notes/master.txt
```diff
@@ -21,3 +21,56 @@ Build System
...
+* Added the option to choose the metadata export format in dump-klv.
+
+* Added the derive_metadata filter.
+
+* Enabled exporting components of composite data types separately in
+  metadata_map_io_csv.
+
+* Added frame decimation to metadata_map_io_csv.
+
+* Fixed a bug producing invalid CSV files in metadata_map_io_csv.
```
mleotta, line 46: In the past we have a whole separate section at the bottom to document bug fixes and this "Updates" section is only for new features.  See previous release notes for examples.  Also note that bug fixes only need to be listed if they are bugs that present in the previous release.  If it's a bug that was introduced and then fixed since the last release, we don't need to mention it (not saying that's the case here).  Typically we have few bug fixes in these major/minor releases because most bugs get pushed to the release branch and end up in a patch release.

mleotta (changes requested):
Looks good, but could you separate "updates" and "fixes" into two top level sections?

mleotta (approved):

PR #1580, opened 2022-04-11
Add api and vxl implementation for KDTree search tree

chetnieter:
Added a simple api for the KDTree search algorithm along with a basic implementation in the vxl arrow. This can be expanded on by exposing some options in the vxl implementation and potentially add implementations from other arrows such as OpenCV and VTK.


vital/algo/kd_tree.h
```diff
@@ -0,0 +1,85 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/// \file
+/// \brief kd_tree algorithm definition
+
+#ifndef VITAL_ALGO_KD_TREE_H_
```
mleotta, line 8: I suggest calling the abstract algorithm `nearest_neighbors`.  The implementation of this class using could still be called `kd_tree` and may be use the name "vxl_kd_tree".  KD-tree is a specific algorithm for finding nearest neighbors.  There are others including brute force search and binning.  This way we can bundle other nearest neighbors algorithms under this same abstract algorithm.  At some point we may also want to extend this to dimensions other than 3, but that can wait.

mleotta (approved):

mleotta (changes requested):
Looks good, but can you add a release not to master.txt about this.

chetnieter:
@mleotta I added release notes so unless you have any other concerns I am going to go ahead and merge.

PR #1587, opened 2022-04-18
Refactor ST0104 timestamp conversion

daniel-riehm:
This PR moves the function which converts a ST0104 timestamp string to a UNIX timestamp from vital to the KLV arrow, where it should be. It also newly allows for omission of the `T` in `YYYYMMDD[T]hhmmss`, which is not technically correct, but has been seen in real-world data and is harmless to add.


mleotta (approved):

mleotta (changes requested):
We should get in the habit of maintaining release notes.  Can you can a note to master.txt about this change.  Otherwise looks good.

daniel-riehm:
> We should get in the habit of maintaining release notes. Can you can a note to master.txt about this change.

I would, but looks like `master.txt` currently summarizes all of the KLV overhaul work, since 1.6.0 was 12 months ago. This is a change to code that hasn't appeared in a release yet, so I'm not sure what change in the release notes would be appropriate.

mleotta (approved):

PR #1591, opened 2022-04-21
Fixes for essential_matrix.cxx

willdunklin:
Contains a critical change to the conversion copy constructor ([here](https://github.com/Kitware/kwiver/blob/master/vital/types/essential_matrix.h#L97)) which fixes it accessing protected variables outside of scope.

Removed unused and unrunnable conversion copy constructors as they are generalized by the constructor above.

Changed `essential_matrix_<T>::clone()` to use `std::make_shared` when creating the shared pointer.

### DIFF
--- a/vital/types/essential_matrix.cxx
+++ b/vital/types/essential_matrix.cxx
@@ -102,7 +102,7 @@ essential_matrix_sptr
 essential_matrix_<T>
 ::clone() const
 {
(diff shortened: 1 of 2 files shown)


chetnieter:
This looks fine me. I don't understand why those templated specialized constructors are there. I am going to defer to @mleotta since I believe this is his code.

chetnieter (approved):

PR #1593, opened 2022-04-21
Use base64 to encode bytestrings

daniel-riehm:
Since `cereal`'s JSON interface doesn't give us much control over formatting, sequences of bytes are currently printed one byte per line (see deleted lines in `klv_gold.json` for examples). This wastes a lot of space unnecessarily, which can make these JSON files quite unwieldy for longer videos; also much harder for a human to navigate. This PR addresses this problem by encoding sequences of bytes as base64 strings instead.

### DIFF
--- a/arrows/serialize/json/klv/load_save_klv.cxx
+++ b/arrows/serialize/json/klv/load_save_klv.cxx
@@ -35,6 +35,7 @@
 
 #include <vital/internal/cereal/archives/json.hpp>
 #include <vital/internal/cereal/cereal.hpp>
+#include <vital/internal/cereal/external/base64.hpp>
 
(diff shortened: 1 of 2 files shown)


arrows/serialize/json/klv/load_save_klv.cxx
```diff
@@ -573,7 +574,7 @@ class klv_json_saver : public klv_json_base< save_archive >
 
   void save( klv_blob const& value )
```
hdefazio, line 575: maybe call this save_base64 or something? "save" in combination with the blob input type seems non-descriptive.

hdefazio:
LGTM

PR #1594, opened 2022-04-21
Move packet timestamp logic inside klv_demuxer

daniel-riehm:
This PR addresses a bug in which, under certain circumstances, KLV packets with no extractable timestamp will be associated with the wrong frame. Solving this problem was more complicated than one might think, and involved giving up on `klv_demuxer.seek()` (which wasn't actually used anywhere, anyway). A unit test is included to verify resolution of the bug.


daniel-riehm:
(Linux failure appears to be unrelated)

PR #1595, opened 2022-04-22
Add init_list constructor to klv_blob

daniel-riehm:
Add a new constructor to klv_blob which prevents the need for double braces when constructing from literal bytes.


daniel-riehm:
(Linux failure unrelated)

hdefazio:
LGTM

PR #1598, opened 2022-04-22
Add misp_timestamp_now

daniel-riehm:
Add a function giving the current time according to the MISP timesystem. Necessary for encoding a new ST1108 packet. This is the best way to do it that I know of given C++11, but the result may be inaccurate until we can use `std::chrono::tai_clock` (C++20).


daniel-riehm:
(Linux failure unrelated)

hdefazio:
LGTM

PR #1603, opened 2022-04-27
Make smoothing KLV packets when reading optional

daniel-riehm:
Currently, we normalize, or "smooth" KLV packets when reading them in from a video. This is nice for a lot of things, but not for just dumping the KLV packets as they exist in the video (the expected behavior for `dump-klv`). This PR makes reading the "raw" KLV packets the default, and "smoothing" them a configurable option.


hdefazio:
Assuming Mac failure is unrelated, LGTM

PR #1607, opened 2022-05-04
Add support for multiple KLV streams

daniel-riehm:
Finally add support for reading multiple KLV streams as separate streams. Previously, we had been piping all KLV packets from all streams into a single processing stream, which caused all sorts of problems with interpretation. Usually the streams are separate because they come from different parts of the UAV system: two different sensors, or one sensor and one central processing module. These provide different and sometime contradicting data, and update out of sync with each other, causing various problems later in the pipeline.

Each stream is now processed separately in its own `vital::metadata` object, marked with its respective stream index, and exported in the `metadata_vector` the FFmpeg reader provides. The CSV and KLV-JSON exporters have been updated accordingly.

Additional reviewers: @hdefazio


hdefazio:
LGTM

PR #1608, opened 2022-05-04
Add unimplemented keys

daniel-riehm:
Add the UDS key, name, and description for a few KLV local sets we haven't implemented. Not strictly necessary for anything, but a few of the keys do appear in some of our test files, and it might be nice to know what data is in there, even if we can't interpret it right now.


hdefazio:
LGTM

PR #1613, opened 2022-05-06
Add raw video/metadata transfer structure

daniel-riehm:
This PR adds the structure in `vital` (and the `transcode` applet) to allow for direct copying of video and metadata from `video_input` to `video_output`. This is intended to solve the problem of losing video fidelity every time the video is transcoded, even if the only thing edited is the metadata.

The new `video_raw_image` and `video_raw_metadata` classes work much the same way as `video_settings`: an empty class meant to be extended by the specific implementation.

(Note that the target problem hasn't _actually_ been solved here: the FFmpeg-specific implementations of these new methods will come in future PRs.)

### DIFF
--- a/arrows/core/applets/transcode.cxx
+++ b/arrows/core/applets/transcode.cxx
@@ -20,6 +20,53 @@ namespace arrows {
 
 namespace core {
 
+namespace {
+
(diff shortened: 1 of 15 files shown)


hdefazio:
LGTM

PR #1615, opened 2022-05-10
Implement direct video stream copying

daniel-riehm:
Implement direct video-stream copying for our FFmpeg video input/output. Metadata stream copying not fully implemented here. The unit test verifies that no video fidelity is lost in the copy, and using the `--copy-video` option in the `transcode` applet provides ~3x speedup on my machine over re-encoding the video. 


hdefazio:
LGTM

PR #1617, opened 2022-05-16
Adding API and PDAL implementation to load point cloud data

chetnieter:


chetnieter:
This MR also includes some formatting fixes from uncrustify.

willdunklin (changes requested):

PR #1618, opened 2022-05-20
Refactor ST1010 sigma encoding

daniel-riehm:
In preparation for implementing the ST1107 standard, refactor the ST1010 standard to allow its standard deviation members to have separate IMAP parameters. 


hdefazio:
LGTM

PR #1619, opened 2022-05-20
Add basic ffmpeg output config params

daniel-riehm:
Add a few basic configuration options to `ffmpeg_video_output`, allowing opening files (by specifying the required width/height/frame rate) and some other knob-turning without having to pass in a `ffmpeg_video_settings`. An example of a successful `open()` is given in the short unit test added here.


hdefazio:
LGTM

PR #1621, opened 2022-05-24
Refactor KLV checksums

daniel-riehm:
Make the checksums their own `klv_data_format`s, which avoids the duplication of checksum reading/writing logic across all standards that have checksums.


hdefazio:
LGTM

PR #1622, opened 2022-05-25
Avoid dereferencing an end iterator in track_features_core

predicative:
This PR fixes undefined behavior in the `core` implementation of `track_features`. In the event that no tracks were created on the first frame, the attempt to dereference `prev_tracks->all_track_ids().crbegin()` is undefined behavior (and in practice seems to lead to a crash). Now `next_track_id` will be left as 0 if `prev_tracks` is empty.

### DIFF
--- a/arrows/core/track_features_core.cxx
+++ b/arrows/core/track_features_core.cxx
@@ -370,8 +370,11 @@ track_features_core
     return new_track_set;
   }
 
-  // get the last track id in the existing set of tracks and increment it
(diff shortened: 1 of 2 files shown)


mleotta (approved):

mleotta (changes requested):
Looks good but needs a release note in `release.txt` to document the issue that was fixed.

predicative:
Thanks for taking a look. I've created `release.txt` and added a note for this issue.

PR #1624, opened 2022-05-27
Make kwiversys regex compile() thread-safe

daniel-riehm:
KwiverSys' regex compiler uses global variables to store state, meaning if we compile multiple regexes simultaneously in different threads, there is a race condition which can cause crashes. This came up when trying to load two `.conf` files simultaneously in BurnOut.

This PR adds a mutex to prevent these race conditions, which is really a band-aid until we can replace KwiverSys regex with `std::regex`. Technically we should be able to do that _now_, since `<regex>` is in C++11, but the `gcc` version in our CentOS CI machine doesn't actually fully support C++11, so we're stuck with this for the time being.

### DIFF
--- a/vital/kwiversys/RegularExpression.cxx
+++ b/vital/kwiversys/RegularExpression.cxx
@@ -37,6 +37,8 @@
 # include "RegularExpression.hxx.in"
 #endif
 
+#include <mutex>
+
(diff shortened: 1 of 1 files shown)


hdefazio:
LGTM

PR #1627, opened 2022-05-31
Fix race condition in close_loops_exhaustive

predicative:
I discovered when using `close_loops_exhaustive` that it would sometimes crash, and debugging revealed a race condition, namely that the call to `merge_tracks` can ultimately call `erase` on `simple_track_set_implementation::data_`, which can cause the call to `simple_track_set_implementation::tracks`, which copies `data_` and is indirectly made by `match_tracks`, to copy `data_` while elements are actively being removed from it. This manifests as empty `shared_ptrs` in the return value of tracks that are dereferenced in `track_set_implementation::active_tracks` to cause a segfault.

This PR removes the race by finishing computing the matches before modifying the feature tracks. This is straightforward, but perhaps something more involved would be more performant.

With some digging I found 17c7936e6b3514fc9ce677f74eb2b283cfe36ebf, which at a glance may have created the issue or made it worse. It also suggests that `close_loops_keyframe` may have a similar issue.

### DIFF
--- a/arrows/core/close_loops_exhaustive.cxx
+++ b/arrows/core/close_loops_exhaustive.cxx
@@ -152,10 +152,17 @@ close_loops_exhaustive
     all_matches[f] = pool.enqueue(match_func, f);
   }
 
+  std::map<vital::frame_id_t, track_pairs_t> all_matches_done;
+  // ensure all parallel jobs are finished before modifying input
+  for(auto& pair : all_matches)
+  {
+    all_matches_done[pair.first] = std::move(pair.second.get());
+  }
+
   // retrieve match results and stitch frames together
   for(vital::frame_id_t f = frame_number - 2; f >= last_frame; f-- )
   {
-    auto const& matches = all_matches[f].get();
+    auto const& matches = all_matches_done[f];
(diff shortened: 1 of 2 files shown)


arrows/core/close_loops_exhaustive.cxx
```diff
@@ -152,10 +152,16 @@ ::stitch( vital::frame_id_t frame_number,
     all_matches[f] = pool.enqueue(match_func, f);
   }
 
+  std::map<vital::frame_id_t, track_pairs_t> all_matches_done;
```
mleotta, line 155: You might want to add a comment to document that this loop and transferring to a new map is for thread synchronization.  E.g. make sure all the parallel jobs are finished before merging the tracks.

mleotta (changes requested):
looks like a reasonable fix, but we should apply this to the release branch and add a note in the release notes doc.

predicative:
Thanks for the review, @mleotta. I've rebased the PR onto `release` and added a comment and release notes. Please take a look.

mleotta (approved):

PR #1630, opened 2022-06-03
Porting tests from initialize_cameras_landmarks_basic to test initialize_cameras_landmarks

OwenMcGee:
Currently doesn't compile due to issues in the `evaluate_initializer()` helper function. With the lines calling that function commented out, all tests currently run and succeed aside from `subset_init`, which I haven't yet altered. 

The helper function `helper()` (to be renamed) condenses repeat code from the original `initialize_cameras_landmarks_basic` tests. I also adjusted the `configure_algo()` helper function to set additional config values that weren't needed in the original basic tests.


OwenMcGee:
Tests currently fail in the `evaluate_initialization` helper function, as the `est_cams` variable is a `nullptr`. The variable _should_ be set in the `initialize` function in `initialize_cameras_landmarks.cxx`, but isn't due to functions returning too early (these checks should probably throw exceptions, rather than just returning early)

arrows/mvg/algo/initialize_cameras_landmarks.cxx
```diff
@@ -2437,7 +2437,8 @@ ::metadata_centric_keyframe_initialization(
       vector_3d pos_loc;
       rotation_d R_loc;
 
-      if (constraints->get_camera_position_prior_local(fid, pos_loc) &&
+      if (constraints == nullptr &&
```
OwenMcGee, line 2440: This is necessary because the `initialize()` function sets its fourth parameter `constraints` to have a default value of `nullptr`, which as far as I can tell will _always_ result in a segmentation fault in this function . This addition (checking for `nullptr`) stops the segmentation fault, but I believe might cause issues later on with variables not being defined when they should be.

Unless I'm mistaken, nowhere currently calls the `initialize()` function _using_ the default value for `constraints` (doing so would cause a seg fault, I believe), so it would seem the better option here would be to simply change the function to not have a default value for that parameter. This would also necessitate passing in real values for it in the test file

OwenMcGee: Documented this issue in #1639

mleotta, line 2440: Yes, this is the right thing to do.  Though you could more compactly write `if (constraints &&` since the value of a nullptr is zero which casts to `false` in Boolean.

sedrakSofthenge: Better to move this check right after cam->clear(); call before "for" loop, since if constraints is nullptr, the loop is useless. Something like this
    `
    cams->clear();
    if (!constraints)
    {
      return false;
    }
    `
it can be returned immediately because anway after it goes out of if-else block, `cams->size() < 2` condition is being checked, which means it will always return false there, so we can leave the function immediately. Otherwise `for` loop needs to be in block `if (constraints) { for ...`

PR #1634, opened 2022-06-21
Refactor ffmpeg_video_output

daniel-riehm:
Refactor the FFmpeg video writer in the same style as the recent FFmpeg video reader refactoring: additional comments and better error/exception handling.

@hdefazio

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_output.h
+++ b/arrows/ffmpeg/ffmpeg_video_output.h
@@ -3,17 +3,17 @@
 // https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
 
 /// \file
-/// \brief Declaration of FFmpeg video writer.
+/// Declaration of FFmpeg video writer.
 
 #ifndef KWIVER_ARROWS_FFMPEG_FFMPEG_VIDEO_OUTPUT_H
-# define KWIVER_ARROWS_FFMPEG_FFMPEG_VIDEO_OUTPUT_H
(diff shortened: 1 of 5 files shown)


arrows/ffmpeg/ffmpeg_video_output.h
```diff
@@ -51,11 +51,11 @@ class KWIVER_ALGO_FFMPEG_EXPORT ffmpeg_video_output
...
 
   void add_image(
-    kwiver::vital::image_container_sptr const& image,
-    kwiver::vital::timestamp const& ts ) override;
+    vital::image_container_sptr const& image,
+    vital::timestamp const& ts ) override;
   void add_image( vital::video_raw_image const& image );
 
-  void add_metadata( kwiver::vital::metadata const& md ) override;
+  void add_metadata( vital::metadata const& md ) override;
```
hdefazio, line 58: Not saying this is wrong, but I am noticing inconsistent use of kwiver::vital vs vital vs ::kwiver::vital throughout kwiver. Not sure which one is the best.

mleotta: There are legitimate use cases for `::kwiver::` instead of `kwiver::` but I don't think we have that case here.  If I created nested namespaces `blah::kwiver::` and my scope is currently in the `blah` namespace then `kwiver::` becomes ambiguous and you need to use `::kwiver::` to indicate the kwiver namespace in the global namespace.

I would not use `::kwiver::` unless there is a case where it is needed.

PR #1635, opened 2022-06-24
Enable configuration of KLV update rates

daniel-riehm:
When converting a `klv_timeline` back to `klv_packet`s, the current behavior is to just dump out every value every frame. While this correctly expresses the data, it is inefficient with regards to values that do not often change, such as the name of the platform or the security classification of the footage.

This PR adds the capability to specify how often unchanging values should be repeated, at multiple levels of granularity (all standards / single standard / specific tag). This can result in a manyfold smaller KLV stream over the course of a long video. A few tests are provided.

The `klv_recommended_update_intervals()` function will be implemented in a future PR.

@hdefazio


hdefazio:
LGTM

PR #1643, opened 2022-07-11
Limit extreme gaussian outliers

daniel-riehm:
Recent upgrades of the KWIVER Windows CI machine from Visual Studio 2017 to Visual Studio 2019 had the unfortunate effect of changing the random number generation scheme in such a way that two of our Ceres automated tests will always fail. These tests add pseudorandom Gaussian noise to ground-truth points to simulate real-world data. However, the new MSVC random normal distribution seeded with the default seed happens to create a very extreme outlier within the first couple dozen numbers generated, which causes the Ceres bundle adjuster to fall outside of "acceptable" error limits. This PR limits the normal distribution used by our tests to three standard deviations on each side, redistributing the 0.3% remaining probability mass back inside those limits and keeping things sane.

In the spirit of "fix as you go", this PR also reworks the small file it changes to be in accordance with the style guide.

### DIFF
--- a/tests/test_random_point.h
+++ b/tests/test_random_point.h
@@ -2,52 +2,69 @@
 // OSI-approved BSD 3-Clause License. See top-level LICENSE file or
 // https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
 
-/**
- * \file
- *
- * \brief Functions for creating test points with added random Gaussian noise.
- *
- */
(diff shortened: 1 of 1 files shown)


daniel-riehm:
Looks like a lot of our tests rely on the exact random points produced by the default random seed. Closing because this breaks more things than it fixes.

PR #1645, opened 2022-07-21
Pass constant seeds to random point generator

daniel-riehm:
This PR replaces the `static` rng instance feeding the `random_point_Xd()` functions with a function parameter. Previously, the `static` nature of the rng means that running one test actually affected the behavior of the other tests in the file, since it changed the random seed. This PR ensures consistent behavior no matter if you are running all of a test file's tests (typical user behavior) or only one at a time (how our CI does it). Since changing the random seed changes the points fed into tests, one or two of them need their tolerances expanded slightly.

Hopefully this will fix the test failures with Visual Studio 2019. If not, different seeds or tolerances may need to be tried.

@hdefazio


hdefazio:
LGTM

PR #1647, opened 2022-07-22
Eliminate templating on iterators

daniel-riehm:
Early on in the construction of the KLV system, I didn't know exactly how raw bytes would be fed to the reader/writer, so I wrote code which could work with any iterator type. While this has some advantages, it also complicates things. This PR removes that now-unnecessary complication, using simple pointers to `uint8_t` everywhere instead.

@hdefazio


daniel-riehm:
(Windows failure unrelated)

hdefazio:
LGTM

PR #1648, opened 2022-07-25
Make control command timestamp kv::optional

daniel-riehm:
Fix a small inconsistency in the control command data structure which treats at timestamp of `0` as "timestamp not present", where it would be better to make the timestamp field explicitly `optional`.

@hdefazio

### DIFF
--- a/arrows/klv/klv_0601.cxx
+++ b/arrows/klv/klv_0601.cxx
@@ -1283,9 +1283,7 @@ operator<<( std::ostream& os, klv_0601_control_command const& value )
 {
   return os << "{ " << "ID: " << value.id
             << ", String: \"" << value.string
-            << "\", Timestamp: " << ( value.timestamp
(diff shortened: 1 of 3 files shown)


arrows/klv/klv_0601.cxx
```diff
@@ -1283,9 +1283,7 @@ operator<<( std::ostream& os, klv_0601_control_command const& value )
 {
   return os << "{ " << "ID: " << value.id
             << ", String: \"" << value.string
-            << "\", Timestamp: " << ( value.timestamp
-                                  ? std::to_string( value.timestamp )
-                                  : "(empty)" )
+            << "\", Timestamp: " << value.timestamp
```
hdefazio, line 1286: I think I remember that the optional type takes care of printing the empty value. In that case, there are some 0601 prints that should be updated. (LGTM)

PR #1649, opened 2022-07-27
Ensure we don't segfault when KLV parsing fails

daniel-riehm:
This PR fixes a bug in the "backup system" when KLV parsing fails, where the read iterator didn't get reset properly.

@hdefazio


daniel-riehm:
jenkins test this please

hdefazio:
LGTM

PR #1650, opened 2022-07-28
Add missing error checking for calls to klv_value::get()

daniel-riehm:
This PR adds error handling in a few places where the possibility of KLV parsing having failed earlier in the pipeline is not properly accounted for.

@hdefazio


hdefazio:
Assuming mac failure is unrelated... LGTM

PR #1652, opened 2022-08-02
Fix FFmpeg duration bug

daniel-riehm:
This PR fixes a bug in which the duration is calculated from a start and an end time, but those times are calculated relative to different reference points. This was causing `num_frames()` to return wildly incorrect values. Unsure exactly how this got past regression tests - maybe our test video, being artificial and not cut from a longer video, has a PTS which starts at 0 for the first frame, which could have hidden the bug? In any case, the problem should be fixed now.

@hdefazio


daniel-riehm:
(Mac failure unrelated)

hdefazio:
LGTM

PR #1654, opened 2022-08-04
Use interval to express IMAP bounds

daniel-riehm:
The IMAP and flint KLV types map a range of floating point numbers to integer values. The functions to do this were written before the `vital::interval<>` template was created, so they use separate `minimum` and `maximum` parameters.This PR replaces those with `interval`s, which is more consistent with the rest of the KLV code and allows us to use `interval`'s built-in utility methods. Since `interval` will flip the minimum and maximum if they are given in the wrong order, that case is removed from the KLV unit tests.

@hdefazio


hdefazio:
Is there a reason why some save() functions are being deleted in arrows/serialize/json/klv/load_save_klv.cxx ?

daniel-riehm:
> Is there a reason why some save() functions are being deleted in arrows/serialize/json/klv/load_save_klv.cxx ?

Yes: getting this change to compile would have required updating them, but actually they are unused code, so it made sense to me to just delete them instead. They used to be a part of serializing MISB 1010 SDCC-FLP structures, recording which of the two possible data formats (float vs IMAP) was used. However, in some past PR those `klv_data_format`  fields were switched to a simple `bool` (since there are only two possible formats), meaning there's no need to serialize the formats themselves anymore.

hdefazio:
okay, sounds good

LGTM

PR #1655, opened 2022-08-05
Refactor MISP timestamps to use std::chrono

daniel-riehm:
H.265 MISP frame timestamps can be nanosecond resolution. Right now, we just round these to the nearest microsecond. This PR changes that, storing the full nanosecond value.

@hdefazio


hdefazio:
LGTM

PR #1656, opened 2022-08-10
Revert erroneous video PTS changes

daniel-riehm:
The recent addition of CUDA-accelerated encoding capabilities included an erroneous change which broke the frame timings on normal videos. This PR reverts that change and adds the unit test which should have caught the break in the first place.

@hdefazio

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_output.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_output.cxx
@@ -54,6 +54,8 @@ public:
     bool write_next_packet();
     void write_remaining_packets();
 
+    int64_t next_video_pts() const;
+
     impl* parent;
 
     size_t frame_count;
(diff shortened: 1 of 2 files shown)


arrows/ffmpeg/ffmpeg_video_output.cxx
```diff
@@ -776,6 +778,17 @@ ::write_remaining_packets()
...
 
+// ----------------------------------------------------------------------------
+int64_t
+ffmpeg_video_output::impl::open_video_state
+::next_video_pts() const
+{
+  return static_cast< int64_t >(
+    static_cast< double >( frame_count ) *
+    video_stream->time_base.den / video_stream->time_base.num /
+    codec_context->framerate.num * codec_context->framerate.den + 0.5 );
```
hdefazio, line 789: Not necessary, but it may be worth breaking this up or adding parentheses for readability

PR #1657, opened 2022-08-10
Rename frame center HAE tag

daniel-riehm:
This PR just renames a ST0601 tag to be more consistent with all the other ST0601 tags, by use the shorter phrase `ELLIPSOID_HEIGHT` rather than `HEIGHT_ABOVE_ELLIPSOID`.

@hdefazio


hdefazio:
LGTM

PR #1658, opened 2022-08-10
Fix bug when writing last element of view domain

daniel-riehm:
This PR fixes a bug which would not write the length of the final element of the ST0601 view domain field (when present).

@hdefazio


hdefazio:
LGTM

PR #1659, opened 2022-08-15
Add missing std::set< uint16_t > type to KLV JSON exporter

daniel-riehm:
The JSON exporter was not updated when the type of the ST0601 Active Payloads tag was changed. This PR fixes that bug.

@hdefazio


hdefazio:
LGTM

PR #1660, opened 2022-08-15
Tweak FFmpeg output settings assignment

daniel-riehm:
This PR fixes some issues with settings from different sources (`ffmpeg_video_settings`, config values, ffmpeg-suggested values) overriding each other incorrectly. It also ensures that the video stream's codec type is always set to `VIDEO`, which is necessary for `avcodec_parameters_to_context()` to function correctly.

@hdefazio


hdefazio:
LGTM

PR #1661, opened 2022-08-18
Tolerate packets with invalid data

daniel-riehm:
We already carry on when FFmpeg fails in decoding a video frame due to invalid or corrupted data, proceeding to the next frame instead of giving up. However, evidently FFmpeg can sometimes detect the invalidity of data immediately when a packet is given to it, as opposed to later during the decoding process. This PR ensures that such early detection does not crash the program.

@hdefazio


hdefazio:
LGTM

PR #1662, opened 2022-08-18
More robustly handle invalid or multiple video streams

daniel-riehm:
This PR adds checks and log messages dealing with cases where there are multiple video streams in the same file, or where the video stream is sufficiently corrupted as to be unreadable.

@hdefazio


hdefazio:
LGTM

PR #1663, opened 2022-08-22
Add configuration options to disable imagery or KLV processing

daniel-riehm:
This PR adds two configuration options which direct the FFmpeg video reader to avoid processing either the video imagery, or the KLV streams. This would be useful, for example, in the case of `dump-klv`, where processing the image data is often a waste of CPU cycles. Or in the case where one of the streams is so corrupted as to be unreadable (as we encountered recently), but we still want to recover the other stream's data.

Two tests are included which activate these options and check that they generally behave as expected. Many of the other changes consist only of putting some existing block of code in an `if` statement and leaving it otherwise untouched.

@hdefazio


hdefazio:
LGTM

PR #1664, opened 2022-08-24
Add nanosecond timestamp metadata item

daniel-riehm:
This PR adds a nanosecond version of the vital timestamp metadata item, because it is possible for H.265 frame timestamps to be expressed in nanosecond resolution. Not intended to replace the microsecond version in regular use, only to prevent unnecessary rounding and facilitate more faithful transcoding in this particular case.

@hdefazio


hdefazio:
LGTM

PR #1665, opened 2022-08-25
Create wrapper around avformat_query_codec()

daniel-riehm:
Apparently FFmpeg's "does this file type support this codec" function, `av_query_codec()`, is not sure that H.264 and H.265 video streams can be put in a Transport Stream, even though they can. This PR wraps `av_query_codec()` in a utility function where we can add exceptions for cases like this.

@hdefazio


mleotta:
the change looks okay. Is this a bug in certain versions of FFMPEG?

mleotta (approved):

daniel-riehm:
> the change looks okay. Is this a bug in certain versions of FFMPEG?

I'm not sure it counts as a bug inasmuch as an unimplemented feature, since the function returns "information not known" rather than "definitely not supported", so it's not quite 100% _wrong_. I tested this in 4.4.1; don't know if it's fixed in v5.

PR #1666, opened 2022-09-07
Fix UV coords going to NaN for 0 area triangles

willdunklin:


chetnieter (approved):
Changes look fine and it resolves the issues I was having with Danesfield. I see a seg fault at the end of my script but I confirmed this occurs for current master as well. Mac failure appears to due qtExtensions missing.

PR #1667, opened 2022-09-08
Add KWIVER_ENABLE_FFMPEG_CUDA option

daniel-riehm:
Add a CMake option to disable compilation of the FFmpeg CUDA functionality, but still allow the use of CUDA in other arrows.

@hdefazio


willdunklin (approved):
The desired functionality works and the implementation looks good to me

PR #1668, opened 2022-09-08
Fix cereal loading bug

daniel-riehm:
This PR fixes a bug which causes JSON parsing to fail when a member of a local set is completely empty / null, and it is not the last member of that local set. The unit test now covers this edge case.

@hdefazio

### DIFF
--- a/arrows/serialize/json/klv/load_save_klv.cxx
+++ b/arrows/serialize/json/klv/load_save_klv.cxx
@@ -287,7 +287,14 @@ protected:
   set_next_name( std::string const& name )
   {
     m_next_name = name;
-    m_archive.setNextName( m_next_name.c_str() );
+    if( name.empty() )
+    {
+      m_archive.setNextName( nullptr );
+    }
+    else
+    {
+      m_archive.setNextName( m_next_name.c_str() );
+    }
   }
 
 private:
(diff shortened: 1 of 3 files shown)


arrows/serialize/json/klv/load_save_klv.cxx
```diff
@@ -287,7 +287,14 @@ class klv_json_base
   set_next_name( std::string const& name )
   {
     m_next_name = name;
-    m_archive.setNextName( m_next_name.c_str() );
+    if( name.empty() )
```
hdefazio, line 290: Should this be `m_next_name`? Or if not, should `m_next_name = name;` be moved to the `else{}` ?

daniel-riehm: Since they have the same value due to the previous line, using either `m_next_name` or `name` should be equivalent here. But I can change it if it would reduce confusion.

hdefazio: Yeah, I meant that the assignment `m_next_name = name` is kinda pointless before the if statement if it is only being used in the `else` case. So, I'd either use it in the condition or just move the assignment.

daniel-riehm: But `m_next_name` is a member variable, so assigning to it has effects outside of this method, for example in the method preceding this one.

PR #1669, opened 2022-09-09
Fix bug when seeking

daniel-riehm:
Apparently the `AVSEEK_FLAG_FRAME` flag is not supported in all cases, resulting in FFmpeg sometimes interpreting the number as a (very low) native timestamp and seeking to the beginning of the video no matter what frame is given. This necessitates flipping through each frame of the video until the desired frame is found, an unnecessarily lengthy process. This PR converts the frame number into a native timestamp instead, removing the reliance on `AVSEEK_FLAG_FRAME` and fixing the problem. 

@hdefazio


hdefazio:
LGTM

PR #1671, opened 2022-09-12
Fix integer overflow bug

daniel-riehm:
This PR fixes a bug in which the conversion from seconds to microseconds happened before the conversion from integer to double, causing integer overflow when passing ~35 minutes of video, as FFmpeg uses 32-bit integers internally in most cases.

@hdefazio

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_input.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_input.cxx
@@ -1407,8 +1407,7 @@ ffmpeg_video_input::priv::open_video_state
     return {};
   }
   return kv::timestamp{
-    static_cast< kv::time_usec_t >(
-      av_q2d( av_mul_q( curr_time(), av_make_q( 1000000, 1 ) ) ) + 0.5 ),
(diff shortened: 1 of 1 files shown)


mleotta (approved):
Looks like a reasonable fix.  I assume that `kv::time_usec_t`, which is an currently an alias for `std::time_t`, is defined as a 64-bit integer, otherwise this would still be an issue.  I wonder if we should explicitly define `kv::time_usec_t` as `int64_t` (which is probably what it is, but is not guaranteed by the standard). However, that change is outside the scope of this fix and may introduce other issues.

daniel-riehm:
> Looks like a reasonable fix. I assume that `kv::time_usec_t`, which is an currently an alias for `std::time_t`, is defined as a 64-bit integer, otherwise this would still be an issue. I wonder if we should explicitly define `kv::time_usec_t` as `int64_t` (which is probably what it is, but is not guaranteed by the standard). However, that change is outside the scope of this fix and may introduce other issues.

`time_usec_t` is defined as `int64_t` on line 38 of `vital/vital_types.h`, so we're all good there.

mleotta:
Ah, good.  I was looking at the wrong thing and thought I saw that `time_usec_t` was an alias for `std::time_t`, but I had that wrong.

PR #1672, opened 2022-09-13
Add missing std::set< uint16_t > type to KLV JSON exporter

daniel-riehm:
This PR adds a missing type to the KLV JSON exporter, and a test case to the associated unit test.

@hdefazio


hdefazio:
LGTM

PR #1673, opened 2022-09-13
Take start_time into account when seeking

daniel-riehm:
One more bug fix for seeking when the starting pts of a video is not zero.

@hdefazio


daniel-riehm:
Closed in favor of #1675

PR #1675, opened 2022-09-14
Fix further FFmpeg timestamp bugs

daniel-riehm:
This PR fixes integer overflow problems, a CUDA-only oversight which would drop the frame timestamp, and a bug with the seek timestamp calculation sometimes causing the video reader to seek much earlier in the video than desired.

@hdefazio


hdefazio:
LGTM

PR #1676, opened 2022-09-20
Pad image memory to avoid segfault

daniel-riehm:
This PR adds a few bytes of padding onto allocated video image memory, since the image format conversion algorithms packaged with FFmpeg can use vectorized algorithms which potentially write a few bytes past the end of the logical image.

@hdefazio


hdefazio:
LGTM

PR #1678, opened 2022-09-30
Add VITAL_META_UNIX_TIMESTAMP_SOURCE

daniel-riehm:
This PR adds a new vital metadata item indicating how the frame timestamp was derived - from a MISP frame timestamp, or from a KLV packet. We also clean up a small verbosity in timestamp logic, in which the line to add the timestamp is included in each standard-specific function, instead of just happening once before calling all of them.

@hdefazio


hdefazio:
LGTM

PR #1680, opened 2022-10-03
Fix KLV packet length bug

daniel-riehm:
Fixes a bug in which, when calculating the length of the length field of a packet, the length (usually 4 bytes) of the checksum at the end of the packet is omitted. This bug would only show up if adding those 4 ignored bytes would bump the length field over the threshold requiring it to be longer (127 -> 128). In other words, only packets with a payload between 124 and 127 bytes long would cause this error; hence it only being caught now.

@hdefazio


arrows/klv/klv_packet.cxx
```diff
@@ -202,9 +202,10 @@ klv_packet_length( klv_packet const& packet )
   auto const checksum_format = format.checksum_format();
   auto const length_of_key = packet.key.length;
   auto const length_of_value = format.length_of( packet.value );
-  auto const length_of_length = klv_ber_length( length_of_value );
   auto const length_of_checksum =
     checksum_format ? checksum_format->fixed_length() : 0;
+  auto const length_of_length =
+    klv_ber_length( length_of_value + length_of_checksum );
   return length_of_key + length_of_length + length_of_value +
          length_of_checksum;
```
hdefazio, line 210: Does this need to be removed since `length_of_checksum` is accounted for in `length_of_length` now?

daniel-riehm: Good question, but no. The packet is `[key|length|value|checksum]`. The total packet length is the sum of the lengths of those pieces. The `length` piece is a number equal to the length of `[value|checksum]` (not just `value` - that was the bug fix), and `length` will sometimes be longer if `[value|checksum]` is longer, since storing a larger number requires more bytes. So the length of `checksum` is used twice: once to calculate the value of `length` and thereby its length, and once directly as part of the entire packet.

PR #1681, opened 2022-10-10
Add klv_length_constraints

daniel-riehm:
There are some data formats (strings, floats, IMAP, flint...) where the data can be of variable length, depending on how much information the user wants to encode, or how precise they want to be. The MISB standards put different constraints on these fields depending on the tag: some tags must always be a certain length, some can be any of a range of lengths (e.g 1-4 bytes), and a few have only a handful of discrete lengths to pick from (e.g. a 4-byte raw IEEE float or an 8-byte double). Without this PR, most of these constraints were on the user to self-enforce: only formats with a single fixed length "knew" about that constraint. This PR adds a `klv_length_constraints` class which can be made to represent a wider variety of constraint scenarios. This replaces the `fixed_length` member of `klv_data_format`.

@hdefazio

### DIFF
--- a/arrows/klv/CMakeLists.txt
+++ b/arrows/klv/CMakeLists.txt
@@ -9,6 +9,7 @@ set( sources
   klv_data_format.cxx
   klv_demuxer.cxx
   klv_lengthy.cxx
+  klv_length_constraints.cxx
   klv_key.cxx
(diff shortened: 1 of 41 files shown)


hdefazio:
LGTM

PR #1684, opened 2022-10-14
Improve FindFFMPEG search paths

daniel-riehm:
FFmpeg installs the `.lib` files in the `bin/` directory (instead of in `lib/`) on Windows, which can sometimes prevent KWIVER from finding them. This PR tells KWIVER to search in the `bin/` directory as well.

According to the [documentation](https://cmake.org/cmake/help/v3.24/command/find_library.html), the "search paths" currently given to CMake in `FindFFMPEG.cmake` without the `HINTS` or `PATHS` keyword are interpreted as alternate names for the file, and not actual search paths. As apparently no one ever needed them to work, I feel comfortable removing them.

@hdefazio

### DIFF
--- a/CMake/FindFFMPEG.cmake
+++ b/CMake/FindFFMPEG.cmake
@@ -8,16 +8,9 @@
 
 set( FFMPEG_FOUND "NO" )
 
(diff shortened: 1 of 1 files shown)


mleotta (approved):
Looks good to me.  I recently hit this issue.

PR #1685, opened 2022-10-18
Reorder image/metadata writing

daniel-riehm:
Metadata is supposed to be provided before image data when writing video. The `transcode` applet currently does this backward, and this PR fixes that.

@hdefazio


hdefazio:
LGTM

PR #1686, opened 2022-10-26
Increase GTest discovery timeout

daniel-riehm:
Sometimes CI (notably Mac but sometimes Windows) errors out, saying that gtest timed out when discovering tests. This causes CI pipeline failures even when nothing is truly wrong. This PR increases the test discovery timeout in hopes of preventing these false negatives.

@hdefazio


daniel-riehm:
Looks like this works - first PR in a while with a passing Mac build

hdefazio:
> Looks like this works - first PR in a while with a passing Mac build

cool, LGTM

PR #1687, opened 2022-10-28
Remove VITAL_META_ICING_DETECTED

daniel-riehm:
This PR removes the generally unused `ICING_DETECTED` tag. This PR is the first of a few which aim at pruning the list of `vital` metadata tags to only those actually useful for vision algorithms.

@hdefazio


hdefazio:
LGTM

PR #1689, opened 2022-11-02
Add ST0601 length constraints

daniel-riehm:
This PR adds length constraint information to applicable tags in ST0601. String tags generally have length limits, and IMAP tags allow for a range of possible lengths depending on desired precision. Suggested lengths are provided for this latter case.

@hdefazio


hdefazio:
LGTM

PR #1692, opened 2022-11-04
Add equality operators for metadata_item

daniel-riehm:
This PR adds `==` and `!=` operators to `vital::metadata_item`. The notable special case here is `double` `NaN` values, which would normally never compare equal to each other, so we add custom logic for them.

@hdefazio

### DIFF
--- a/vital/types/metadata.h
+++ b/vital/types/metadata.h
@@ -114,6 +114,9 @@ public:
     }
   }
 
+  bool operator==( metadata_item const& other ) const;
+  bool operator!=( metadata_item const& other ) const;
+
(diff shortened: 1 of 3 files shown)


vital/types/metadata.h
```diff
@@ -114,6 +114,9 @@ class VITAL_EXPORT metadata_item
     }
   }
 
+  bool operator==( metadata_item const& other ) const;
+  bool operator!=( metadata_item const& other ) const;
```
hdefazio, line 118: Maybe add comment saying this is to support double NaN values?

hdefazio (approved):

PR #1697, opened 2022-11-15
Enable loading null characters using cereal

daniel-riehm:
Our copy-paste of the cereal library does not properly handle loading strings that have a null character (`\0`) embedded in them. This PR fixes that, without trying to fix any of cereal's existing bad style.

@hdefazio


hdefazio:
LGTM

PR #1698, opened 2022-11-15
Increase JSON float precision

daniel-riehm:
This PR increases the JSON decimal limit to the maximum recommended by RapidJSON, the base of cereal. The existing `DBL_DIG + 1` is not correct here, as that would be the correct maximum _digits of precision_ (excludes leading zeros after decimal), not the maximum _decimal places_ (includes leading zeros after decimal). Thus, without this PR, `1e-20` would be rounded down to `0.0`, which is not desired behavior. The unit test is adjusted to cover this edge case of numbers much less than `1`.

@hdefazio


hdefazio:
LGTM

PR #1702, opened 2022-11-21
Remove erroneous child set warnings

daniel-riehm:
This PR ensures that amend and segment sets, which are meant to be addenda to their parent set and not sets of their own, will not trigger warnings about not being a proper full set. I wrote all the functions to fix this a while ago, but apparently never actually changed the call sites.

@hdefazio


hdefazio:
LGTM

PR #1704, opened 2022-11-28
Require C++17

daniel-riehm:
This PR switches KWIVER to the C++17 standard, allowing use of those language features in future development. Those building KWIVER on their local machines will likely need to pull fletch master and upgrade their Boost version to 1.78.0, as the previous supported Boost version is not compatible with C++17.

### DIFF
--- a/CMake/kwiver-configcheck.cmake
+++ b/CMake/kwiver-configcheck.cmake
@@ -53,8 +53,8 @@ endmacro()
 # Set default visibility
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 
-# C++11 is required
-set(CMAKE_CXX_STANDARD 11)
+# C++17 is required
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
 vital_check_required_feature(CPP_AUTO         auto.cxx            "auto type specifier")
--- a/doc/release-notes/master.txt
+++ b/doc/release-notes/master.txt
@@ -8,6 +8,10 @@ over the previous v1.7.0 release.
 Updates
 -------
 
+KWIVER
+
+ * Adopted C++17.
+
 Vital
 
 Vital Algo


mattdawkins:
libjson in fletch doesn't support C++17, yielding:

/home/matt/Dev/viame/build/install/include/json/_internal/Source/JSONNode.h:200:54: error: ISO C++17 does not allow dynamic exception specifications
  200 |     JSONNode JSON_PTR_LIB pop_back(json_index_t pos) json_throws(std::out_of_range);
      |                                                      ^~~~~~~~~~~
/home/matt/Dev/viame/build/install/include/json/_internal/Source/JSONNode.h:201:64: error: ISO C++17 does not allow dynamic exception specifications
  201 |     JSONNode JSON_PTR_LIB pop_back(const json_string & name_t) json_throws(std::out_of_range);
      |                                                                ^~~~~~~~~~~
/home/matt/Dev/viame/build/install/include/json/_internal/Source/JSONNode.h:203:71: error: ISO C++17 does not allow dynamic exception specifications
  203 |     JSONNode JSON_PTR_LIB pop_back_nocase(const json_string & name_t) json_throws(std::out_of_range);
      |                                                                       ^~~~~~~~~~~

etc...

I just removed the dependency from the code in VIAME as opposed to fixing the problem, but maybe the lib can be removed? I don't think it's the most popular C++ json library and is a little old now. Not sure if anyone's using it.

Hit two other unrelated issues so far though resolved (e.g. vidl_ffmpeg broken on master, made minor branch I'll push up), but besides that tested it on Linux, trying a windows build tonight.

mattdawkins:
I got KWIVER building though some downstream elements are failing (VIVIA) due to the boost upgrade. I'm hoping it's a fast upgrade to use either 1.78 or std functions but going to assess later today

mattdawkins:
It was pretty fast getting all downstream programs using C++17 and an upgraded boost, and now I'm at the point where there's no outstanding issues. I'm just going to push a couple fletch branches and modified kwiver: viame/master accordingly (which required updates in its database node).

mattdawkins (approved):

PR #1706, opened 2022-11-29
Recognize MSVC 2022

daniel-riehm:
Add a case for MSVC 2022 in our compiler-recognizing logic.

@hdefazio


mleotta (approved):
I haven't tested it, but it looks good to me.

PR #1708, opened 2022-12-06
Fix VIDL FFmpeg Build Issues

mattdawkins:


daniel-riehm (approved):
I've added a commit to use the more recent `klv_demuxer` interface and avoid putting `vxl` things in a different arrow. While I believe the changes after that commit should "work",  this is a band-aid on top of a band-aid on top of unmaintained code. Hopefully we can deprecate and remove the VIDL reader in favor of the `ffmpeg` arrow at some point.

(Also, when you're done looking at the commit I pushed, the commits should be squashed to get rid of the ghostflow error)

mattdawkins:
Agree on eventually deprecation, I don't use klv parsing here just the reader for the time being (so just need it to build)

PR #1709, opened 2022-12-07
Add update_klv metadata filter

daniel-riehm:
This PR adds a metadata filter which takes vital metadata values and puts them in KLV. Right now it only works on VNIIRS and GSD values, putting them in as one ST1108 packet per frame. There are definitely ways to do this better, but many of the difficulties are the fault of the standard's particular construction. For example, only a handful of the possible H.264 video profiles are given enumeration values in ST1108, and there's no "Other" option. Also, almost all tags in ST1108 are mandatory, so if we don't have the video's bitrate information, we can't technically encode GSD or VNIIRS into KLV. So long-term we would hope the standard improves, but for now this works.

@hdefazio


mattdawkins:
Agree on eventually deprecation, I don't use klv parsing here just the reader for the time being

chetnieter (approved):

PR #1710, opened 2022-12-07
Allow transcode applet to use metadata filters without crashing

daniel-riehm:
Since the metadata filter video input only copies capabilities after the video is opened, this PR moves the capabilities check in the transcode applet to after that point. Additionally, the two raw data copying capabilities are now handled in `video_input_metadata_filter`.

@hdefazio


chetnieter (approved):
Looks fine to me

PR #1719, opened 2023-01-11
Fix MISP timestamp writing byte bug

daniel-riehm:
This PR fixes two oversights in the MISP timestamp writing code: `<< 7` shifts seven bits, not seven bytes as intended, and `std::chrono::nanoseconds` uses an `int64_t` internally, not an `uint64_t` as the MISP timestamp format requires.


daniel-riehm:
jenkins test this please

chetnieter (approved):
Looks fine but I would consider adding a test that would catch the original issue that this PR resolves.

daniel-riehm:
Mac failure unrelated; merging

PR #1705, opened 2022-11-29
Add synthetic KLV test videos

daniel-riehm:
This PR adds five 30-frame test videos with comprehensive ST0601 test data, and FFmpeg input tests to ensure there are no regressions as KLV development continues. Also included is a tweak to the JSON reader to make it report the stream index of the packet when reading, which is necessary for one of the tests.

The data file sizes are interesting here: each video is below 200KB, but the two JSON reference files are 2MB each. Reviewers may opine whether these sizes are a problem; the JSON files could be reduced significantly by removing whitespace and unnecessary field labels, albeit at the cost of readability.

@hdefazio

PR #1682, opened 2022-10-12
bundle adjustment applet

borovik135:
This bundle adjustment applet has been reviewed and approved for public release by DARPA as DISTAR case 36967. It appears to build and run based on current master. Now it needs a review before merging into the master branch.

PR #1720, opened 2023-01-11
Put MVG bundle adjust applet destructor in cxx

daniel-riehm:
#1682 is still causing issues with Windows CI (see [here](http://canifis:8080/job/KwiverWindowsPR/392/consoleFull) for #1705). I think the `= default` for the destructor may be the problem, as it doesn't know how to destruct the `unique_ptr` when included from `register_applets.cxx`, which doesn't have access to the `priv` definition. We'll see if I'm right when the CI runs on this PR.

### DIFF
--- a/arrows/mvg/applets/bundle_adjust_tool.cxx
+++ b/arrows/mvg/applets/bundle_adjust_tool.cxx
@@ -1277,6 +1277,11 @@ bundle_adjust_tool
   : d( new priv( this ) )
 {}
 
+// ----------------------------------------------------------------------------
+bundle_adjust_tool
+::~bundle_adjust_tool()
+{}
+
 } // mvg
 
 } // arrows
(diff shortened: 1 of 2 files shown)


arrows/mvg/applets/bundle_adjust_tool.cxx
```diff
@@ -1277,6 +1277,11 @@ ::bundle_adjust_tool()
   : d( new priv( this ) )
 {}
 
+// ----------------------------------------------------------------------------
+bundle_adjust_tool
+::~bundle_adjust_tool()
+{}
```
Purg, line 1283: LGTM. In the past I've also been able to just write in the CXX the likes of `bundle_adjust_tool::~bundle_adjust_tool() = default;`, but there is certainly nothing wrong with this.

Purg (approved):

Purg:
Are the 4 kwiver builds all that are expected? (would be great if there was a linkage with a github check so they could get checked off :joy_cat:)

arrows/mvg/applets/bundle_adjust_tool.cxx
daniel-riehm: Yeah, this may technically be a compiler bug (but I don't know the C++ standard well enough to be sure).

PR #1728, opened 2023-02-07
Implement remaining ST1303 MDAP APAs

daniel-riehm:
This PR implements the BOOLEAN, UINT, and RLE array proccessing algorithms for a KLV ST1303 MDAP (multi-dimensional array pack). MDAPs are used to compactly store numeric arrays of data in KLV. The MDAP specification allows for several APAs (array processing algorithms) to compress multi-dimensional arrays in KLV.

- NATURAL: No special processing; encode each element as-is, with a fixed length for all elements (already implemented)
- IMAP: Encode floating-point values that always fall in a specific range as a fixed-length integer which maps to that range (already implemented)
- BOOLEAN: Encode boolean values as tightly-packed bits instead of using a whole byte for each element (newly implemented)
- UINT: Encode unsigned integer values using the BER-OID algorithm to self-encode each integer's length instead of using a fixed length, subtracting out a constant bias from each integer to save bytes (newly implemented)
- RLE: Losslessly compress data using run-length encoding, describing the array in terms of rectangular blocks of the same value. By far the trickiest of the five, especially since MDAPs are not restricted to one or two dimensions (newly implemented)

Unit tests are included, using sample data from the standard when available.

### DIFF
--- a/arrows/klv/klv_1303.hpp
+++ b/arrows/klv/klv_1303.hpp
@@ -14,7 +14,10 @@
 
 #include <vital/range/iota.h>
 
+#include <list>
 #include <numeric>
+#include <sstream>
+#include <type_traits>
 
 namespace kv = kwiver::vital;
 namespace kvr = kwiver::vital::range;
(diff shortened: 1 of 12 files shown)


daniel-riehm:
jenkins test this please

arrows/klv/klv_1303.hpp
```diff
@@ -25,6 +28,343 @@ namespace arrows {
...
+      entry = tmp_entries.end();
+    };
+
+  // Create an entry starting at the current coordinates
+  auto create_entry =
+    [ & ]( size_t i ){
+      auto parent = tmp_entries.end();
+      auto const element = value.elements[ i ];
+
+
```
chetnieter, line 255: Remove excess white space

chetnieter (approved):

PR #1730, opened 2023-02-15
Replace set_klv() with direct access

daniel-riehm:
This PR changes the `klv_metadata` interface to allow direct modification of the KLV packets, instead of requiring modifications to be made only in bulk via `set_klv()`. This prevents unnecessary copying / awkward temporary variables when, for example, one just needs to append a single new item to the vector of KLV packets.


daniel-riehm:
jenkins test this please

PR #1731, opened 2023-02-17
Add buffered_metadata_filter

daniel-riehm:
This PR adds `buffered_metadata_filter`, an algorithm type heavily based on the existing `metadata_filter`, except that it allows an algorithm to buffer multiple "lookahead" frames before returning the next result. This context may be necessary for some algorithms to function properly.


daniel-riehm:
jenkins test this please

chetnieter (commented):
I don't see any tests for this new class, although I noticed there are also no tests for the original `video_input_metadata_filter` class. Is this due to the challenge of getting/creating a test video with appropriate metadata.

I also wonder about how this class and `video_input_metadata_filter` derive from both `video_input` and the filter class. Is there a reason for this or would it be possible to engineer these two classes into on class where the filter is just a data member? Maybe a task for a separate PR?

daniel-riehm:
You are thinking that `metadata_filter` would derive from `buffered_metadata_filter` , acting as a simplifying layer on top so existing `metadata_filter`s wouldn't have to change? I think that's a good idea. Would that entail removing the factory method for `metadata_filter`, meaning config files would have to change?

chetnieter:
> You are thinking that `metadata_filter` would derive from `buffered_metadata_filter` , acting as a simplifying layer on top so existing `metadata_filter`s wouldn't have to change? I think that's a good idea. Would that entail removing the factory method for `metadata_filter`, meaning config files would have to change?

Yes - something like that. Not being able to see how either class is used in context makes it difficult to know what the best approach is.

daniel-riehm:
I have a branch to modify the `update_klv` `metadata_filter` into a `buffered_metadata_filter` (the impetus behind this PR). You could look at the current version of `update_klv`, or `derive_metadata`, for examples of `metadata_filter`s, and [here](https://gitlab.kitware.com/kwiver/burnout/-/blob/master/config/gui_image_video_reader.conf) for an example of how a config file would construct a pipeline in which an `ffmpeg` reader reads a video, then passes its output through a `derive_metadata` `metadata_filter`. Whether or not we end up combining the classes in a future PR, the buffered version of the logic in this PR would need to land first.

chetnieter:
> I have a branch to modify the `update_klv` `metadata_filter` into a `buffered_metadata_filter` (the impetus behind this PR). You could look at the current version of `update_klv`, or `derive_metadata`, for examples of `metadata_filter`s, and [here](https://gitlab.kitware.com/kwiver/burnout/-/blob/master/config/gui_image_video_reader.conf) for an example of how a config file would construct a pipeline in which an `ffmpeg` reader reads a video, then passes its output through a `derive_metadata` `metadata_filter`. Whether or not we end up combining the classes in a future PR, the buffered version of the logic in this PR would need to land first.

Let's go ahead and land this MR then

chetnieter (approved):

PR #1733, opened 2023-02-23
Add compare-klv applet

daniel-riehm:
This PR adds the `compare-klv` applet, which will pretty-print any differences found between the KLV contents of two files.

### DIFF
--- a/arrows/klv/applets/compare_klv.cxx
+++ b/arrows/klv/applets/compare_klv.cxx
@@ -1,0 +1,856 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/// \file
+/// Definition of compare-klv applet.
+
+#include <arrows/klv/applets/compare_klv.h>
+
+#include <arrows/core/metadata_map_io_csv.h>
+#include <arrows/core/metadata_stream_from_video.h>
+#include <arrows/klv/klv_all.h>
+#include <arrows/klv/klv_key_traits.h>
+#include <arrows/klv/klv_metadata.h>
+
+#include <vital/types/metadata_stream.h>
+#include <vital/types/metadata_stream_from_map.h>
+#include <vital/algo/metadata_map_io.h>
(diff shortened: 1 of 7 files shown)


bradh (commented):
If you sort, are you losing context in something like SDCC?

The case I'm thinking about is where the values are the same, but the previous entries (what ST 1010 calls the Refined Source List) are in a different order, so the SDCC doesn't have the same meaning.

daniel-riehm:
> If you sort, are you losing context in something like SDCC?
> 
> The case I'm thinking about is where the values are the same, but the previous entries (what ST 1010 calls the Refined Source List) are in a different order, so the SDCC doesn't have the same meaning.

Good question. The `klv_1010_sdcc_flp` structure (`arrows/klv/klv_1010.h`) has the (in retrospect, not super well-named) `members` member, which contains the ordered tags of the Refined Source List. Two `klv_1010_sdcc_flp`s do not compare as equal unless their `members` are equal in addition to their actual data. So once an SDCC knows what should come before it, no important information is lost when reordering.

arrows/klv/applets/compare_klv.cxx
```diff
@@ -0,0 +1,901 @@
...
+namespace kwiver {
+
+namespace arrows {
+
+namespace klv {
+
+namespace {
+
+// ----------------------------------------------------------------------------
+// Video or JSON file
```
chetnieter, line 40: At the very least I would suggest this interface and its derived classes be in a separate header and source file. I think we should consider moving the interface to an appropriate place in vital and putting the implementations in `arrows/core`. They are pretty light weight but I can see them being useful to other applications or tools.

chetnieter (changes requested):

vital/algo/metadata_stream_from_map_io.h
chetnieter, line 1: Empty file or a GitHub quirk that it is still listed?

chetnieter (approved):
Remove that empty file if it is really there.

vital/algo/metadata_stream_from_map_io.h
daniel-riehm: Oops - I must have messed up a rebase somewhere.

PR #1734, opened 2023-02-27
Check for common checksum implementation errors

daniel-riehm:
It appears that some KLV data producers implement the checksum algorithms incorrectly: they compute the checksum of only the payload instead of the entire KLV packet. When processing such data, the output is flooded with `ERROR` logs noting the incorrect checksums, even though the packets are not actually corrupted. This PR adds checks for this sort of incorrect implementation, downgrading these cases to a `DEBUG` log level.

To avoid the checksum section becoming too unwieldy in the larger `klv_read_packet()` function, it is split off into its own function.

### DIFF
--- a/arrows/klv/klv_packet.cxx
+++ b/arrows/klv/klv_packet.cxx
@@ -9,6 +9,8 @@
 
 #include <arrows/klv/klv_all.h>
 
+#include <vital/range/iterator_range.h>
+
 #include <iomanip>
 
 namespace kv = kwiver::vital;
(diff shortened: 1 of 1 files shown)


daniel-riehm:
jenkins test this please

chetnieter (approved):

bradh:
Quick question on the background to this: are you seeing this on common local sets (e.g ST 0601 ) or is it more niche?

daniel-riehm:
> Quick question on the background to this: are you seeing this on common local sets (e.g ST 0601 ) or is it more niche?

I've seen it at least in ST0806 (RVT). Possibly in others, but I haven't looked comprehensively.

PR #1740, opened 2023-03-13
Make update_klv a buffered_metadata_filter

daniel-riehm:
Currently, `update_klv` can only look at the current frame's metadata when putting GSD and VNIIRS into KLV packets. However, the ST1108 packets that these metrics go in are somewhat verbose and probably not intended to accompany each frame; that's why they include start and end timestamps. Using the new `buffered_metadata_filter`, this PR modifies `update_klv` to possibly sample or average these metrics over any number of frames, greatly reducing overhead by only encoding one ST1108 packet every `n` frames.

### DIFF
--- a/arrows/klv/update_klv.cxx
+++ b/arrows/klv/update_klv.cxx
@@ -15,6 +15,9 @@
 #include <arrows/klv/klv_1108_metric_set.h>
 #include <arrows/klv/misp_time.h>
 
+#include <vital/range/iterator_range.h>
+
+#include <list>
 #include <optional>
 
 namespace kwiver {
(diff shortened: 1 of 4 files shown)


arrows/klv/update_klv.cxx
```diff
@@ -176,7 +410,26 @@ vital::config_block_sptr
...
+    "How often (in frames) to encode a ST1108 packet." );
+  config->set_value(
+    "st1108_inter", d->st1108_inter,
+    "How to deal with a group of multiple frames when st1108_frequency > 1. "
+
+    "'sample' will create a packet with the metric values of the first frame "
+    "of the group and associate it with the first frame only, leaving the rest "
+    "of the frames in the group with no associated values. "
+
+    "'sample_smear' will create a packet with the metric values of the first "
```
chetnieter, line 426: I don't see this addressed anywhere (no check or conditional) or is this the effective default so no checks are required?

chetnieter (changes requested):

chetnieter (approved):

PR #1746, opened 2023-03-29
Add the text_codec interface and ASCII, UTF-8, UTF-16 codecs

daniel-riehm:
This PR adds an interface to handle different text encodings, as well as several implementations of that interface. The interface is fairly low-level (e.g. uses pointers instead of `std::[io]stream`s) and detailed (e.g. allows customization of error behavior) to provide a maximally correct, complete, flexible, and performant backbone.

I can split this into smaller PRs if that would make it easier for reviewers, but I thought that seeing the needs of different encodings that informed design decisions might be informative.


vital/util/text_codec_priv.h
```diff
@@ -0,0 +1,68 @@
+// This file is part of KWIVER, and is distributed under the
+// OSI-approved BSD 3-Clause License. See top-level LICENSE file or
+// https://github.com/Kitware/kwiver/blob/master/LICENSE for details.
+
+/// \file
+/// Utilities to make implementing text codecs easier.
+
+#ifndef KWIVER_VITAL_UTIL_TEXT_CODEC_PRIV_H_
+#define KWIVER_VITAL_UTIL_TEXT_CODEC_PRIV_H_
+
```
chetnieter, line 10: I know it will probably complicate things with the nested macros but having some of the variables such as `decoded_begin` and `encoded_begin` being assumed to be present without context could make maintenance a challenge later. Could we define these as parameters to the macros, say like `ENCODE_HANDLE_ERROR(c, begin, end)`. I know this makes things more verbose but I think it would make things more maintainable.

chetnieter (changes requested):

vital/util/text_codec_priv.h
chetnieter: Let's pass `c` and `code` as parameters and continue to assume that the others will be present.

chetnieter (changes requested):

chetnieter (approved):

PR #1747, opened 2023-03-29
Ensure compile-time compatibility with FFmpeg v5 API changes

daniel-riehm:
FFmpeg v5 introduces a few small API tweaks, which are addressed here.


daniel-riehm:
jenkins test this please

PR #1769, opened 2023-05-24
Release v1.8.0

daniel-riehm:
This PR merges `master` into `release`, adding a single commit at the end which updates the version number and release notes appropriately for the change to v1.8.0. The existing `release` branch can still be worked on via the new `v1.7.X` branch - we may want to consider releasing v1.7.1 at some point with accumulated bug fixes. When this PR is merged, a `v1.8.0` tag should be added to the merge commit, then `release` should be merged back into `master`.


daniel-riehm:
I anticipate CI failure due to fletch's `release` being old.

daniel-riehm:
> I don't think we need to make this merge. We usually only merge `release` into `master`, not the other way around. When we want to make a minor release we just update the `release` tag to point to the same commit as `master`.

That works too; I've changed the target branch to `master` since we still need to change the version number and release notes.

chetnieter (approved):

PR #1770, opened 2023-06-02
Fix integer promotion bug

daniel-riehm:
This PR fixes a bug in which C++'s arithmetic integer promotion rules cause an integer value to end up as an `int` instead of a `uint8_t`, which messes up the writing operation because `int` is signed.


chetnieter (approved):

daniel-riehm:
jenkins test this please

PR #1771, opened 2023-06-02
Better define comparison behavior with NaNs

daniel-riehm:
`NaN` comparisons always return false, which is not desirable behavior for consistently sorting or comparing. We had already accounted for this when comparing floating-point `klv_value`s, but not for floating-point members of a `struct` (pack). This PR fixes that.


daniel-riehm:
jenkins test this please

daniel-riehm:
jenkins test this please

chetnieter (approved):

PR #1773, opened 2023-06-02
Consistently catch video open failures in applets

daniel-riehm:
This PR ensures that failure to open a video file does not kill an applet by uncaught exception, but rather is handled gracefully. I realize that this code is very similar across the three applets, but am unsure if it is worth the effort to make this a utility function, or where we would put that if it is worth the effort. Reviewer input is welcome.


daniel-riehm:
jenkins test this please

arrows/core/applets/dump_klv.cxx
```diff
@@ -223,14 +223,12 @@ ::run()
   }
   catch ( kv::video_exception const& e )
   {
-    std::cerr << "Video Exception-Couldn't open \"" << video_file << "\"" << std::endl
-              << e.what() << std::endl;
+    std::cerr << e.what() << std::endl;
```
chetnieter, line 226: I assume the addition text was redundant to the exception message?

chetnieter (approved):
Definitely some redundant code but it is only repeated once and the applets could use a larger refactor in the future so it could be addressed then.

PR #1775, opened 2023-06-02
Account for missing frames in compare-klv

daniel-riehm:
This PR ensures that `compare-klv` will let one video "catch up" if it has a frame that the other one does not. Previously, the frames were simply run through in order, but that meant that if one video was missing e.g. frame 100, then for the rest of the videos the frame numbers being compared would be mismatched by one (99 vs 99, then 100 vs 101, 101 vs 102, etc), giving the impression of there being many more differences than are actually present.


daniel-riehm:
jenkins test this please

chetnieter (approved):

PR #1789, opened 2023-09-25
Expose FFmpeg codec options

daniel-riehm:
This PR allows manipulation of implementation-specific codec behavior settings that `AVCodecParameters` is too generic to include.


daniel-riehm:
CI failures unrelated

PR #1792, opened 2023-09-26
Add more constructors for klv_tag_traits

daniel-riehm:
This PR adds default and copy constructors.


borovik135 (approved):
Proposed changes look OK.

PR #1793, opened 2023-09-27
Enable FFmpeg threading by default

daniel-riehm:
This PR tells FFmpeg to use its internal multithreading capabilities when encoding and decoding frames. Some encoders, e.g. libx264, have this enabled by default already, but most decoders do not, unnecessarily slowing down processing. Perhaps a future PR will add config parameters to allow manual control over the number of threads, but for right now FFmpeg gets to choose.


borovik135 (approved):
Proposed changes are OK.

PR #1796, opened 2023-10-10
Update uncrustify

daniel-riehm:
This PR updates a few deprecated / renamed options for `uncrustify` as of version `0.77.1`, and tweaks various other options to bring them more in line with current KWIVER style. This PR is organized in three commits - the first one contains only the updates to the configuration; the second contains the raw output when uncrustify is now run on the entire FFmpeg arrow; the third contains some tweaks to the raw output which produce a somewhat cleaner result. These tweaks are "`uncrustify`-stable" - i.e. uncrustify will not revert them when run again, but uncrustify did not produce them on the first run because it attempts to respect existing newlines in e.g. function argument lists, which is often but not always desirable. Simply removing extraneous newlines yourself is often sufficient to fix any suboptimal results.

With these new config changes and an upgrade to `0.77.1`, I no longer observe `uncrustify` insisting on style that is counter to KWIVER practices, as I have in the past. However, I've only manually tested on a few arrows, so I welcome counterexamples.

Reviewers may comment on whether the second two commits should be merged along with with the first one, or perhaps any actual large-scale reformatting can wait for another day. Regardless, these commits serve as a preview of `uncrustify`'s behavior.

I believe `uncrustify` is at a point where we could potentially start using it in CI to enforce style, once enough of the codebase is conformant. This would have the benefit of removing most style nitpicks from regular PRs.


borovik135 (approved):
Changes appear to be OK.

PR #1797, opened 2023-10-10
Delay discovery of GTest tests

### LINKED ISSUE #1370: Building tests requires sourcing `setup_KWIVER.sh`
This is something I've hit a number of times and find relatively annoying. If I try to build KWIVER with tests, I get a strange gtest linking error if I don't source `setup_KWIVER.sh` in the build tree. 

My config is Ubuntu 18.04, `g++7.5.0`, and the `zsh` shell. If I end up hitting the issue again I'll post the error.

daniel-riehm:
Fixes #1370 with CMake >= 3.18. Shouldn't require updating minimum CMake version, since we are just setting a variable which can be ignored by earlier versions. Of course, actually running the tests will still require `source setup_KWIVER.sh`.

### DIFF
--- a/CMake/utils/kwiver-utils-tests.cmake
+++ b/CMake/utils/kwiver-utils-tests.cmake
@@ -63,6 +63,9 @@
 
 include(GoogleTest)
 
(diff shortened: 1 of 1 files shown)


borovik135 (approved):
Proposed change looks OK

PR #1798, opened 2023-10-10
Add sphinx_rtd_theme as an RTD dependency

daniel-riehm:
Attempting to fix the docs build which has broken again.


borovik135 (approved):
change looks fine

PR #1799, opened 2023-10-16
Implement packet lookahead for FFmpeg video input

daniel-riehm:
This code has already been reviewed by @chetnieter and merged in another repo, but these particular improvements should be merged into the main KWIVER repo so that they will be maintained in the future. This MR improves the FFmpeg video reader's logic in cases where different streams are out of sync, have corrupt timestamps, etc.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_output.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_output.cxx
@@ -397,7 +397,7 @@ ffmpeg_video_output
   result->frame_rate = d->video->video_stream->avg_frame_rate;
   avcodec_parameters_from_context( result->parameters.get(),
                                    d->video->codec_context.get() );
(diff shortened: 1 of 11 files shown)


arrows/ffmpeg/ffmpeg_video_output.cxx
```diff
@@ -397,7 +397,7 @@ ::implementation_settings() const
   result->frame_rate = d->video->video_stream->avg_frame_rate;
   avcodec_parameters_from_context( result->parameters.get(),
                                    d->video->codec_context.get() );
-  result->klv_stream_count = 0; // TODO
+  result->klv_streams = {}; // TODO
```
chetnieter, line 400: Do we know why there is a TODO comment here with no context? Can we just remove it?

chetnieter (approved):

PR #1800, opened 2023-10-16
Add interfacing for copying uninterpreted data streams

daniel-riehm:
This PR adds interfaces and pass-through code enabling the copy of data in a video which our readers cannot currently interpret, but that we may want to preserve when writing the video back out. Audio, secondary video streams, and unrecognized metadata streams are the target examples here. This PR does not actually implement doing the copy, but merely adds room for it in our interfacing and existing tools.

### DIFF
--- a/vital/algo/video_input.h
+++ b/vital/algo/video_input.h
@@ -20,6 +20,7 @@
 #include <vital/types/timestamp.h>
 #include <vital/types/video_raw_image.h>
 #include <vital/types/video_raw_metadata.h>
+#include <vital/types/video_uninterpreted_data.h>
 #include <vital/types/video_settings.h>
 
 #include <string>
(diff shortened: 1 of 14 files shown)


vital/algo/video_input.h
```diff
@@ -344,6 +346,16 @@ class VITAL_ALGO_EXPORT video_input
...
   virtual video_raw_metadata_sptr raw_frame_metadata();
 
+  /// Return an implementation-defined representation of uninterpreted data in
+  /// this frame.
+  ///
+  /// This method enables passage of miscellaneous data - such as audio,
+  /// unrecognized metadata, or secondary image streams - to a video output when
+  /// transcoding.
+  ///
+  /// \return Pointer to raw data.
```
borovik135, line 356: consider changing [raw]->[uninterpreted]

borovik135 (changes requested):
a minor change request

borovik135 (approved):
OK, looks fine to me now.

PR #1802, opened 2023-10-30
Add audio stream copying

daniel-riehm:
This PR adds the capability to copy any number of audio streams from an FFmpeg video input to output. No interpretation of the audio data is available, but the information will not be lost during transcoding.

This functionality has been manually tested on a couple random video clips (`.ts` and `.mp4`). None of our existing test videos have audio; not sure to what extent we should put effort into constructing a short test clip to put in the repo for unit testing, or if this is a niche enough feature that it's not worth the time / additional file size.


borovik135:
ghostflow-check-master check fails, needs to be fixed first. A short test video with multiple audio streams recommended.

arrows/ffmpeg/ffmpeg_audio_stream_settings.h
```diff
@@ -0,0 +1,49 @@
...
+  ffmpeg_audio_stream_settings( ffmpeg_audio_stream_settings&& other );
+
+  ~ffmpeg_audio_stream_settings();
+
+  ffmpeg_audio_stream_settings&
+  operator=( ffmpeg_audio_stream_settings const& other );
+  ffmpeg_audio_stream_settings&
+  operator=( ffmpeg_audio_stream_settings&& other );
+
+  int index;
```
borovik135, line 38: Is there a reason for public data members?

borovik135 (commented):
a question regarding public data members

arrows/ffmpeg/ffmpeg_audio_stream_settings.h
daniel-riehm: This `struct` is used only to pass these data items from input to output, with perhaps occasional use-case-specific editing going on in between. We could, of course, add setters and getters for each member and make them private, but in the absence of any complex behavior being implemented by this `struct` that would necessitate a layer of abstraction, I opted not to -- seemed unnecessary at the time. You may see I made similar decisions in e.g. `struct ffmpeg_video_settings`.

borovik135: Since those structures are defined in the header files, a comment explaining public data members, e.g. "used only to pass these data items from input to output, with perhaps occasional use-case-specific editing going on in between", would be nice.

arrows/ffmpeg/ffmpeg_audio_stream_settings.h
```diff
@@ -22,6 +22,11 @@ namespace arrows {
 namespace ffmpeg {
 
 // ----------------------------------------------------------------------------
+/// Parameters describing the general characteristics of an audio stream.
+///
+/// This struct will be filled in by ffmpeg_video_input, to be used by
+/// ffmpeg_video_output when creating an audio stream. They have been left
```
borovik135, line 28: Consider replacing [They]->[Members]

PR #1805, opened 2023-11-07
Increase output precision of floats when writing CSVs

daniel-riehm:
This PR increases the number of digits printed when writing floating-point numbers to a CSV, so that information is not lost in the process.


borovik135 (approved):
Change looks OK.

PR #1806, opened 2023-11-08
Fix packet synchronization bug

daniel-riehm:
This bug got missed in #1802 due to a typo in one of the unit tests causing it to run on the wrong video file. Run on the correct file, the unit test fails, alerting us that copying both the video and audio streams causes a desync between them. The bug may be fixed by adjusting how we adjust timestamps in accordance with `start_timestamp` - instead of having FFmpeg shift all packets forward, then manually shifting back the ones that are direct from the original video (and therefore already take `start_timestamp` into account), we tell FFmpeg to do nothing and manually shift forward packets that aren't from the original video (i.e. only video packets encoded by us). Alternately, we could have added backward shifting to the copied video packets and kept the rest as is, but this results in code duplication and is probably less clear anyway.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_output.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_output.cxx
@@ -511,8 +511,6 @@ ffmpeg_video_output::impl::open_video_state
   }
   output_format = format_context->oformat;
 
-  // Set timestamp value to start at
-  format_context->output_ts_offset = settings.start_timestamp;
   format_context->flags |= AVFMT_FLAG_AUTO_BSF;
   format_context->flags |= AVFMT_FLAG_GENPTS;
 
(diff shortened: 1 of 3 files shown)


arrows/ffmpeg/ffmpeg_video_output.cxx
```diff
@@ -512,7 +512,6 @@ ::open_video_state(
   output_format = format_context->oformat;
 
   // Set timestamp value to start at
```
borovik135, line 514: may want to remove/change the comment as well

arrows/ffmpeg/ffmpeg_video_settings.h
```diff
@@ -71,9 +71,9 @@ struct KWIVER_ALGO_FFMPEG_EXPORT ffmpeg_video_settings
   /// guaranteed to determine the time base in the output video.
   AVRational time_base;
 
-  /// Desired PTS of the first video frame, in AV_TIME_BASE units
-  /// (microseconds). For some formats, the actual first PTS may exceed this
-  /// value to ensure non-negative PTS or DTS.
+  /// Start time of the input video, in AV_TIME_BASE units (microseconds).
+  /// This information is necessary for copied and newly-encoded packets to sync
+  /// correctly.
   int64_t start_timestamp;
```
borovik135, line 77: consider variable initialization

borovik135 (commented):
there are a couple of comments regarding the changes

arrows/ffmpeg/ffmpeg_video_settings.h
daniel-riehm: Present in the constructor in `ffmpeg_video_settings.cxx`.

borovik135 (approved):
Changes look OK.

Issue #1640, opened 2022-06-29
Mesh_io should throw exceptions on invalid input


OwenMcGee:
Currently in [`read_mesh`](https://github.com/Kitware/kwiver/blob/master/vital/io/mesh_io.cxx#L98) in `vital/io/mesh_io.cxx`, if the `filename` parameter does not end with a valid file extension (`.ply2`, `.ply`, or `.obj`), the function still returns the uninitialized `mesh_sptr`, when it should throw an exception indicating the bad input.

Outside of the function itself, the only reference to this code that would need to be updated as far as I can tell is in [`vital/tests/test_mesh_io.cxx`](https://github.com/Kitware/kwiver/blob/master/vital/tests/test_mesh_io.cxx#L62)

PR #1807, opened 2023-11-16
Add error handling to mesh_io

aramSofthenge:
- Introduced exceptions in mesh read functions for better error management. 
- Updated tests to validate new behavior.

### DIFF
--- a/vital/io/mesh_io.cxx
+++ b/vital/io/mesh_io.cxx
@@ -95,7 +95,8 @@ read_mesh(const std::string& filename)
     return read_obj(input_stream);
   }
 
-  return mesh_sptr();
+  VITAL_THROW( invalid_file, filename,
+               "Unrecognized file extension for mesh file");
 }
 
 /// Read a mesh from a PLY2 file
@@ -104,6 +105,7 @@ read_ply2(const std::string& filename)
 {
   check_input_file(filename);
   std::ifstream input_stream(filename.c_str());
+
   return read_ply2(input_stream);
 }
 
@@ -341,7 +343,8 @@ read_obj(std::istream& is)
                 else
                 {
                   LOG_ERROR(logger, "improperly formed face line in OBJ: "<<line);
-                  return mesh_sptr();
+                  VITAL_THROW( invalid_data,
+                               "Improperly formed face line in OBJ file.");
                 }
               }
             }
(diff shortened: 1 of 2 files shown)


aramSofthenge:
fixes #1640

chetnieter:
jenkins test this please

vital/io/mesh_io.cxx
```diff
@@ -104,6 +105,13 @@ read_ply2(const std::string& filename)
 {
   check_input_file(filename);
   std::ifstream input_stream(filename.c_str());
+
+  if ( ! input_stream.is_open() )
```
chetnieter, line 109: It looks like a very similar check is already done in the `check_input_file` function: https://github.com/Kitware/kwiver/blob/75369be0a22ff9b719ce216dd05125fd4f14328e/vital/io/mesh_io.cxx#L69

vital/io/mesh_io.cxx
```diff
@@ -263,6 +271,13 @@ read_obj(const std::string& filename)
 {
   check_input_file(filename);
   std::ifstream input_stream(filename.c_str());
+
```
chetnieter, line 274: See comment above, file stream is checked in `check_input_file`.

chetnieter (changes requested):
Looks good in general. I would suggest two additional changes. Since a file stream is opened in `check_input_file` we might as well return that or a smart pointer to that file stream rather than just opening a new one.

https://github.com/Kitware/kwiver/blob/75369be0a22ff9b719ce216dd05125fd4f14328e/vital/io/mesh_io.cxx#L68

Likewise, an output file stream is opened for no apparant reason in `check_output_file`. Perhaps we should again just return this stream and use it in the various `write_` methods. If this is done I would suggest renaming the help functions to `open_input_file` and `open_output_file` respectively.

vital/io/mesh_io.cxx
aramSofthenge: Thanks, I've removed the redundant checks as you suggested. The commit has been amended, and the changes are now pushed.

chetnieter:
jenkins test this please

chetnieter:
The linux test failure is due to a test that is unrelated to these changes and is known frequently fail due to timing out.

PR #1808, opened 2023-11-27
Add null checks and exception handling to processing functions

aramSofthenge:
- Implement try-catch for track association to catch invalid values.
- Ensure non-null inputs for cameras, landmarks, or tracks to avoid errors.

### DIFF
--- a/sprokit/processes/core/associate_detections_to_tracks_process.cxx
+++ b/sprokit/processes/core/associate_detections_to_tracks_process.cxx
@@ -134,7 +134,7 @@ associate_detections_to_tracks_process
     }
     catch ( vital::invalid_value const& e )
     {
-      LOG_ERROR( logger(),  "An error occurred: " << e.what());
+      LOG_ERROR( logger(), "Failed to associate tracks: " << e.what());
     }
   }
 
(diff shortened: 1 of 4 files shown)


sprokit/processes/core/associate_detections_to_tracks_process.cxx
```diff
@@ -126,9 +126,16 @@ ::_step()
...
-           tracks, detections, ass_matrix, output, unused );
+    try
+    {
+      // Run associator
+      d->m_track_associator->associate( frame_id, image,
+            tracks, detections, ass_matrix, output, unused );
+    }
+    catch ( vital::invalid_value const& e )
+    {
+      LOG_ERROR( logger(),  "An error occurred: " << e.what());
```
willdunklin, line 137: Would edit the message to have a more specific description of the error. For example, `"Failed to associate tracks: " << ...`.
Also, you have an extraneous space following `logger(),  ` :)

willdunklin (changes requested):

willdunklin:
Looks fine to me outside of that one minor formatting discrepancy

sprokit/processes/core/associate_detections_to_tracks_process.cxx
aramSofthenge: Thanks a lot for the feedback! I've implemented the suggested changes :)

PR #1810, opened 2023-12-04
Add ffmpeg_video_input_clip

daniel-riehm:
This PR  adds the `ffmpeg_video_input_clip` class, which is a wrapper around a regular `ffmpeg_video_input` that takes in a frame range in its config, and attempts to behave exactly as if the input video consisted only of that frame range. There is also an option to automatically seek back to the most recent keyframe, to avoid the first few frames being undecodable when copying the raw video stream. Accomplishing this required some significant work to be done on the seeking logic of `ffmpeg_video_input`.

### DIFF
--- a/arrows/ffmpeg/ffmpeg_video_input.cxx
+++ b/arrows/ffmpeg/ffmpeg_video_input.cxx
@@ -291,14 +291,9 @@ ffmpeg_klv_stream
   klv_result.add< kv::VITAL_META_METADATA_ORIGIN >( "KLV" );
   klv_result.add< kv::VITAL_META_VIDEO_DATA_STREAM_INDEX >( stream->index );
 
-#if LIBAVCODEC_VERSION_MAJOR > 57
-  if( stream->codecpar->profile >= 0 )
-  {
-    klv_result.add< kv::VITAL_META_VIDEO_DATA_STREAM_SYNCHRONOUS >(
-      stream->codecpar->profile == FF_PROFILE_KLVA_SYNC
-    );
-  }
-#endif
+  klv_result.add< kv::VITAL_META_VIDEO_DATA_STREAM_SYNCHRONOUS >(
+    settings().type == klv::KLV_STREAM_TYPE_SYNC
+  );
 
   return result;
 }
@@ -414,8 +409,10 @@ public:
 
     bool try_codec();
     void init_filters( filter_parameters const& parameters );
-    bool advance();
-    void seek( kv::frame_id_t frame_number );
+    bool advance( bool is_first_frame_of_seek = false );
+    void clear_state_for_seek();
+    void seek_to_start();
+    void seek( kv::frame_id_t frame_number, seek_mode mode );
     void set_video_metadata( kv::metadata& md );
     double curr_time() const;
     double duration() const;
@@ -443,12 +440,15 @@ public:
 
     sws_context_uptr image_conversion_context;
 
+    std::optional< kv::frame_id_t > frame_count;
     int64_t start_ts;
+    AVRational maybe_frame_rate;
     std::map< int64_t, klv::misp_timestamp > pts_to_misp_ts;
     std::map< int64_t, int64_t > packet_pos_to_dts;
     int64_t prev_frame_dts;
     int64_t prev_video_dts;
     std::multimap< int64_t, packet_uptr > lookahead;
+    std::list< packet_uptr > raw_image_buffer;
 
     std::list< ffmpeg_klv_stream > klv_streams;
     kv::metadata_map_sptr all_metadata;
@@ -916,12 +916,15 @@ ffmpeg_video_input::priv::open_video_state
     filter_sink_context{ nullptr },
     filter_source_context{ nullptr },
     image_conversion_context{ nullptr },
+    frame_count{},
     start_ts{ 0 },
+    maybe_frame_rate{ 0, 0 },
     pts_to_misp_ts{},
     packet_pos_to_dts{},
     prev_frame_dts{ AV_NOPTS_VALUE },
     prev_video_dts{ AV_NOPTS_VALUE },
     lookahead{},
+    raw_image_buffer{},
     klv_streams{},
     all_metadata{ nullptr },
     audio_streams{},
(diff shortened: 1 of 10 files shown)


arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -1544,11 +1538,31 @@ ::advance()
...
+        if( frame_count )
+        {
+          ++( *frame_count );
+        }
 
         // Look up the dts of the packet that contained this frame
         if( auto const it = packet_pos_to_dts.find( frame->frame->pkt_pos );
             it != packet_pos_to_dts.end() )
         {
+          for( auto jt = raw_image_buffer.begin();
```
sedrakSofthenge, line 1550: This looks like more a while loop, why not use while instead?

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -1697,71 +1798,204 @@ ::seek( kv::frame_id_t frame_number )
...
+            parent->logger,
+            "KLV re-seek failed; "
+            "KLV reported for first frame may be incomplete" );
+        }
+      }
+
+      if( mode != SEEK_MODE_EXACT )
+      {
+        auto& image_packets = frame->get_raw_image().packets;
+        for( auto it = image_packets.begin(); it != image_packets.end(); )
```
sedrakSofthenge, line 1899: Why not to use while instead? Looks more like a while loop

arrows/ffmpeg/ffmpeg_video_input.cxx
```diff
@@ -1965,9 +2217,21 @@ ::timestamp() const
   {
     return {};
   }
-  return kv::timestamp{
-    static_cast< kv::time_usec_t >( curr_time() * 1000000.0 + 0.5 ),
-    frame_number() + 1 };
+
+  kv::timestamp ts;
```
sedrakSofthenge, line 2221: I assume this can be rewritten in the following way
```
  if( !frame.has_value() )
  {
    return {};
  }

  kv::timestamp ts;
    if( frame_rate().num > 0 )
  {
    ts.set_frame( frame_number() + 1 );
  } else if( frame->frame->best_effort_timestamp != AV_NOPTS_VALUE )
  {
    ts.set_time_usec(
      static_cast< kv::time_usec_t >( curr_time() * 1000000.0 + 0.5 ) );
  }
  return ts;
```
with this we don't need to check frame.has_value() twice (because initially its just returning with `!frame.has_value()`. Also avoid two independent if blocks, instead on if-else is enough.

arrows/ffmpeg/ffmpeg_video_input.cxx
daniel-riehm: Doing it this way limits the scope of `jt` to only inside the loop. This would allow another loop at the same scope level to use the variable name `jt` when iterating over a different sequence. The reasoning is similar to why `it` is declared in the `if` statement instead of just before it: minimizing unnecessary variable scope.

arrows/ffmpeg/ffmpeg_video_input.cxx
daniel-riehm: See response to similar suggestion above.

arrows/ffmpeg/ffmpeg_video_input.cxx
daniel-riehm: Changing it to an `else if` is not correct, because that changes the logic - in most cases we want both blocks to run. However, you are correct that the second `frame.has_value()` check is redundant.

arrows/ffmpeg/ffmpeg_video_input_clip.cxx
```diff
@@ -0,0 +1,438 @@
...
+
+  vital::metadata_map::map_metadata_t result;
+  vital::timestamp ts;
+  while( tmp_video.next_frame( ts ) )
+  {
+    result.emplace( ts.get_frame(), tmp_video.frame_metadata() );
+  }
+
+  d->all_metadata.reset(
+    new vital::simple_metadata_map{ std::move( result ) } );
```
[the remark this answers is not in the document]
daniel-riehm: It is generated by the compiler; see "Implicitly-defined move constructor" here: `https://en.cppreference.com/w/cpp/language/move_constructor`

arrows/ffmpeg/ffmpeg_video_input.cxx
sedrakSofthenge: Agree, my fault, somehow I had an impression like in two blocks its updating the same value.

arrows/ffmpeg/ffmpeg_video_input.h
```diff
@@ -27,6 +27,12 @@ class KWIVER_ALGO_FFMPEG_EXPORT ffmpeg_video_input
   : public vital::algo::video_input
 {
 public:
+  enum seek_mode {
+    SEEK_MODE_EXACT,
+    SEEK_MODE_KEYFRAME_BEFORE,
+    SEEK_MODE_KEYFRAME_AFTER,
```
chetnieter, line 33: Correct me if I am wrong, but currently these options are hidden from the user. If they are ever exposed via the configuration we should insure they are all tested. I assume the non-default option of SEEK_MODE_KEYFRAME_BEFORE is effectively tested in `test_ffmpeg_video_input_clip`.

chetnieter (approved):

arrows/ffmpeg/ffmpeg_video_input.h
daniel-riehm: You bring up a good point. These are in the implementation-specific header, so they are hidden from the user in the sense of not being accessible from a generic `video_input`. However, they are publicly accessible from an `ffmpeg_video_input` (and they need to be, for `ffmpeg_video_input_clip` to use them). `SEEK_MODE_EXACT` is the default behavior, and passes existing tests. `SEEK_MODE_KEYFRAME_BEFORE` is used in, and indirectly tested via, `ffmpeg_video_input_clip`. But `SEEK_MODE_KEYFRAME_AFTER` is not tested anywhere, nor is it used (currently) in `ffmpeg_video_input_clip`. I just included it for completeness. So I guess there are a few options here:

1. Leave as-is because none of this is accessible from `video_input`.
2. Add `ffmpeg_video_input` tests for `KEYFRAME_AFTER`.
3. Implement `KEYFRAME_AFTER` behavior in `ffmpeg_video_input_clip`, and add a test there.
4. Remove `KEYFRAME_AFTER` for now because there's no pressing reason to have it and it complicates the code.

Any thoughts on which is best?

daniel-riehm: @chetnieter (bump)

chetnieter: Let's go with 4. Correct if I am wrong, but it that option becomes necessary in the future, it is relatively trivial to add it back.

daniel-riehm: Sounds good, will do. I've retained the current version in a local branch in case it comes up again in the future.

PR #1817, opened 2024-01-15
FFmpeg corruption adjustments

daniel-riehm:
Some partially-corrupted videos will have segments of stream, sometimes quite long ones, which are undecodable. This currently causes a crash if it happens with the audio stream, and can cause performance issues if it's the video or KLV stream. These problems can be addressed by adjusting our algorithms to limit the size of buffers and/or reduce the need to iterate over them, and by being pickier about what we try to decode.


daniel-riehm:
Test failure unrelated.

chetnieter (approved):

PR #1819, opened 2024-02-07
Prevent crashes when writing out invalid wrapped-int data

daniel-riehm:
Currently, fixed-length `int`-like KLV formats (enums and bitfields) will crash if they are asked to write out a value longer than is allowed by the standard (as can happen when incorrect/corrupted data is being processed). With this PR, those formats will no longer crash, instead writing out the value as requested and printing a warning.


daniel-riehm:
(CI failure unrelated)

chetnieter (approved):

PR #1827, opened 2024-03-25
Implement rewiring from pure metadata sources

daniel-riehm:
This would allow us to, e.g., insert a KLV stream into a video sourced from a JSON file.


daniel-riehm:
jenkins test this please

chetnieter (approved):

PR #1835, opened 2024-05-23
Map FFmpeg DEBUG log level to vital TRACE

daniel-riehm:
DEBUG is the most verbose FFmpeg log level, and prints stuff like the name of every single CUDA function that is dynamically loaded in. This would be nice to be able to filter out, so this PR demotes it to our most verbose log level, TRACE.


daniel-riehm:
Closing as new development is moving to KWIVER 2.0 on GitLab.

PR #1838, opened 2025-12-06
Fix deadlock in embedded pipelines on python 3.8+ that didn't occur in 3.6

mattdawkins:


mattdawkins:
This has been handled in the gitlab instance

